PR #24, opened 2017-03-10
issue23: fix calculation of the type of MinusExpression

### LINKED ISSUE #23: Bug in inference of the MinusExpression on numbers
We have a case in which we think the inferred type is wrong.
When subtracting a value two values:

- a: number[1|20]
- b: number[1|10]

The inferred type is: number[0|10]{0}

I would expect it to number[-9|19]{0}

I think the error should be corrected in r:050f6d52-a81b-4b31-9a1c-531c1a04708e/5115872837157441716

Here:
```
node<NumberType> left = leftOperandType : NumberType; 
node<NumberType> right = rightOperandType : NumberType; 
node<NumberType> res = new node<NumberType>(); 
res.setRange(InfHelper.sub(left.lowerBound(), right.lowerBound(), false), InfHelper.sub(left.upperBound(), right.upperBound(), false)); 
res.setMaxPrecision(left, right);
```

Because you want to subtract the left.lowerBound and the right.upperBound to get the lower bound, and to get the resulting upperbound you should subtract the left.upperBound and the right.lowerBound

ftomassetti:
Fix #23 

I could not test that appropriately, sorry about that

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/typesystem.mps
@@ -1172,8 +1172,8 @@
                         <node concept="37vLTw" id="7Wa2sv3Kyvs" role="2Oq$k0">
                           <ref role="3cqZAo" node="3f3yNhCUncU" resolve="right" />
                         </node>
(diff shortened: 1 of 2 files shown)


ftomassetti:
Added tests:
screen shot 2017-03-13 at 10 01 07

I ran the tests from the command line and I had either 2 tests failing or all of them passing. I reported the issue about flaky tests #25

PR #79, opened 2017-10-27
added possibility for attributes to nest inside other attributes

bkolb:
* added this expression and context type provider to expr.base
* added pathelement to expr.path
* added EmptyAttribute and AttributeContainer to core.attributes
* various concepts in components.core implement context type provider

@arimer @wsafonov please review. see releated change in core

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ if (!project.hasProperty("mbeddrVersion")) {
     }
     else {
         //default version
-        ext.mbeddrVersion = '1.0.+'
+        ext.mbeddrVersion = '1.0.19511.3937f34'
     }
 }
 
--- a/code/languages/org.iets3.opensource/.mps/modules.xml
+++ b/code/languages/org.iets3.opensource/.mps/modules.xml
@@ -28,6 +28,7 @@
       <modulePath path="$PROJECT_DIR$/languages/org.iets3.glossary/org.iets3.glossary.mpl" folder="req" />
       <modulePath path="$PROJECT_DIR$/languages/org.iets3.req.core/org.iets3.req.core.mpl" folder="req" />
       <modulePath path="$PROJECT_DIR$/languages/org.iets3.req.plugin/org.iets3.req.plugin.msd" folder="req" />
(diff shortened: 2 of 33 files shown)


arimer:
Tests are missing for this feature.

wsafonov:
@arimer, I have the following two points:
- IContextTypeProvider should consistently return the node to calculate type from, not the type - so we can for example use that node in the type system rules to assign type to the ThisExpr.
- the good approach to combine with IContextTypeProvider would be now to check additionally for availability of IContextTypeProvider in the IConceptSpecificAttribute constraint. This way all IConceptSpecificAttribute would continue to work in the context of approriate type

wsafonov:
And as suggested before, we can mark EmptyAttribute as main attribute to enable its usage in the mainAttributes section to have consistent behaviour for all attributes list editors.

arimer:
@wsafonov 

**Regarding the first comment**

- I agree that IContextTypeProvider  should return the node where the type should be calculated from. I wanted to avoid to use the typesystem inside a constraint. Therefore the component 
- Refactorings of IConceptSpecificAttribute(s) are neccessary aswell.

**Regarding the second comment**

- I would **dissagree** here. If we make the EmptyAttribute a mainAttribute by default, the #mainAttribute# editorcomponent would show up every time the user (maybe accidentially) creates an EmptyAttribute in the inspector.

By the way I'll merge the changes inside this branch so that have the changes and the discussion in the same place.

arimer:
@wsafonov 
I've intorduced a new interface IAttributeContextType which can be used to flag types for scope calulation in the IConceptSpecificAttribute contraint. I think this should do the job for now.
Please have a look

wsafonov:
I'm not quite sure if this IAttributeContextType could be useful outside of the IConceptSpecificAttribute. For my taste it's too specific and we can actually solve the task here with a more light, generalized type interface, e.g. something like ITypeWithTarget providing method "node<> getTarget()" that would return the target node of any type - this would be sufficient to retrieve the concept to compare against canBeUsedUnderParents in constraint of IConceptSpecificAttribute. What do you think, @arimer ?
I would not then place this interface under components.core to allow its reuse outside of the component context, may be in core.base.

arimer:
Added the discussed changes.

PR #83, opened 2017-11-13
build setup: added new gradle task

arimer:
### Gradle build
-added setup task to gradle
-commited projectlibraries.overrides.properties files

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,4 @@
+import de.itemis.mps.gradle.GenerateLibrariesXml;
 // detect if we are in a CI build
 if (project.hasProperty("forceCI")) {
     ext.ciBuild = true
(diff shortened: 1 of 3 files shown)


projectlibraries.overrides.properties.example
```diff
@@ -0,0 +1,5 @@
+# To override values defined in this file create projectlibraries.overrides.properties and put overridden values there.
+mbeddr.plattform=${mbeddr.github.core.home}/code/languages
+mbeddr.sl-all=${mbeddr.github.core.home}/code/plugins
+iets3.os=${iets3.github.opensource.home}/code
```
arimer: Lines 3 and 4 does not make any sense right? @sergej-koscejev

sergej-koscejev: Why not?

build.gradle
```diff
@@ -1,3 +1,15 @@
+import de.itemis.mps.gradle.GenerateLibrariesXml;
+
+//will pull the groovy classes/types from nexus to the classpath
+buildscript {
+    repositories {
+        maven { url 'https://projects.itemis.de/nexus/content/repositories/mbeddr' }
+    }
+    dependencies {
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.0+'
```
sergej-koscejev: Perhaps we should fix the used plugin version to make builds reproducible? Why do we use wildcards (`1.0+`) here?

projectlibraries.overrides.properties.example
```diff
@@ -1,5 +1,3 @@
 # To override values defined in this file create projectlibraries.overrides.properties and put overridden values there.
 mbeddr.plattform=${mbeddr.github.core.home}/code/languages
-mbeddr.sl-all=${mbeddr.github.core.home}/code/plugins
-iets3.os=${iets3.github.opensource.home}/code
```
sergej-koscejev: Okay I get what you meant now :)

PR #89, opened 2017-11-27
#alias# editor components flagged with iets3Keyword styles should be deletable and non-editable

arimer:
In iets3.os we have currently 4x editors using an **# alias #** editor component which is flagged with the **iets3Keyword** style. 

grafik

In some cases (OptionType, AssertTestItem, TestOp) it is confusing when aliases of these concepts-instances can be edited. If we look in **iets.core** we can find about other 9 editors whith the same problem.

The proposal would be to set these # alias # editor components to `editable:false` and attach a generic delete action to this editor cells when it makes sense.

@markusvoelter and @coolya, @joao-silveira have you any opinion on it?

For now I will just create : 

- [x]  editor component (iets3KeyWordAlias) for not editable styled aliases which uses the generic delete action
- [x] a new styleclass which allows not editable keywords
And will not touch the iets expressions.


arimer:
I should probably also change the cell-id not to interfere with the editor component from mps.

PR #93, opened 2017-11-28
Issue 92: Making an integer constant assignable to number{inf}

### LINKED ISSUE #92: Numbers and subtyping
I have this test:
screen shot 2017-11-28 at 10 08 21
screen shot 2017-11-28 at 10 08 31
In practice an integer constant is not consider to be a subtype of a number with infinite precision, while a number with one decimal is consider to be a subtype of a number with infinite precision.
I would suggest it is a bit surprising.
What do you think? Should I look into writing a PR to change that or this behavior is intended?

ftomassetti:
Fix #92

Currently a number with decimal digits is assignable to number{inf} while an integer number is not.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/typesystem.mps
@@ -2439,6 +2439,59 @@
                   </node>
                 </node>
               </node>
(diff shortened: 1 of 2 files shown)


ftomassetti:
All tests passing but I think it would be a good idea to get a second opinion on this one

markusvoelter:
I really have no strong opinion here. If the tests pass, then nobody had one so far. So then we shoild merge.

PR #97, opened 2017-12-08
components.core: IBusInstanceAttribute added

arimer:
Currently there is no interface that allows easily to create attributes which should appear on bus instances. 
This interface can then be easily used for IAttributes, which should be applicable on bus instances.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
@@ -6744,5 +6744,35 @@
       <node concept="17QB3L" id="1yY6_Uj8$j$" role="3clF45" />
     </node>
   </node>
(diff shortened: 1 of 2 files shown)


joao-silveira:
This seems to be another of those problems which could be easily implemented with the "global" constraints library we discussed before.

I don't know any examples of an attribute that should implemented. and I might be missing something but:
-Why is the interface implemented in components.core, in opensource? It sounds weird to me because there is no such thing as a "Bus" implementation in open source. For me it sounds like this is at the wrong place.

-I also don't understand the 'canBeUsedUnder()' implementation. 

-Maybe for getting rid of yet another specific interface and just having (for now) an utility class implementing the constraint and called in the different attributes constraint aspect would be a solution more in line to the already discussed before.

arimer:
@joao-silveira you are right, I placed it in the wrong repo. Maybe the implementation is too specific for the framework at all.

PR #99, opened 2017-12-13
build: added missing org.iets.req devkit to org.iets3.req.os plugin

arimer:
Looks like you removed accidentially the devkit from packing in d26d5f1. 
I've added it back now.

### DIFF
--- a/code/languages/org.iets3.opensource/.mps/modules.xml
+++ b/code/languages/org.iets3.opensource/.mps/modules.xml
@@ -3,6 +3,7 @@
   <component name="MPSProject">
     <projectModules>
       <modulePath path="$PROJECT_DIR$/_spreferences/CodeReviewPreferences/module.msd" folder="_spreferences" />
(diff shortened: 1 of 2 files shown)


code/languages/org.iets3.opensource/.mps/modules.xml
```diff
@@ -3,6 +3,7 @@
   <component name="MPSProject">
     <projectModules>
       <modulePath path="$PROJECT_DIR$/_spreferences/CodeReviewPreferences/module.msd" folder="_spreferences" />
+      <modulePath path="$PROJECT_DIR$/devkits/org.iets3.req/org.iets3.req.devkit" folder="" />
```
coolya: can we put this in req folder so that all requirements stuff is a the same location in the project explorer?

arimer: sure. you mean the virtual "rec" folder?

arimer:
done.

PR #101, opened 2017-12-18
Refactor WireConnector to inherit from AbstractConnector

wsafonov:
Required adjustments as suggested in IETS3/iets3.core#180


coolya:
Could you explain when these methods would return null? Ans why this check shouldn't get executed in this cases?

coolya:
I'm closing this one and adding a new one with my proposed changes.

PR #104, opened 2017-12-19
Add AbstractConnectorBase as common super concept for all connectors

coolya:
comp: introduce base concept for all kind of connectors  …
In order to allow a more abstract representation of connectors for a new
connector `AbstractConnectorBase` was introduced. This concept assumes
that at least a source port (governing port) exists. The target for this
connector is still a abstract node. It is usefull in cases where a the
target of the connection is a bus or some other target that is not a port.

In additon the `AbstractConnector` was renamed to `AbstractPortToPortConnector`
It is a specialisation of the `AbstractConnectorBase` which asumes a that
the connection is made between two ports. Most of the behaviour methods
and interface implementations have been moved up to `AbstractConnectorBase`.

The `AbstractConnectorRefTarget` was changed to point to a `AbstractConnectorBase`.

Also the typesystem rules from `AbstractConnector` are moved up to the
`AbstractConnectorBase` except for the checking rule that checks the port
to port connection.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
@@ -1627,7 +1627,7 @@
                             </node>
                             <node concept="v3k3i" id="7Atos1yb5cG" role="2OqNvi">
                               <node concept="chp4Y" id="7Atos1yb5j6" role="v3oSu">
-                                <ref role="cht4Q" to="w9y2:mIQkxg5ZSA" resolve="AbstractConnector" />
+                                <ref role="cht4Q" to="w9y2:mIQkxg5ZSA" resolve="AbstractPortToPortConnector" />
                               </node>
                             </node>
                           </node>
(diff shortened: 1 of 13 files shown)


coolya:
I also removed the commit of @wsafonov that added the null checks because they are no longer required.

arimer:
Looks ok to me. Found only a small typo in  the description of AbstractPortToPortConnector. 
"base **conecpt** for all port to port connections"

coolya:
@wsafonov any opinions for you?

wsafonov:
The part with introducing the new AbstractConnectorBase looks good for me. As for the WireConnector I have several remarks:
- WireConnectorExprType: I think, it can be removed - we introduced it before only because there were no common base concept for WireConnector and the other connectors and hence a separate type rule was required. Having now AbstractConnectorBase in place, we can just use AbstractConnectortExprType for all. Actually there is now also no sense for having IConnectorExprType and it can be removed. @arimer what do you think? Btw, there is a typo in the concept name AbstractConnector**t**ExprType
- in WireConnector.getTarget() you return this.wireTarget. It should be actually this.wireTarget.target to return either the other port or the BusInstance.
- for the new method AbstractConnectorBase.getTarget(), when just looking at it, the semantics is not clear. May be there should be a comment on it saying that the target could be port or also some other element depending on the specific connector form.

coolya:
@wsafonov thanks for the feedback. I'm looking into the removal and adding some api docs at the moment.

coolya:
@wsafonov I have addressed all of your comments in open source and core. Is this good to get merged then?

PR #106, opened 2018-01-08
Allow kind aware scope calculation of ComponentInstances

arimer:
Currently it is not possible to restrict the scope of an <a href="http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F7804632404593436654&project=org.iets3.opensource">ComponentRef.ref <a/> to specific kinds. By default all components (of all kinds) are in scope of an `AbstractComponentInstance.`
and only an error is show, when for ex.  a `kindB` instance is created inside of a `kindA` component.
```
kindA component A_Comp1 {
instance [ B_Comp and A_Comp2  are in scope ]
}
kindA component A_Comp2 {}

kindB component B_Comp {}
```
The following implementation moves the calculation of all visible components, which are referenced from a `AbstractComponentInstance` to the `ComponentKind` itself. Every kind can then itself decide which components should be used in the scope calculation.

I know that introducing  some new methods to handle this "kind specific" scenario is not the proper solution. Furthermore this pullrequest should be used to discuss how the `ComponentKind `, `IKindSpecific`, `IKindContext` behaviour methods could be refactored to make the usage of kinds (scoping constraints and checking rules) more understandable. 

Opinions/Ideas/Critics are welcome!

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
@@ -8,7 +8,6 @@
     <use id="83888646-71ce-4f1c-9c53-c54016f6ad4f" name="jetbrains.mps.baseLanguage.collections" version="-1" />
     <use id="7866978e-a0f0-4cc7-81bc-4d213d9375e1" name="jetbrains.mps.lang.smodel" version="-1" />
     <use id="f2801650-65d5-424e-bb1b-463a8781b786" name="jetbrains.mps.baseLanguage.javadoc" version="-1" />
-    <use id="7a5dda62-9140-4668-ab76-d5ed1746f2b2" name="jetbrains.mps.lang.typesystem" version="1" />
     <devkit ref="fbc25dd2-5da4-483a-8b19-70928e1b62d7(jetbrains.mps.devkit.general-purpose)" />
   </languages>
   <imports>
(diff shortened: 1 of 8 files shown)


arimer:
After some internal discussions I refactored the scope calculation ComponentRef.ref. Now the kind of every potential component in scope is asked if the instances (with a kind of the component) can be used in the current context. 

So during scope calculation the `canBeInContext(kind)` of the reference target is asked. With this approach we can keep the previous usage of `canBeInContext(kind)` and instead of the the priviously  
<a href="http://127.0.0.1:63320/node?ref=r%3A769eaa92-d4cb-4fa9-87e4-269f7f35a1eb%28org.iets3.components.core.typesystem%29%2F7804632404594085398&project=iets3.opensource"> shown error </a> only the correct compontens, with the correct kind are in scope.

arimer:
1. Answer to your question
The initial idea of the default `canBeInContext()` implementation is not clear to me aswell. I think (!) the intention was to allow the user the usage of component instance of all component kinds, but show an error by default. From my point of view, we could simply return true in the default case. For **functional** the `canBeInContext() `impl. would need to be adapted, which is currently not the case

2. Answer to your question
"Wouldn't it make sense in the scope for ComponentRef.ref to delegate visibility into the component at first and then inside of the behaviour of the component delegate further into the **type**?"
---> you mean **kind** not type **right**?
But yes, we could do it. There is already a beh. method <a href="http://127.0.0.1:63320/node?ref=r%3Abe0c7a50-96d7-41ce-8522-0a6d4431fcc5%28org.iets3.components.core.behavior%29%2F3523817923956099791&project=iets3.opensource">canBeInComponentContent()</a> which is doing a similar job:
```
public boolean canBeInComponentContent(conceptNode<> concept) { 
  return this.kind.canBeInContent(concept); 
}
```
Question:

- So when I extend the language lets say with a new kind, the only thing I would need to do for configure the scope is, to override the `canBeInContext()` method, right?

arimer:
Just recognized that these changes will break scope calcualtion of ComponentInstances in mappings (on core).

arimer:
Closed due to internel design decisions.

PR #110, opened 2018-01-12
Allow kind aware scope calculation of Component Instances

arimer:
### Current Situation

Assume you have the following model: 
```
kindA component A_Comp1 {
--------------------substructure--------------------
    instance [ B_Comp and A_Comp2 are in scope ]
}

kindA component A_Comp2 {}

kindB component B_Comp {}
```
Currently when you create a component instance inside of a component substructure, every component regardless of its kind is in scope. There is only a generic error message that indicates that an instance where the kind is not the same as the parent component kind can not be used.
There is currently no possibility to restrict this scope.

### Target Situation
Customer language should be able to decide if “additional” constraints should be applied to restrict the scope of such component refs.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
@@ -65,6 +65,9 @@
         <property id="1225194472834" name="isAbstract" index="13i0iv" />
         <reference id="1225194472831" name="overriddenMethod" index="13i0hy" />
       </concept>
+      <concept id="1225194628440" name="jetbrains.mps.lang.behavior.structure.SuperNodeExpression" flags="nn" index="13iAh5">
+        <reference id="5299096511375896640" name="superConcept" index="3eA5LN" />
+      </concept>
       <concept id="1225194691553" name="jetbrains.mps.lang.behavior.structure.ThisNodeExpression" flags="nn" index="13iPFW" />
     </language>
     <language id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage">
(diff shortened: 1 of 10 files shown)


arimer:
### Solution
The Component.Ref should not be changed. It will still look for an  IVisibleElementsProvider to deliver a list of visible components for its scope. In addition the Component itself should be an IVisibleElementsProvider. It overrides the visibleContentsOfType() method and delegates the decision if elements list should be restricted to its kind. The ComponentKind shall have a method virtual boolean hasRestriction(), which returns by default false and a method  virtual boolean canBeReferencedInContext() which is true by default. So we don’t have any restrictions and allow any context-kind inside a “container”-kind.

### TODO

- Currently Component implements IFunctionContainer which itself is a IVisibleELementsProvider, we need to take this into consideration when overriding ListScope visibleContentsOfType(concept<> targetConcept) in Component
- Make visibleContentsOfType() in IVisibleElementsProvider virtual
-Create an opt-in method in ComponentKind to decide if the scope should be restricted. By default this method returns false
- Override visibleContentsOfType() in Component and delegate it to its kind
-For the implementation of the new method in ComponentKind that should decide if the given component can be referenced in the current component we should probably consider the decisions made during the design of canBeInContext() method - it also establishes a conditioned availability of kind-specific elements w.r.t to a context kind (however its implementation only affects if an inappropriate element would be marked with an error, but not removed from the scope or filtered out from the possible child elements). Kind of the kind-specific elements can control if the element is allowed under given context kind. The direction of the check for the referenced components should be probably made consistent with the context kind check

arimer:
https://github.com/IETS3/iets3.opensource/blob/feature/CustomizableKindRestrictions/doc/design/01-kindaware-scope.md

arimer:
@wsafonov you're right. Because the kind of the referenced component (from ComponentRef.ref) is asked if the parentkind `canBeReferencedInContext()` we dont need `hasRestrictions()`. 

The whole idea of the implementations is: 
`referencedComponent.kind.canBeReferencedInContext(parentComponent.kind)`

coolya:
While that is technically correct from readability and understanding the code standpoint I would rather have it two methods.

joao-silveira:
I think there is some stuff we can improve, in example the visibleContentsOfType impl, I would replace it with something simpler by in example using a FilteredScope.

in example:
``
Scope parentScope = super. visibleContentsOfType(concept) //gets all visible elements
if(hasRestriction){
     return new FilteringScope(parentsScope){//here implements exclude method}
}
return parentScope;
``                              
Just as an idea.

arimer:
Added the required changes and implemented scoping tests for newly created TestKinds.

PR #111, opened 2018-01-17
comp: add parameter dependencies for ports

coolya:
Includes a new port type that can wrap a IPortType. It can reference a parameter. When using these ports in connectors the type system ensures that only ports with same payload and configuration can be connected.

I removed the for "is statically evaluatable" from the checking rule of a component instance. Since that property of an expression in not maintained at all it will produce a lot of false positives.

This PR doesn't contain any of the static checking of the values. It is only the type system to specify the dependency from the port to the parameter.

WIP:
 - [x] Error message form the type system need to be better
 - [x] Typesystem for other connectors than `AssemblyConnector`s 
- [x] Testing this on OS is pretty cumbersome, I added the tests in core
 - [ ] ~~Restrict Port with configuration to only reference parameters~~
- [ ] ~~Restriction on how to pass values to a parameter~~
- [ ]  ~~allow to associate the configuration type on dataitem level~~
- [ ]  ~~check if ports that require a configuration actually have one~~
- [ ] ~~_optionally: add a error intention that adds the `with` clause and references the correct parameter (if possible)_~~

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/behavior.mps
@@ -7423,7 +7423,7 @@
   </node>
   <node concept="13h7C7" id="6P1C6lFyl17">
     <property role="3GE5qa" value="components" />
-    <ref role="13h7C2" to="w9y2:4kCIAUZDmGW" resolve="PortWithConfiguraion" />
+    <ref role="13h7C2" to="w9y2:4kCIAUZDmGW" resolve="PortWithConfiguration" />
     <node concept="13hLZK" id="6P1C6lFyl18" role="13h7CW">
       <node concept="3clFbS" id="6P1C6lFyl19" role="2VODD2" />
     </node>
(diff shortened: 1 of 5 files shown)


coolya:
As the checks which expression we can analyse to check the values statically I would rather implement theses checks in core when the analyses is required.

coolya:
I'm getting this back in to work in progress. We got some early user feedback on this one that will require us to do some more work. 

The feedback was that the overall notation is fine but they want to associate the configuration **type** already on the dataitem so that it is clear that the payload has have a configuration. The part where we assign the actual parameter is unaffected by this. 

Additional tasks to do: 
- [ ] allow to associate the configuration type on dataitem level
- [ ] check if ports that require a configuration actually have one
- [ ] optionally: add a error intention that adds the `with` clause and references the correct parameter (if possible)

I also added the relevant parts of the design document . This document doesn't include the latest customer feedback yet.

coolya:
Given the structure of the projects it makes no sense to implements these additional features in the opensource repo.

coolya:
Not sure why this strict ordering would be an issue as it's part of the type and not of the port.

PR #112, opened 2018-01-17
update to latest gradle

coolya:


sergej-koscejev:
You also need to actually run the wrapper task and commit the changed files (gradle-wrapper.properties).

coolya:
@szarnekow unfortunately you are correct it doesn't imply 4.4.1 and gradle doesn't support version specifiers here. So we have to add the patch version as well.

pngowda:
+1
looks good to me.

PR #115, opened 2018-02-06
Allow to control ManualChecks via IControlAdvancedFeatures

coolya:
I had to move IControlAdvancedFeatures into core to allow this. It requires a migration on the customer projects but no manual code changes. 

I also change the implementation in `ISolveable` to include the result from the super call into the result of `areManualCheckAvailable`

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
@@ -33,6 +33,7 @@
         <property id="1225194472834" name="isAbstract" index="13i0iv" />
         <reference id="1225194472831" name="overriddenMethod" index="13i0hy" />
       </concept>
+      <concept id="1225194628440" name="jetbrains.mps.lang.behavior.structure.SuperNodeExpression" flags="nn" index="13iAh5" />
       <concept id="1225194691553" name="jetbrains.mps.lang.behavior.structure.ThisNodeExpression" flags="nn" index="13iPFW" />
     </language>
     <language id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage">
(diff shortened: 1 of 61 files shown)


bj2015:
I focussed on the Solver-related parts of the Change.

As far as I understand, IControlAdvancedFeatures is a Replacement for IUseSolver that can toggle multiple different features and not just the solver.

In your change, you basically ANDed the two ways of toggling the Solver. Of course it makes much more sense to remove IUseSolver and adopt IControlAdvancedFeatures as the only way to toggle the solver, since it is the more general mechanism. However, this does not have to be part of your changes. I can take care of that.

As to your changes I have the following remarks:

1) the method-name "areRunManualChecksAllowed" sounds weird. How about "areManualChecksAllowed" or "isRunningManualChecksAllowed" instead?

2) The Solver should be opt-In, not opt-Out. In `IControlAdvancedFeatures.allowRunManualChecks()` you correctly set the default to "false", but in `IControlAdvancedFeatures.areRunManualChecksAllowed()` you used the logic

return r == null || r.allowRunManualChecks();

which implements optOut. Please replace this with

return r != null && r.allowRunManualChecks();

just like used in IUseSolver.

Otherwise, the change looks good. Good Job!

coolya:
@bj2015 thank you for the feedback.

`IControlAdvancedFeatures` is not a replacement for the opt in logic in your `ISolveabe` at least not from my understanding. It's use case is a different one. The language itself might support the solver and want's to use it but based on an arbitrary logic we want to show certain features to the user or not. This logic can include the role of the user for instance and might even change at runtime because the user is in a different point in the *workflow* of the model. It also allows us to have fine grained control over which features are allowed in certain root nodes. In one root we want to have the solver on other not based on the semantics in the customer specific project. 

`IControlAdvancedFeatures` is also totally optional, the containing root might implement it or not. That why it is implemented in an opt out fashion. 

For the name you are right I guess `areManualChecksAllowed` is much better.

PR #116, opened 2018-02-07
Introduce inline components instantiation

wsafonov:
Depending on the modeling context it is sometimes required to declare new components inside of other components and instantiate them at the same time, essentially allowing to define component instances and corresponding definitions in one place and one step. This is in contrast to creating a global component definition (e.g. in a chunk) and separately referring to it when defining its instances as content of other components.

To support such inline component definition and instantiation we introduce a new abstract concept `AbstractComponentInstanceBase` that would take the role of common abstraction layer for previous conventional `ComponentInstances` and the new `InlineComponentInstance`. The `AbstractComponentInstanceBase` encapsulates the origin of the actual component allowing the `InlineComponentInstance` to contain its own `Component` definition, and serves as an component instance interface for `Connectors`, that can be then connected with all implementations of the `AbstractComponentInstanceBase`.

**TODO**:

- [ ] clarify if we need to introduce a separate interface that would denote only component instances that are `ISubstructureContent` (`ComponentInstance` and `InlineComponentInstance`), to differentiate from those instances that are not (`SimbenchRootCI`).


wsafonov:
Simple example of the inline components usage is shown in the test [InlineComponents](http://127.0.0.1:63320/node?ref=r%3Aeec34ba1-5aa9-470e-9ddd-540a67f94f22%28test.ts.components.core.tests%40tests%29%2F8209493818909716176&project=org.iets3.opensource)

wsafonov:
- Yes, we need parameters for the inline components as well (TV and @kbirken mentioned this already)
- No, we don't have the requirements for the access to the outer component, but we can discuss this and introduce what needed
- What is exactly your suggestions for the editor and why it is not usable? This is actually implemented the straightforward way how TV suggested what they expect from the editor. Should we may be provide some outline box for the inline component to make it visually clearer?
- InstanceTree works exactly the same way with inline components. An inline components can be only accessed inside of the component where it is defined (it is more similar to java anonymous classes). Attributes can be assigned to Inline Components same as to conventional ComponentInstances
- I thought sticking to "Abstract-" prefix would be good to have a clear indication to an abstract concept. Otherwise we would have abstract AbstractComponentInstanceWithRef inheriting from abstract ComponentInstanceBase. But if this is fine for you, I'll rename it.
- So we would still need to adjust all uses of the ComponentInstance.component, that but we would not introduce another abstraction layer for the ComponentInstance - that would be the only simplification, right? Basically the whole inline component definition would just move to the new replacement of ComponentRef instead of InlineComponentInstance. This way we would say that every existing ComponentInstance could be potentially changed to use an inline component definition instead of the ComponentRef. It is here of course a question of the appropriate cut for the new abstraction layer. The only problem I see so far would be that we would be forced to use existing editor of the ComponentInstance also when having an inline component instead of a component reference. This would not fit for example due to possible `optionalName` that doesn't make sense for a inline component (where we already can pick up the name of the inline component to be unique in its parent component). Also the layout of the ComponentInstance editor with the summary editor for attributes in front of the component definition probably would not be very nice looking. In case of the current implementation I can define a better editor for the whole InlineComponentInstance replacing the ComponentInstance. Actually this was mainly my motivation with replacing ComponentInstance with a new InlineComponentInstance to get rid of those problems. But at least we can give that alternative a try if we think it would still a better approach.
- I don't know if we should move it to the commercial part. Could make sense if we say we want to keep opensource as simple as possible. Other opinions?
- Check the intention that I've already implemented for InlineComponentInstance - it already does exactly this ;)
- Don't think that adjusting to the new abstraction layer is a smell for the instance tree generator - basically we do this with all uses of the ComponentInstance where we think that the new abstraction would be more approriate.

wsafonov:
- We decided to not support passing parameters explicitly to the inline component - instead we allow to access parameters of the outer components directly. For this I have adjusted scope implementantion of the ParamRef to use IVisibleElementProvider and Component to implement IVisibleElementProvider.visibleContentOfType(). However, we don't disallow defining Parameters inside of the inline component - the user just needs to provide a default value for them (an error will be shown if this not the case). This can be helpful if user requires a computed value inside of an inline component that can be used as a parameter.
- The editor looks now almost like a component editor (w/o parameter values on the right), but I have additionally added brackets to visually set the inline components off from other component contents

PR #125, opened 2018-04-06
build scripts

coolya:
Simplified some of the parts of the builds scripts. Added a variable where people can simply put the mbeddr branch name they would like to build against. The script will take care of escaping the name etc. 

Also replaced some `println` with `logger.log` calls. 

Added the github release plugin to be able to push releases to github automatically.


sergej-koscejev:
Fixed version makes it easier to reproduce the build in the future if necessary. I would personally use the fixed version everywhere, I don't understand why it's useful to use the current version.

PR #144, opened 2018-11-06
build: updated to latest mbeddr platform

coolya:
- using mbeddr with version 1.1+ which was introduced during the grammar cells migration 
- reloaded build scripts to reflect the dependency changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ def forceLocal = project.hasProperty("forceLocalDependencies")
 // Dependency versions
 ext.mpsVersion = '2018.2.4'
 
-def mbeddrVersion = "1.0+"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -36,7 +36,7 @@ def forceLocal = project.hasProperty("forceLocalDependencies")
 // Dependency versions
 ext.mpsVersion = '2018.2.4'
 
-def mbeddrVersion = "1.0+"
+def mbeddrVersion = "1.1+"
```
vlsi: Do you really mean 1.1+ here?
Is it different from 1.1.+?

sergej-koscejev: 1.1.+ would be more correct because 1.1+ would match 1.10

Issue #142, opened 2018-10-01
Behaviour implementations missing or left behind


davidmpaz:
Hello,

after pulling latest updates from repo I did start getting these when building my language:

![selection_077](https://user-images.githubusercontent.com/210742/46298673-ee1ab600-c59f-11e8-9ccd-e03fbf5023dc.png)

Could someone take a look and check whether the methods should be removed or the interface added to those concepts? or vice versa.

Many thanks in advance,
David

alexanderpann:
Closing this ticket because it was fixed long ago.

PR #151, opened 2018-11-22
fixed some kernelf errors

coolya:
Fixed a bunch of model checking errors in kernelf. Most notably some missing behaviour methods. 

Most implementations where straight forward but at least for `GenericErrorType` someone should take a look. I'm not sure why this implements a wrapper type interface even if the types doesn't wrap anything.


coolya:
This also fixes #142

coolya:
> Is it necessary to
> 
> * add a dependency on `de.slisson.mps.tables.runtime` to `org.iets3.core.expr.util.editor`
> * add a dependency on `de.slisson.mps.tables.editor` to `org.iets3.core.expr.lookup.editor`
>   It seems that they are not used.
> 
Yes, MPS thinks they are unused. If I remove the dependency the modelchecker complains in the editor aspect that is needs a dependency to these module because of the table stylesheets used.

PR #158, opened 2018-12-14
build.gradle: enabled incremental build for languages

arimer:
The mentioned change enables the feature: `mps.generator.skipUnmodifiedModels` by default. 
This change reduces the local build time for the command line build extreamly because only the modified models are build locally and not all models/modules by default.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,7 @@ ext.dependencyRepositories = [
 ]
 
 ext.artifactsDir = new File(rootDir, 'artifacts')
-
+ext.incrementBuild = !project.hasProperty("disableIncrementBuild")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -94,7 +94,7 @@ ext.dependencyRepositories = [
 ]
 
 ext.artifactsDir = new File(rootDir, 'artifacts')
-
+ext.incrementBuild = !project.hasProperty("disableIncrementBuild")
```
sergej-koscejev: `increment` should be `incremental` here (both in the ext name and the project property name)

arimer: Ok

PR #161, opened 2018-12-26
Path elements can now be L-values

fvollmer:
The PathElement checks the member to evaluate wether it should be a L-value. Implementation is similar to the DotExpression.


fvollmer:
I just rebased my commits (this and #162). Please let me now if you need any more information.

PR #162, opened 2018-12-26
IRecordMembers can now be L-values

fvollmer:
This pull request actually depends on #161. It adds the canMemberBeLValue() function to the behaviour of the IRecordDeclaration. A IRecordMember can be a L-value if this functions returns true.

I'm happy to hear feedback regarding this pull request.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.path/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.path/models/behavior.mps
@@ -2,7 +2,7 @@
 <model ref="r:e876148b-672e-4264-9fee-d6d24a2d1223(org.iets3.core.expr.path.behavior)">
   <persistence version="9" />
   <languages>
-    <use id="af65afd8-f0dd-4942-87d9-63a55f2a9db1" name="jetbrains.mps.lang.behavior" version="2" />
+    <use id="af65afd8-f0dd-4942-87d9-63a55f2a9db1" name="jetbrains.mps.lang.behavior" version="-1" />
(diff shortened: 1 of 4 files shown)


markusvoelter:
I don't even think it should be on either. It should be on the RecordMemberAccess target (or whatever it's called), right? Or are you forwarding to the delcaration from the target?

sergej-koscejev:
@markusvoelter RecordMemberAccess is actually PathElement. Yes, it's forwarded from PathElement to the referenced IMember (in #161).

markusvoelter:
Then I am ok with it (sorry for asking the stupid questions. I currently don't have the bandwidth to actually look at the code).

sergej-koscejev:
I'm still not ok with the change, so I'd like to understand the problem better. Will wait until @fvollmer replies.

fvollmer:
Thanks for the review!
The reason behind this change was, that I implemented `IRecordDeclaration` in my own concept , but couldn't access the members of concept `RecordMember` as LValue:

    user.lastName = "foo"

I changed this by introducing a `canMemberBeLValue()` function for the `IRecordDeclaration` Interface. Since this function returns false by default the core language stays immutable. By overwriting this function the RecordMember can now be accessed via a PathElement as LValue.

The alternative is to also implement `IRecordMember` yourself and to use the constraints aspect to make sure that the `RecordMember` concept can't be used. Would you prefer this solution? Or what do you have in mind?

sergej-koscejev:
@fvollmer yes, I'm suggesting you implement your own IRecordMember and constraints. At least until another user has the same problem, then we'd think about a more general solution.

PR #168, opened 2019-02-08
text generation for python

ShivanandBiradar:


lhartl:
please follow the iets3 nameing conventions for the language and the solution.

ShivanandBiradar:
> please follow the iets3 naming conventions for the language and the solution.

Could you please let me know the conventions , so I can update accordingly.

MLanghammer:
Hi,

I checed the marks from Lisas review, which are (from my perspective) done. All others seem to be left open.
I have the following addtional remarks:

[ ] for the beginning you can change the name to org.iets3.py.lang.textgen. This should be fine.
[ ] The reduction of the base language concept, e.g., FloatingPointConstant is not necessary. Only the concepts of the iets3 languages need to be reduced.
[ ] for the logical expressions: we cannot use the alias of the concepts as "&&" and "||" need to be named "and" and "or" for python

Further remarks, which I do not consider as show stoppers:
- Not all Binary logical expressions are used in Python (for instance the LogicalImpliesExpression does not exists), but for the first text-gen this should be fine, as our future transformations should not create them anywhy.
- the implementation of the PyListExpr seems more complicated as it should be. In general, we should try to avoid so much "handwritten" java-code within the generator. Instead a SWITCH or trying to use COPY_SRC, would have been the better option.

ShivanandBiradar:
Michael could you please review the changes.

MLanghammer:
For me it looks good now.

lhartl:
outdated; not needed anymore.

PR #173, opened 2019-02-18
update mps

bkolb:


build.gradle
```diff
@@ -36,7 +36,7 @@ def major = "0"
 def minor = "2"
 
 // Dependency versions
-ext.mpsVersion = '2018.3'
+ext.mpsVersion = '2018.3+'
```
coolya: Same here. `2018.3+` matches `2018.4` as well. 
```suggestion
ext.mpsVersion = '2018.3.+'
```

PR #187, opened 2019-02-22
Fix empty DateLiteral

lhartl:
This fixes DATEV Issue 427. There was a Nullpointer Exception for empty DateLiterals.


lhartl:
Using the typesystem rule instead of the constraint was a usability issue from DATEV. With the constraint, if one deletes e.g. the year, the last digit will stay in the property, but one can not see it and the property is displayed red. The model checker does not show these constraint errors and the generator can be triggered with the deleted property only containing the last digit that was deleted. This would result in compilable but wrong generated code. Using typesystem rules, there is a typesystem error in the model if the property is deleted and the generator can not be executed (as long as the model checker is not swiched off).

I changed `toDate()` to return null if the Exception occurs.

sergej-koscejev:
FYI now the build doesn't work.

PR #207, opened 2019-04-07
Add parseInteger dot target for KernelF strings

kbirken:
This PR adds a parseInteger dot target to KernelF. Use it like this: `"123".parseInteger`.

It contains the actual concept, test cases and the Java generator part.


kbirken:
Today's commits address all review comments except the last. Regarding the generator, I had to mention the new concept in the condition because the generated code doesn't follow the DotExpression pattern (similar with StringLengthTarget). So every AbstractStringDotTarget with generated code which doesn't follow the DotExpression pattern has to be mentioned there.

PR #209, opened 2019-04-09
Escape the prime symbol ' for the java generator

fvollmer:


fvollmer:
@sergej-koscejev Thanks for the review! I applied the requested changes and rebased the branch on the current master.

PR #211, opened 2019-04-11
updated descriptors with MPS 2018.3.6

arimer:
Just ran the updates of the module descriptors


arimer:
Sure, I'll check it.

arimer:
You're right!

PR #215, opened 2019-04-16
Type System now tolerates the use of List equality

bj2015:
This is a fix for #442 in iets3.core.


kbirken:
After some research, it seems to me that PR #233 (which has been merged already) addresses the same issue as the PR at hand. Could #215 thus deleted, @bj2015 ?

What concerns me a bit is that the implementation which was merged to master with PR #233 is quite different from the implementation here... please comment.

bj2015:
Apparently, I solved this problem more than once. However, the version that PR #233 already merged to master is much more mature due to a very thorough review I would thus discard this PR and the branch `bugfix/generic-type-equals-problem#442`.

PR #221, opened 2019-05-02
Add generator rule for AndMatcher

sergej-koscejev:


lhartl:
Hi Sergej, I can do the review but not before monday.

lhartl:
Implementation is fine :)
You have written a Java generator and a Test for it, which is great. But the test is in a model for which no Java genplan is associated, so only the Interpreter Test will be executed on the CI. At the moment, the only model for which Interpreter and Java generation works is `test.in.expr.os.m1`. There is also a `records` Test Suite in it: http://127.0.0.1:63320/node?ref=r%3Acd8f023a-94dd-4e45-bbf0-a419cfafa530%28test.in.expr.os.m1%40tests%29%2F7740953487931216087&project=org.iets3.opensource. You might want to put your test there. The `todo` model is used to collect cases that do not work for the Java generator at the moment, but should, because they are part of languages for which a Java generator exists.

PR #222, opened 2019-05-02
Default Value for all datetime range types in KernelF

lhartl:
All defaults are set to the beginning of the UNIX time  (1.1.1970). Default ranges are according to Type one year or one month from this 1.1.1970)


lhartl:
Good Point. Defaults for DateDeltaTypes are not necessary at the Moment but might be in the future. When they are required, we can implement them.

PR #226, opened 2019-05-17
Use IArgument for the args of IFunctionLike

fvollmer:
grafik


sergej-koscejev:
This may break dependent projects, they will need to update their references to `Function.args`, if they have any. @markusvoelter are we ok with it?

@fvollmer you also need to update Function_Editor so that `args` points to the newly defined `args` in Function.

arimer:
@fvollmer were there any follow up commits after the findings from Sergej?
- Did you manage to implement your change backwards compatible to not break user projects?
- In case this changes is still relevant, please change the target branch of this PR to the corresponding matching maintenance branch and update your PR accordingly

arimer:
No development since 2019. Therefore closing. 

In case it is required please re-open a new PR.

PR #229, opened 2019-05-23
Java Generator for Tuple Assignments

lhartl:
I added the generator for tuple assignments and moved the corresponding Test from todo (interpreter only) to m1. Lodas requested this (Issue 564)


lhartl:
Failing swap test is fixed using a preprocessing script: 
If there is a TupleValue on the right side of an assignment, the tuple is stored in a new LocalVarDeclExpr. So the values do not get mixed up during swap.

lhartl:
Preprocessing for storing right side of Tuple Assignment in LocalVarDeclExpr is removed, because it ist not necessary anymore. Instead, it is done directly within the reduction rule.

PR #237, opened 2019-07-02
Feature/kernelf datetime genjava

lhartl:
With this PR, generator and interpreter tests are merged and the java generator is able to execute all interpreter tests successfully. 
- devkit for interppreter and java generator for the "advanced" kernelF stuff is created
- tests are merged and located in the interpreter tests
- tests are executed by both interpreter and java generator
- code that can be shared between generator and interpreter was found and relocated and is now reused (slight adaptations to interpreter meaning calling of some functions instead of having the code directly in the interpreter)


lhartl:
As you suggested @sergej-koscejev, I created a temporal.runtime solution and moved all classes from the language plugin to the runtime solution, that are used by the interpreter. Could you please have a look at it again?

PR #240, opened 2019-07-16
refactor gradle build and get rid of old ant build file

wsafonov:
**This PR migrates the current project to gradle 5.5.1**
The following adaptions are done:

- upgraded to latest mps-gradle-plugin version: 1.2.148.1eeeded
- use tasks from mps gradle plugin instead of old build orchestration ant script
- minor build script clean up / refactoring


arimer:
Thanks for contributing and refactoring away the old painfull ant script approach 🎉 ! 
I am glad to review them!

build.gradle
```diff
@@ -144,35 +156,69 @@ task resolveMpsArtifacts(type: Copy) {
...
+
+task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
+    script "$buildDir/iets3.opensource.tests/build.xml"
+    doLast {
+        ant.taskdef(name: 'junitreport',
+                classname: 'org.apache.tools.ant.taskdefs.optional.junit.XMLResultAggregator',
+                classpath: configurations.junitAnt.asPath)
+        ant.junitreport {
+            fileset(dir: "$buildDir", includes: '**/TEST*.xml')
+            report(format: 'frames', todir: "$buildDir/junitreport")
```
arimer: Why do you do this? Since when are we interresting in the report being part of the the test execution? Typically this can be easily configured on the CI server.

wsafonov: This was always part of the old ant build and I believe, makes sense to have to be able to quickly browse test results locally.

build.gradle
```diff
@@ -95,38 +107,38 @@ ext.dependencyRepositories = [
...
 ext.incrementalBuild = !project.hasProperty("disableIncrementalBuild")
-apply plugin: 'maven-publish'
-apply plugin: 'base'
 
 
-task wrapper(type: Wrapper) {
-    gradleVersion '4.4.1'
-    distributionType 'all'
+wrapper {
+    distributionType Wrapper.DistributionType.ALL
```
arimer: Is there a reason why did not encode the gradle version directly as part of the wrapper configuration? 
I think it is way easier to spot if we have `gr.adleVersion  '5.5.1'` in here. Its easier then digging out the gralde-wrapper.properties file

wsafonov: Well, may be it makes sense to keep it for documentation purposes only, because it's just a default value for the wrapper task - you can still install a different wrapper version by providing different version as command line argument w/o updating this value and gradle would not object.

build.gradle
```diff
@@ -144,35 +156,69 @@ task resolveMpsArtifacts(type: Copy) {
...
+ext["itemis.mps.gradle.ant.defaultScriptArgs"] = defaultScriptArgs.collect { "-D$it.key=$it.value" }
+ext["itemis.mps.gradle.ant.defaultScriptClasspath"] = defaultScriptClasspath
+
+task buildAllScripts(type: BuildLanguages, dependsOn: [resolveMps, resolveMpsArtifacts]) {
+    script "$buildDir/iets3.opensource.allScripts/build-allScripts.xml"
+    doLast {
+        // patch heap size setting for tests execution
+        def antTestScriptFile = rootProject.file('build/iets3.opensource.tests/build.xml')
+        antTestScriptFile.text = antTestScriptFile.text.replace("-Xmx1024m", "-Xmx2048m")
+    }
```
arimer: Why patching it on the fly? With this you will replace all kind of  values of `-Xmx1024m` in the entire file! If you want to batch file it that way, then make sure that you only change this value in this block:
```

    <junit showoutput="true" fork="true" haltonfailure="false" dir="${basedir}">
      <jvmarg value="-ea" />
      <jvmarg value="-Xss2048k" />
      <jvmarg value="-Xmx1024m" />
      <jvmarg value="-XX:+HeapDumpOnOutOfMemoryError" />
      <jvmarg value="-Didea.config.path=${build.mps.config.path}" />
      <jvmarg value="-Didea.system.path=${build.mps.system.path}" />
      
      <classpath>
        <fileset dir="${build.tmp}">
          <include name="mps.class.path.jar" />
        </fileset>
      </classpath>
      
      <env key="NO_FS_ROOTS_ACCESS_CHECK" value="true" />
      <sysproperty key="plugin.path" value="${mps.plugins.path.string}" />
      <sysproperty key="mps.libraries" value="${mps.libraries.path.string}" />
      <sysproperty key="mps.test.modules" value="${mps.tests.path.string}" />
      
      <sysproperty key="mps.macro.iets3.github.opensource.home" value="${iets3.github.opensource.home}" />
      
      <test name="jetbrains.mps.testbench.junit.suites.AntModuleTestSuite0" outfile="TEST-jetbrains.mps.testbench.junit.suites.AntModuleTestSuite0-tests" />
      <formatter type="xml" />
    </junit>
```

wsafonov: No, it's really only for those static values created by the build model generators - in the build model we don't use 1024m value anywhere (and definitely won't use as it's too low).

wsafonov: Upvote or better contribute a fix: https://youtrack.jetbrains.com/issue/MPS-27715

arimer: Didn't see that the patched heapsize was already there. If there is no other way to fix it, I'm fine with it.

arimer:
@wsafonov is there any migration to gradle 5.5.1 planed for https://github.com/IETS3/iets3.core/blob/master/build.gradle?

PR #241, opened 2019-07-19
KernelF GenJava: FromValue Method in Generated Code of EnumDeclarations

lhartl:
I added a `fromValue(...)` method to the Java Code that is Generated for EnumDeclarations in KernelF. This was a customer request. There are no additional language constructs for this, because it is only needed for the code that uses the generated code.

The Generated code was already reviewed by the IT of the customer.

Only one reduction rule was changed, so this is pretty quick to review.


lhartl:
Done :)

PR #242, opened 2019-07-23
Remove usages of Node_GetContainingLinkOperation

slisson:
It is removed in MPS 2019.2


sergej-koscejev:
Can use `n.link` instead of `n/.getContainingLink()`. This operation returns `SContainmentLink` and is not deprecated.

wsafonov:
we can drop this now, as the changes were already applied on the migration branch

PR #245, opened 2019-08-02
kernelF-genjava: fixed java generation for ValueAtOp

lhartl:
Just a small fix in two reduction rules:
- added Cast in generated code
- changed parameter for generated function call


sergej-koscejev:
Can you add a unit test for the bug?

lhartl:
The bug appears only in the context of lodas, it can not be reproduced with kernelF only. Within lodas, there is a unit test for the bug. I'll send the node URL as private message.

PR #250, opened 2019-08-16
Proper support of collections in test asserts

wsafonov:
This PR includes support of handling collections with numbers when executing asserts in test suites. Before EqualsTestOp in the assert only took care of comparing BigInteger and BigDecimal objects by wrapping BigInteger into BigDecimal and using corresponding `compareTo()` method in BigDecimal (instead of `equals()` method which could not handle BigInteger/BigDecimal). However in case of collections it simply delegated to the `equals()` method and therefore never correctly handled numbers in the collections, which could be semantically equal, but not according to the strict equals() implementation. For example:

```
val fraction = list(1 / 4)                                                                               
test case divide [fail] {                                                                                
  assert fraction equals list(0.25) [1 ms] [error] Test Failed. Expected: [0.25], but was: [0.2500000000]
}                                                                                                        
```                                                                                               
This fails because number precision after division doesn't equal precision of the number literal, even if the number is the same. Comparing these two parts w/o a list would work, because we would then apply custom equals logic from above, which ignores different precisions.

To handle this correctly EqualsHelper has been extended to compare collections (Lists and Sets) and apply custom equals logic to its elements.

This PR also includes another small extension: `IContextTypedOp` interface, which is an extension of `IDotTarget` that automatically derives its type from the context expression in `DotExpression`

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.adt/models/structure.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.adt/models/structure.mps
@@ -474,6 +474,9 @@
     <property role="TrG5h" value="Copy" />
     <property role="34LRSv" value="copy" />
     <ref role="1TJDcQ" node="7aipPVpLOlr" resolve="AlgebraicDotTarget" />
+    <node concept="PrWs8" id="6z5WYdPcNAn" role="PzmwI">
+      <ref role="PrY4T" to="hm2y:6z5WYdPcLnz" resolve="IContextTypedOp" />
+    </node>
   </node>
   <node concept="1TIwiD" id="28$LOSAMCgX">
     <property role="EcuMT" value="2460310434923316285" />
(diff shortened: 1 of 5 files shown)


markusvoelter:
... from a dependency perspective this works? Does the EqualsHelper see the Lists and Sets?

wsafonov:
> ... from a dependency perspective this works? Does the EqualsHelper see the Lists and Sets?

Don't quite understand, what do you mean. Why wouldn't it be able to see List and Set? Those are just usual java.util classes, no big difference to BigInteger/BigDecimal.

markusvoelter:
It might need access to the concepts ... but if not, good.

wsafonov:
> It might need access to the concepts ... but if not, good.

Ok, maybe for some other comparision logic. In this case I stay on the level of standard List/Set abstractions and only implement comparision logic according to their corresponding contracts.

PR #257, opened 2019-09-20
Fix BracketOP for IOrderedCollection

fvollmer:
The BracketOP editor works with any IOrderedCollection, but the typesystem fails with a runtime error if it isn't a list. To fix the typesystem rule IOrderedCollection is now inheriting from ICollectionType. Since ICollectionType extends IHasBaseType, we can use getBaseType() to fix the typesystem rule.


coolya:
replaced by #362

PR #265, opened 2019-10-01
Fixed Vertical Layout

markusvoelter:
https://gitlab-extern.itemis.de/datev/datev-steuer/issues/101


wsafonov:
@markusvoelter one case when it starts to look weird is when you use parenthesis:
image
I think, the closing parenthesis should be at least placed at the bottom in this case to not confuse the user. But this has to be fixed in the grammar cells properly. 

Or we say for the current heuristic - inside parenthesis we don't use vertical layout:
image

What do you think?

PR #281, opened 2019-10-21
take RuntimeErrorType into account when inferring type of BinaryExpression

lhartl:
ErrorType was not taken into account during the inference of the Binary Expression Type, which lead to errors when BinaryExpressions were used together with units.

Fix: Test operationType not only for null but also for Runtime ErrorType to get the right error messages.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/typesystem.mps
@@ -725,11 +725,26 @@
                                     </node>
                                   </node>
                                 </node>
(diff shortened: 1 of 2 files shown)


sergej-koscejev:
Can we have a test case for this?

lhartl:
There are two tests for this, one for `PlusExpr` for the common case and one for `EqualsExpr` for the special case. 
They are hidden in the unitsOnly typesystem tests because the problem only showed up using units. 
Node URL: http://127.0.0.1:63320/node?ref=r%3Aee5b1a89-4907-4bd7-bb79-ba99ef537bd3%28test.ts.expr.os.unitsonly%40tests%29%2F5962423428582352094

PR #282, opened 2019-10-22
kernelF: Error when MultiDecTab has multiple empty Conditions

lhartl:
MultiDecTabs can have empty conditions, which is the default. It does not make sense to have multiple defaults. This is now checked and an error appears if there are multiple empty conditions.


sergej-koscejev:
I'm going to merge this but I think it would be more appropriate for the errors to be placed on the default rows rather than the table itself.

PR #288, opened 2019-11-03
Feature/consistent intentions for tables

markusvoelter:
Beide Arten von Tabellen sollten sich konsistent per Intentions modifizieren lassen, wie in Excel: Add Row ABove, Add COlumn Left etc. Dies habe ich jetzt gebaut; bisher war das Chaos.


markusvoelter:
Regaridng your comment above: it is unfortunate that we do not get the intentions in this case; but the problem is that there is no cell at all, and the current node is just the dectab itself. No way to find out where we are, and thus where to insert new colums or rows.

PR #296, opened 2019-11-13
EffectiveContext method for KernelF IMessageNamespace

lhartl:
- effectiveContext() in IMessageNamespace;
- replaced all usages of IMessageNamespace. content with call of effectiveContext();


markusvoelter:
I have renamed the method.

PR #297, opened 2019-11-17
Feature/removal of intentions

markusvoelter:
Screenshot 2019-11-17 at 13 14 45


coolya:
Removing the intentions looks good to me, but can't we just delete them instead of just disabling them? Otherwise I guess they will remain in the code base forever.

markusvoelter:
I disabled them at first because I expect that somebody will start complaining that stuff is missing. I have it on my TODO list for about a month from now to delete them completely.

PR #298, opened 2019-11-21
components.core: removed IComponentsContent checking rule

joao-silveira:
If one inplements the dependencies() behaviour method,
this would imply that if your chunk content is public all
of its dependencies should be public aswell.

I think this depends quite a lot on the particular implementation
of the chunk content and therefore I remove this checking rule to
at least give a change to the particular implementation to decide
how to handle these cases.


coolya:
I agree I think this is highly specific to the semantics of the language and not a general assumption we should have.

PR #306, opened 2019-12-04
Fix Tagged Literals Editor

bkolb:
This commit fixes the editing behavior of tagged literals
in two different cases:
* 10C + 20C : if one tried to delete the C tag, one had
  to hit backspace 3 more times to delete the intermediate
  artifacts
* 10: if one hit C after the 10, the cursor got placed
  after the 10 and not after the C


szabta89:
I tried out the user experience, I think this is definitely an improvement. Looks good to me.

PR #311, opened 2019-12-13
Editor and tracer improvements re Darcula support

sergej-koscejev:


sergej-koscejev:
I just made them darker and checked that they look good with Darcula.

The replacement was supposed to be black -> foreground, white -> background. black -> background would be incorrect. I checked again and couldn't find the place where I did that, can you point me to it?

I only found this original code:
```
ll.setForeground(sel ? Color.white : Color.black);
```
replaced with
```
ll.setForeground(UIUtil.getTreeForeground(selected, focused));
```

and here the idea is that selection is supposed to be inverted, that's probably why it used white foreground (although background was also white - probably a bug). I replaced both with calls to UIUtil methods to get the correct colors.

Was this the code you had in mind?

arimer:
Thanks for clarifying! 👍

PR #313, opened 2019-12-17
Feature/improved error reporting

markusvoelter:
The error message output when tests failed sometimes did not report the little [C] that allows users to navigate to the source of the problem. I've made the code more robust to render the [C] in more cases.

Screenshot 2019-12-17 at 14 02 36


markusvoelter:
Das ist ja schon alles rot, incl. der Fehlermeldung.

PR #316, opened 2019-12-19
Fix caching of PrimitiveTypeMapper

### LINKED ISSUE #315: The current PrimitiveTypeMapper shouldn't be cached
The caching is problematic if you have a custom PrimitiveTypeMapper. This custom PrimitiveTypeMapper won't be loaded after it is build, which causes the first build to fail.

The caching is caused by:
grafik
http://127.0.0.1:63320/node?ref=r%3Aac28053f-2041-47f6-806b-ecfaca05a64a%28org.iets3.core.expr.base.runtime.runtime%29%2F3281846772294987681

[The official documentation](https://www.jetbrains.com/help/mps/extension-support.html) recommends to not cache them:
> Objects returned by the extensions have transient nature: they may become obsolete as soon as a module reloading event happens. It is not recommended to e.g. cache these objects. Instead is it better to get a fresh copy each time.

fvollmer:
Objects returned by the extensions have transient nature: they may become obsolete as soon as a module reloading event happens. It is not recommended to e.g. cache these objects. Instead is it better to get a fresh copy each time.
(https://www.jetbrains.com/help/mps/extension-support.html)

This fixes #315.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/plugin.mps
@@ -6,6 +6,7 @@
     <use id="7a5dda62-9140-4668-ab76-d5ed1746f2b2" name="jetbrains.mps.lang.typesystem" version="5" />
     <use id="c7d5b9dd-a05f-4be2-bc73-f2e16994cc67" name="jetbrains.mps.baseLanguage.lightweightdsl" version="-1" />
     <use id="c0080a47-7e37-4558-bee9-9ae18e690549" name="jetbrains.mps.lang.extension" version="-1" />
+    <use id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage" version="9" />
     <devkit ref="fbc25dd2-5da4-483a-8b19-70928e1b62d7(jetbrains.mps.devkit.general-purpose)" />
   </languages>
   <imports>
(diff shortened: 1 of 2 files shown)


fvollmer:
This patch uses activate() and deactivate() handlers of the extension to avoid instantiating the PrimitiveTypeMapper all the time. I think this should be the better solution.
grafik

fvollmer:
I'm not sure why the last build failed and decided to rebase on top of the current master.

@sergej-koscejev would you mind to trigger the automatic build, again?

sergej-koscejev:
@coolya I've tried restarting this build a few times and it always fails with some timeout getting artifacts from Nexus. Can you have a look?

sergej-koscejev:
Now I'm not sure of the performance implications of this PR. We call PTF quite often and if it would involve going through a list every time I'm afraid it would affect performance significantly. Perhaps the extensions should call a method on PTF to add/remove mappers to a list, and PTF itself would choose the best mapper when the list changes.

coolya:
How about implementing `DeployListener` inside of PTF and invalidate the cache on any load/unload event? PTF would need to move to a singleton pattern than because we would need an instance to register the listener but this could be dealt with behind the scenes while we keep the static API.

sergej-koscejev:
> How about implementing `DeployListener` inside of PTF...

Yes, that would probably also work.

fvollmer:
Thanks for the feedback. I see you point. I'll give the `DeployListener` idea a try.

fvollmer:
I gave the `DeployListener` Idea a try, but could get it working. So I tried @sergej-koscejev suggestion and it works fine. I pushed these change to the branch of this PR.

grafik

---
Details regarding the not working `DeployListener`  implementation (https://github.com/fvollmer/iets3.opensource/tree/fixPrimitiveTypeMapperCacheV2):
 * Added an application plugin to org.iets3.core.plugin to register a custom DeployListener, which invalidates the PTF cache
* Add cache invalidation function to PTF

grafik

But it isn't working. I also tried to make the PTF functions synchronous, to avoid potential thread problems, but it isn't changing anything.

fvollmer:
Just forgot to mention: This implementation should be backward compatible and should have the same performance as the original implementation. 
It just provides a way to invalidate the mapper cache, which should be called any typemapper extension. If you don't it just behaves the same is before.

sergej-koscejev:
It's not backwards compatible in that all extensions have to invalidate cache on activation/deactivation so they will have to be changed. I think it's acceptable though if the DeployListener doesn't work.

That said, what do you mean when you say that DeployListener isn't working? Is it not getting called or what?

fvollmer:
@sergej-koscejev I think it is kind of backward compatible. The behaviour without calling the `invalidateCache()` is the same as before. So the `invalidateCache()` is like a new "feature" ;)

Regarding the DeployListener:
I think the DeployListener wasn't called. My typesystem language wouldn't load during a build and the build would fail.
If you have any idea what could cause this, let me know.

sergej-koscejev:
Hmm, yes, you're right. I think we can merge it then.

PR #320, opened 2020-01-08
Feature/picking

markusvoelter:


lhartl:
Looks good to me, there are even generators!!!

PR #321, opened 2020-01-10
framework for allowing dots on the left

markusvoelter:


wsafonov:
Looks good to me in general. Just couple of things:
- can we also have some simple examples for both new interfaces, so everyone can also try it out right away?
- in `dotOnLeft` menu contribution, we should probably also check in can execute, that node to be replaced is instance of `Expression`, because although we expect it, it's not guaranteed just by the interface

wsafonov:
Merged from `datev-staging` branch to `maintenance/mps20192`

PR #336, opened 2020-02-27
Datev staging

markusvoelter:


wsafonov:
I have reviewed the code in general and fixed some minor findings, but haven't really looked much into the logic. Existing code should not be affected by the changes - only new features and bugfixes included.

Issue #349, opened 2020-03-18
Type calculation has errors if units are involved


MLanghammer:
Given the following if expression as example:
![grafik](https://user-images.githubusercontent.com/5717036/76944761-db19b400-6901-11ea-80a4-99ed67a192a7.png)
The calculated type of the if-expression results in:
![join](https://user-images.githubusercontent.com/5717036/76944840-fdabcd00-6901-11ea-9637-f57df275a410.PNG)
It should, however, be number[0|1.5]{1}s. If the units are not used the calculation works.

PR #350, opened 2020-03-18
issue #349: Fix type calculation for expressions with units

MLanghammer:
The PR  fixes type calculation for if expression, match expression, try expression and alternative expression if units are involved.


joao-silveira:
Offline we discussed the possibility that this could be fixed by implementing a proper TaggedType subtyping rule. This would cause that if both expressions (if and else) share a common supper type this will be found by the if expression inference rule. In other case a jointype is inferred which is anyway (presumably) the correct thing.

PR #355, opened 2020-03-31
updated to MPS 2019.3.4

wsafonov:


build.gradle
```diff
@@ -61,7 +61,7 @@ def major = "2019"
 def minor = "3"
 
 // Dependency versions
-ext.mpsVersion = '2019.3.3'
```
coolya, line 64: Why don't we use `2019.3.+`? This was how we did it back in 2019.2 times.

wsafonov: Generally I'm not against it. It's just because sometimes bugfix releases were not truly just runtime releases and some updates were required in the code (even if minor). And sometimes we even experienced issues with new bugfix releases, so it's a more controlled/traceable way of switching to a new version than just suddenly having it there in some of the builds (you couldn't otherwise easily see from which build the new version has been used).

But if we have a consensus about that and agree to deal with this, we can change to automatically using the bugfix versions. Then whoever updates the MPS should probably check that all downstream projects are still ok. Also we should actually then restrict the circle of people who can trigger MPS updates - right now anyone who can push to the `build.publish.mps` repo initiates an update.

PR #358, opened 2020-04-01
Feature/improve unit conversion typesystem

MLanghammer:
The PR fixes the following issues related with units and their type: 
- the type of ConvertExpression is now calculated correctly. For instance if one converts seconds to milliseconds the calculated type will consider the conversion rule and calculate the type accordingly. (see http://127.0.0.1:63320/node?ref=r%3Aee5b1a89-4907-4bd7-bb79-ba99ef537bd3%28test.ts.expr.os.unitsonly%40tests%29%2F4240468146474310185 for an example)
- introduce converTo Dot target, which allows for using the conversion in dot-expressions. (see http://127.0.0.1:63320/node?ref=r%3Aee5b1a89-4907-4bd7-bb79-ba99ef537bd3%28test.ts.expr.os.unitsonly%40tests%29%2F6132826577343931657 for an example)
- fix error on UnaryMinusExpression if it used together with units (see http://127.0.0.1:63320/node?ref=r%3Aee5b1a89-4907-4bd7-bb79-ba99ef537bd3%28test.ts.expr.os.unitsonly%40tests%29%2F8208891105593573887 )

### DIFF
--- a/code/languages/org.iets3.opensource/.mps/modules.xml
+++ b/code/languages/org.iets3.opensource/.mps/modules.xml
@@ -116,6 +116,7 @@
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.tests.rt/org.iets3.core.expr.tests.rt.msd" folder="expr" />
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.toplevel.interpreter/org.iets3.core.expr.toplevel.interpreter.msd" folder="expr.lang-core" />
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.typetags.lib.interpreter/org.iets3.core.expr.typetags.lib.interpreter.msd" folder="expr.lang-advanced" />
+      <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.typetags.units.interpreter/org.iets3.core.expr.typetags.units.interpreter.msd" folder="expr.lang-advanced" />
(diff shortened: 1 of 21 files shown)


bkolb:
- The convertTo Dot target looks very verbose. Can we put the reference to the conversion also into the inspector, esp, if there is only one rule anyway
- The color in the convertTo Dot target for the unit is wrong
- The convertTo Dot target should be blue as well
- I don't understand the changes in the ValExpr inference rule. Why do you check for the parent expression and modify the range? Where is that information used? What if any other expression than +- / */ is used?
- I had to dig into the code to understand what "eager" means. Is there a better word for that?

MLanghammer:
> 
> 
>     * The convertTo Dot target looks very verbose. Can we put the reference to the conversion also into the inspector, esp, if there is only one rule anyway
good idea
> 
>     * The color in the convertTo Dot target for the unit is wrong
> 
fixed
>     * The convertTo Dot target should be blue as well
> 
fixed
>     * I don't understand the changes in the ValExpr inference rule. Why do you check for the parent expression and modify the range? Where is that information used? What if any other expression than +- / */ is used?
> 
It improves the type of the conversion specifier by calculating a more precise range for +- / */. If any other expression is used than numberType is used without a range.
>     * I had to dig into the code to understand what "eager" means. Is there a better word for that?
As we discussed, a better name (such as unit-aware) should be used. I would, however, prefer to not fix this in this PR.

PR #361, opened 2020-04-06
Allow to add a unit to non-literal expressions

bkolb:
Example is pow(10,-5). It was not possible to add a unit to this expression.
This commit fixes this.


markusvoelter:
I think a unit only makes sense on literals. Or has anybody ever seen a physical formula with "x km/h"? I have not.

mgronover:
Units of expressions should always derived from literals used in that expression.
The only usage here would be to add further redundancy to check that an expression is convertible into a expected unit.
Do we really need that?

arimer:
Not required.

PR #362, opened 2020-04-09
Fix BracketOP for IOrderedCollection

coolya:
The BracketOP editor worked with any IOrderedCollection, but the typesystem would fail with a runtime error if it wasn't a list. To fix the typesystem rule IOrderedCollection is now inheriting from ICollectionType. Since ICollectionType extends IHasBaseType, we can use getBaseType() to fix the typesystem rule.

This replaces #257


coolya:
@fvollmer I rebased your PR on current master and open a new PR because I couldn't push into your PR.

fvollmer:
Ok, the PR was already pretty old and I haven't looked at it for quiete some time.

coolya:
Sorry for huge delay @fvollmer I'm looking through all PRs right now and try to get them merged if possible.

PR #363, opened 2020-04-09
Feature/create additional math expressions#360

dbinkele:


wsafonov:
Let's synchronize with Markus regarding the trigonometric functions - if feasible, he could base his work on this PR and just clean up by adding required changes.

PR #369, opened 2020-04-21
get rid of warnings in languages

wsafonov:
cleaned up:
- empty generators
- superfluous/missing extended languagues


wsafonov:
FYI: remaining warnings on devkits relate to this bug: https://youtrack.jetbrains.com/issue/MPS-32024

PR #374, opened 2020-05-26
Provide mechanism to replace port/connection checking rules

kbirken:
The component model language has a hard-coded checking rule which checks the ports and component instances in a component's substructure. It especially checks that the connections of a port are consistent with the port's cardinality specification.

However, in some usecases this is too restrictive. E.g., when attaching presence conditions to connections (150% model for variability), more relaxed checks are needed. 

This PR establishes a mechanism which allows to replace the checking rule by a more specific one provided by a language extension.


kbirken:
See branch feature/fm_comp_model_variability of iets3.core for an implementation of the above usecase.

PR #377, opened 2020-06-08
Fix wrong duplicate/triplicate scoping for components

kbirken:
Scopes for the Component concept may have duplicate or triple entries - this PR fixes this.


kbirken:
Sounds good at first, but won't work in general: The clients of parentScope (i.e., the caller of Component.visibleContentsOfType) might use getScopes() and then will still get the duplicates...

wsafonov:
> The clients of parentScope (i.e., the caller of Component.visibleContentsOfType) might use getScopes()

That would be only possible if a client would cast `Scope` to `CompositeScope` to retrieve individual scopes. But why would any client want to do that? That's like trying to break into implemetation details. Sounds anyway not like a good idea and right approach from the client point of view.

kbirken:
Ah, didn't see that. Implemented it according to your proposal, see latest commit.

PR #381, opened 2020-06-30
the old getIdentifierName() returned after each "fix", whcih was wron…

markusvoelter:
…g if a name had a paragraph and an umlaut, for example. Now I replace all.


markusvoelter:
Invalid.

PR #382, opened 2020-07-03
expr.util: fix model errors

sergej-koscejev:


wsafonov:
Just couple of remarks from my side:

- Is there already an issue for the grammarcells issue (BaseConcept contribution)? If yes, I'd prefer to reference it in the text in `DummyForGrammarCells` editor so we know later that it can be removed.
- there is an empty `DecTabContent` contraints aspects. Is that left there on purpose?

sergej-koscejev:
I'm not aware of the issue regarding `DummyForGrammarCells`, maybe @slisson knows?

The empty `DecTabContent_Constraints` was a mistake, I removed it now.

slisson:
No, there is no issue in the JetBrains issue tracker.

sergej-koscejev:
It turns out I don't need to add DummyForGrammarCells components anymore as they were already added on master in the meantime.

PR #388, opened 2020-08-06
kernelf: small adaptions

lhartl:
- BuilderAdapter method missingSetters() is now virtual
- org.iets3.core.expr.genjava.messages.rt.rt.Message now implements Serializable


lhartl:
Thank you @sergej-koscejev . 
I added the `Serializable` interface to all member types used in `Message` and checked that their members again are serializable. 
The Set and Map interfaces themselves are not serializable, but their implementations used in `Message` are.

PR #391, opened 2020-08-20
Fix #390 : Accurate matching and precision for number literals with scientific notation

### LINKED ISSUE #390: Number scientific notation requires . in number literal
Currently the number literal (simpleTypes) requires a . in the string to allow the use of scientific notation (XXe-/+XX)

For example : 10e5, 12345e-1 would not match but 10.e5 or 123.45e-1 would be just fine.

I think the regular expression of real number causes this (below code, in NumberLiteral_Behavior)

image

Is it an expected behavior?

banilaste:
- Real values in number literals now allows for expressions such as 3e5 or 4e-4 (without specifying a dot in the left part)
- The precision computed from the string now take magnitude into consideration
  - Using values such as `1.00e-11` previously caused it to become zero with java generated expressions, because the computed type had a precision of 6 in this example (6 chars after '.'). The generated code for such number was `new BigDecimal("1.00E-11").setScale(6, RoundingMode.DOWN)`.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.simpleTypes/models/behavior.mps
@@ -203,6 +203,10 @@
         <child id="1163668934364" name="ifFalse" index="3K4GZi" />
       </concept>
       <concept id="1082113931046" name="jetbrains.mps.baseLanguage.structure.ContinueStatement" flags="nn" index="3N13vt" />
+      <concept id="1208890769693" name="jetbrains.mps.baseLanguage.structure.ArrayLengthOperation" flags="nn" index="1Rwk04" />
+      <concept id="6329021646629104954" name="jetbrains.mps.baseLanguage.structure.SingleLineComment" flags="nn" index="3SKdUt">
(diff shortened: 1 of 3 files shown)


markusvoelter (commented):
We have built the branch on the server, green. Code looks good. But can you please add one or two tests that nail down the behavior?

banilaste:
> 
> 
> We have built the branch on the server, green. Code looks good. But can you please add one or two tests that nail down the behavior?

I have just committed some unit tests (test.in.expr.os.numbers > numbers@tests > numbers, screenshot below), are these what you were expecting? 

These tests does not seem to test the behavior of the java generated code, is there any place this should be done? (as the precision issue occurred only on generation due to a different implementation in the interpreter)

image

markusvoelter:
Yes, this is nice. You could also add a type system test that verifies that type checking still works.

Wrt. to the Java generator, did you modify the generator? I don't think so, right?

banilaste:
Okay, I just committed some type system tests. 

For the java generator, I did not modify it, I believe the issue was mainly with the computed type upon generation (which is not used in the interpreter implementation on the right of the screenshot below), not with the generator.
image

markusvoelter (approved):
LGTM now.

markusvoelter:
Merged. Thanks for your contribution, @banilaste :-)

PR #392, opened 2020-08-27
Fix error about GlobalIconManager being a service, not a component

sergej-koscejev:
Retrieve GlobalIconManager using `GlobalIconManager#getInstance()` rather than `Application#getComponent()`.


coolya (approved):
LGTM

PR #394, opened 2020-09-08
kernelf: fix firstN, lanstN cast exception in Generator

lhartl:
Generating a expression like `list.first(5).last(2)` lead to a cast exception due to a mix of `PVector`s and `TreePVector`s in the generated code. This is fixed with this PR.


sergej-koscejev:
I'm going to approve the pull request because it does what it says, but I see two more differences in firstN and lastN semantics between the interpreter and the generator:
* `zahlen.firstN(100)` is equal to `zahlen` in the interpreter but the generated code throws an `IndexOutOfBoundsException`. Same for `lastN`.
* The generator rules for `firstN` and `lastN` evaluate the context expression (left-hand side of the dot operator) multiple times. If it's a complex expression, it will be inserted multiple times in the generated code, complicating the code and doing unnecessary work. It should be assigned to a temporary variable first, then used by the rest of the generated code.

sergej-koscejev (approved):

PR #398, opened 2020-09-17
Use Objects.equals() in (Not)EqualsExpr genenerated Code

lhartl:
Replace object equals() method with null-safe Objects.equals() in (Not)EqualsExpr genenerated Code.


sergej-koscejev (approved):
The change in `ComparisonHelper` is unnecessary since `foo instanceof string` implies `foo != null`. But it's harmless.

PR #412, opened 2020-11-09
Generalize solver task factory: Find factory also for super-concepts

kbirken:
The current `STF` helper class finds a `SolverTaskFactory` based on a node's concept. However, it cannot find a `SolverTaskFactory` for a base concept of that node's concept or for an interface implemented by the node's concept.

This PR  generalizes the `STF` class in order to do a breadth-first search on the direct super-concepts (and interfaces) of a concept. It starts with the actual node's concept and moves up the hierarchy until it finds a `SolverTaskFactory`. There is also a cycle check in order to avoid infinite loops.


wsafonov (approved):
LGTM. I think, one could alternatively just use `super-concepts/all` - it already iterates through the complete inheritence tree by combining super-concepts/direct (that is also breadth-first) and it would already return list without duplicates, hence eliminating the need for the additional visited check.

kbirken:
> LGTM. I think, one could alternatively just use `super-concepts/all` - it already iterates through the complete inheritence tree by combining super-concepts/direct (that is also breadth-first) and it would already return list without duplicates, hence eliminating the need for the additional visited check.

Thanks, I will check this: simpler is better! Although I've been so proud of my nice algorithm :-)

wsafonov:
> Thanks, I will check this: simpler is better! Although I've been so proud of my nice algorithm :-)

You can still be proud of you that you wrote the same algorithm that others already did without looking at it first 😄

Btw, I didn't mentioned it, but there was even more optimization possible by using the self parameter: `super-concepts/all<+>` - then you don't need to check the node concept separately, as it is already included.

PR #414, opened 2020-11-17
Allow solver check to optionally show subresults

kbirken:
The interface `ICanRunCheckManually` provides a checking rule which converts check results (e.g., from a Z3 solver run) into MPS typesystem errors and warnings. The check results to be converted have to implement the `IResult` interface.

This PR introduces a new interface `IHasSubResults` which can be used to attach additional results to an `IResult`. Via the behavior method `ICanRunCheckManually.showSubResults` (default: false) the subresults of an `IResult` can be converted to MPS typesystem errors and warnings during the conversion of the actual result (aka *main* result).

The sub-results mechanism is already being used by the Z3 solver integration of iets3.core, but there was no possibility to show more than one result in an editor up to now.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
@@ -911,6 +911,18 @@
         </node>
       </node>
     </node>
+    <node concept="13i0hz" id="4MH81Y0VldB" role="13h7CS">
+      <property role="13i0iv" value="false" />
+      <property role="13i0it" value="true" />
+      <property role="TrG5h" value="showSubResults" />
+      <node concept="3Tm1VV" id="4MH81Y0VldC" role="1B3o_S" />
+      <node concept="10P_77" id="4MH81Y0VldD" role="3clF45" />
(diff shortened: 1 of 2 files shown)


dbinkele:
@kbirken refering to this code-snippet from check_ICanRunCheckManually:
image

Wouldn't it suffice to check for IHasSubResults and then simply call .getSubResults()? There could be a default implementation which returns an empty sequence. So any time I want to display subresults I override .getSubResults()! And if not I do nothing. This was the flag  .showSubResults() is redundant.....one step further: why not moving .getSubResults() to ICanRunCheckManually_Behavior?

kbirken:
> Wouldn't it suffice to check for IHasSubResults and then simply call .getSubResults()? There could be a default implementation which returns an empty sequence. So any time I want to display subresults I override .getSubResults()! And if not I do nothing. This was the flag .showSubResults() is redundant.....

@dbinkele The additional flag `showSubResults()` allows to decide against showing them even if the results concept implements IHasSubResults and returns some. This is why I do not want to rely on available subResults only.

> ...one step further: why not moving .getSubResults() to IResult?

I didn't want to change the IResult interface, which is quite commonly used. Instead I wanted to keep the extension separate.

dbinkele (approved):

PR #418, opened 2020-11-24
Fix bug in editor update after solver has been executed

kbirken:
Action `runManuallyOnNode` executes an `ICanRunCheckManually` node and updates the editor afterwards. The UI updating only worked if the `ICanRunCheckManually` is the root node of the editor to be updated.

With this PR, this also works if the `ICanRunCheckManually` is a node attribute attached to the root node.

The PR also resolves some minor typos and removes debug outputs.


dbinkele (approved):
With the fix at least all editor cells will be updated as before (and some more).

PR #420, opened 2020-11-27
units: add quantities

coolya:
Replaces the "category" string on the unit definition with the first
class concept "quantity". Units can optionally reference quantities
in their definition. A migration takes care of migrating existing data
to the new concept. During the migration a new quantity is created for
each unique category in the same library, no merging of categories into
the same unit is done across root nodes / libraries because migrations
have no guaranteed order.

While the quantity itself is in the existing units language a new
language org.iets3.core.expr.typetags.units.quantity is introduced which
holds additional features. The new language allows using quantities as
type tags, similar to units. The type system of that new language also
enabled covariance between units and their quantity. For example a
function accepting a number of the quantity "time" is allowed to take
concrete values of type seconds, hours, etc:

    fun calcSpeed(d: number<metre>, t: number<time>) = d / t
    val s = calcSpeed(100 m, 10 s)

The type system support is currently limited to plain quantities which
doesn't allow to express things like `speed = metre / time`. Right now
the default behavior of the kernelF type system is used in this case
which default to the left operants type for binary expressions. In
addition converting variables of a given quantity to a unit via
conversions is not possible.

No generators or interpreters are implemented to these new concepts at
the moment.


joao-silveira (approved):
Looks good to me! I just commited two small changes:
    * removed a forgotten println
    * Remove optional unit specification child in quantities (I think it was there by mistake)

PR #424, opened 2020-12-16
Units: make units migration rerunable

### LINKED ISSUE #423: Unwanted changes by re-runnable migration for units
If I execute re-runnable migrations, there are new changes generated for Units:
image

It looks like (some of) existing quantities are duplicated and references to existing quatities are changed.

coolya:
A bug in the units migration didn't set the now deprecated "description" property to null therefor
every execution of the rerunable migration would create new quantities. The migration was updated
to detect this case properly and not change the model in this case. In addition the migration now
sets the deprecated property to null.

fixes #423


wsafonov:
@coolya I have quickly checked how the migration behaves with your changes, but I still see those duplications of existing quatity elements. Did it actually worked for you?

coolya:
I'm just stupid, when I cleaned up my code I replaced a `isNull` with `isNotNull`. Please try again.

wsafonov (approved):
That looks now better. Thanks!

I've pushed the changes after re-executing the migration (description property is properly removed)

PR #425, opened 2021-01-05
Bugfix kernelf records equals

lhartl:
Enables Applying Equals Expression to IRecordTypes
Interpreter could already handle this case, typesystem and generators are adapted with this PR


wsafonov (approved):
LGTM

lhartl:
thank you for the review @wsafonov

PR #426, opened 2021-01-21
Build full distro with Github release

wsafonov:
This PR adds support for creating a full distribution package with all language dependencies which can be published as a github release.


coolya (approved):
LGTM

PR #430, opened 2021-01-24
Migrate/mps2020 3

wsafonov:
This PR migrates project to MPS 2020.3.

Summary of the changes required during the migration:
- API changes in the mps.build infrastructure resulted in required refactoring of the `RunKernelFGeneratedTests` concept [behaviour method](http://127.0.0.1:63320/node?ref=r%3A8f2b5e7d-bab3-409e-93c3-6055bdcd2a68%28org.iets3.opensource.build.gentests.behavior%29%2F8236504404382684560) that retrieves `gencontext` and `DependenciesHelper` from the supplied `RequiredDependenciesBuilder`. With parts of the mps.build implementation becoming non-public the current implementation had to be refactored and the custom generator templates had to be adjusted (s. 60202875) to enable proper integration of the `RunKernelFGeneratedTests` build aspect within MPS 2020.3.
- ~due to a bug in `CellMenuPart_ReplaceChild_Group` generator [(MPS-32979)](https://youtrack.jetbrains.com/issue/MPS-32979) parameter object had to be explicitly casted to its type in [`mainAttributes`](http://127.0.0.1:63320/node?ref=r%3A1ee7d695-67a0-4310-b6a9-28ed5f54bc88%28org.iets3.core.attributes.editor%29%2F6173182635676059271) editor component~
- ~a couple of commented definitions in a [TestSuite](http://127.0.0.1:63320/node?ref=r%3Afcb91210-c6db-4de0-81c0-ca99e48fd25a%28test.in.expr.os.records%40tests%29%2F8721870733845863327) had to be uncommented as this led to a generation failure. In previous MPS version there was only a log error reported at this place, so it seems that with MPS 2020.3 some generation checks have become more strict.~
- `mps.ant.log` property in the generated ant scripts gets automatic value based on the gradle log level thus allowing to control the level of logging in generated ant scripts with `-i` and `-d` gradle flags (with the new default `warn` level there is not output during the generation which modules are currently being generated which makes it a bit harder to analyse the generation build log in case of generation problems).
- old `test.ts.core.expr` solution that still existed in the repo and was referenced in the build script, but had been removed from the MPS project long time ago during repo restructuring, now caused generation failures and has been now completely deleted from the repo (its content has been already moved to other TestSuites during the repo restructuring)


wsafonov:
Generated java tests seem currently not to be executed in the ant build, so this is still an open issue. But the current build can at least already be used by @lhartl to create a migration preview for the customer project.

arimer:
@wsafonov can I just pick any of the listed required changes or is there an order etc.?

arimer:
**Some updates regarding the JAVA execution missing**
It looks like the generated ant-xml part which is used for executiong the generated java test cases does not end up the in the final test `build-tests.xml` file.

Basically everything starting with `<target name="test.tests.generated" depends="assemble">`
```
<target name="test.tests.generated" depends="assemble">
    <patternset id="plugin.deployed.jars">
      <include name="lib/**/*.jar" />
      <include name="languages/**/*.jar" />
      <exclude name="**/*-src.jar" />
    </patternset>
    ...
```
is not generated.

The generator of `org.iets3.opensource.build.gentests` is not invoked due to **invalid priority rules**. 
I am not sure how these prios need to be setup properly. Maybe @sergej-koscejev can help out here?

Genplan gives us this info:
```
Generator for language org.iets3.opensource.build.gentests defines invalid priority rule [org.iets3.opensource.build.gentests#01:*] strictly_together [jetbrains.mps.build.mps.tests#4560297596904874237:*], with no mapping configurations specified at one side. The rule is ignored. 
	Rule: [org.iets3.opensource.build.gentests#01:*] strictly_together [jetbrains.mps.build.mps.tests#4560297596904874237:*]
```

wsafonov:
> The generator of `org.iets3.opensource.build.gentests` is not invoked due to **invalid priority rules**.
> I am not sure how these prios need to be setup properly. Maybe @sergej-koscejev can help out here?
> 
> Genplan gives us this info:
> 
> ```
> Generator for language org.iets3.opensource.build.gentests defines invalid priority rule [org.iets3.opensource.build.gentests#01:*] strictly_together [jetbrains.mps.build.mps.tests#4560297596904874237:*], with no mapping configurations specified at one side. The rule is ignored. 
> 	Rule: [org.iets3.opensource.build.gentests#01:*] strictly_together [jetbrains.mps.build.mps.tests#4560297596904874237:*]
> ```

Ok, it looks like the generator design for the build language has been changed slightly and there is no mapping configuration anymore in the `jetbrains.mps.build.mps.tests` generator. Instead there is a new switch template `switch_BuildAspectInBuildMPS` for `BuildAspects` that should be extended in our case for the `RunKernelFGeneratedTests` concept. Do you want to look into it and may be still fix the generation issue then?

arimer:
>Instead there is a new switch template switch_BuildAspectInBuildMPS for BuildAspects that should be extended in our 
case for the RunKernelFGeneratedTests concept. 

I tried this already out, but  the build models can't be generated afterwards. Unfortunatelly I will not be able to continue here in the near future. If you like I can push my changes to the `org.iets3.opensource.build.gentests` language.

wsafonov:
> If you like I can push my changes to the org.iets3.opensource.build.gentests language.

Sure, please do it - someone would need to jump on it and continue with that commit.

coolya:
Regarding ad3026e I made a change in the mps-gradle-plugin which now forwards the log level from gradle to ant for the MPS tasks. Which means they will now also honor "-i" and "-d" for generate tasks. See https://github.com/mbeddr/mps-gradle-plugin/pull/89

wsafonov:
I have reverted previous change in 8984a55 that removed comments from several statements in the [`path`](http://127.0.0.1:63320/node?ref=r%3Afcb91210-c6db-4de0-81c0-ca99e48fd25a%28test.in.expr.os.records%40tests%29%2F7740953487931040977) TestSuite as this prevented the whole TestSuite from being executed as generated java test. The reason is unclear to me and obviously this has happened before because those lines were commented for a reason. Maybe @lhartl can provide some additional details as she edited that TestSuite previously based on the history.

Apart from that I would wait until 2020.3.1 bugfix is out and prepare the migration for merging with that release because current 2020.3 version still has numerous issues with VCS and file system synchronisation which almost made it imposible to work on changes with this version (at least on Windows).

lhartl:
> I have reverted previous change in [8984a55](https://github.com/IETS3/iets3.opensource/commit/8984a55fdcb81c40ccc06a488e3d7b896d3e28c3) that removed comments from several statements in the [`path`](http://127.0.0.1:63320/node?ref=r%3Afcb91210-c6db-4de0-81c0-ca99e48fd25a%28test.in.expr.os.records%40tests%29%2F7740953487931040977) TestSuite as this prevented the whole TestSuite from being executed as generated java test. The reason is unclear to me and obviously this has happened before because those lines were commented for a reason. Maybe @lhartl can provide some additional details as she edited that TestSuite previously based on the history.

I have some details on this: The commented lines here are paths in combination with optionals. For an optional `none` we generate `null` in Java. This works for most cases but when there is a `none` in the middle of a path, this fails. The generator for optionals was never finished, because no known project uses it.

lhartl (approved):
LGTM

PR #431, opened 2021-01-25
Bugfix/kernelf records equals temporal

lhartl:
Fix typesystem of comparison of KernelF `Records` in combination with `TemporalTypes`. `!=` and `==` can now be applied to `TemporalTypes`containing `Records`


wsafonov (approved):
LGTM

lhartl:
thank you!

PR #432, opened 2021-01-26
empty in data tables

coolya:
Adds support for `empty` inside of data tables that are marked for look ups. 

Adds a new interface for literals that represent `empty` to allow reuse of the logic when using language extensions that create domain specific representations of `empty`.

For more details on the design see the commit messages.

No breaking changes are introduced with this PR.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/structure.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/structure.mps
@@ -566,9 +566,6 @@
       <property role="IQ2ns" value="8355348994405084500" />
       <ref role="20lvS9" node="6sdnDbSlaok" resolve="Type" />
     </node>
-    <node concept="PrWs8" id="MNXm1ElbHv" role="PzmwI">
-      <ref role="PrY4T" node="MNXm1ElbHo" resolve="IEmptyLiteral" />
-    </node>
   </node>
   <node concept="1TIwiD" id="2rOWEwsEji_">
     <property role="3GE5qa" value="option" />
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/structure.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/structure.mps
@@ -566,6 +566,9 @@
       <property role="IQ2ns" value="8355348994405084500" />
       <ref role="20lvS9" node="6sdnDbSlaok" resolve="Type" />
     </node>
+    <node concept="PrWs8" id="MNXm1ElbHv" role="PzmwI">
+      <ref role="PrY4T" node="MNXm1ElbHo" resolve="IEmptyLiteral" />
+    </node>
   </node>
   <node concept="1TIwiD" id="2rOWEwsEji_">
     <property role="3GE5qa" value="option" />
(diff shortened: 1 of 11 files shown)


wsafonov (commented):
What's the rationale behind introducing the marker interface for both, none and empty literals? I mean, we actually wanted to hold them apart and in the interpreter they are also represented by completely different classes. So why would we need to put both literals under one single marker interface?

Apart from that one small typo in `check_DataTable` rule: in the inner clause the loop variable is called `row`, but it holds a `DataCell`.

coolya:
> What's the rationale behind introducing the marker interface for both, none and empty literals? I mean, we actually wanted to hold them apart and in the interpreter they are also represented by completely different classes. So why would we need to put both literals under one single marker interface?

There is no rational because I couldn't understand the rational behind none and empty it seems completely strange to me. They essentially do the same thing representing the absence of a value with different implementations. 

> 
> Apart from that one small typo in `check_DataTable` rule: in the inner clause the loop variable is called `row`, but it holds a `DataCell`.

I'm gonna rename it.

wsafonov:
> They essentially do the same thing representing the absence of a value with different implementations.

Although they indeed both represent the absense of value, they are actually intended for different scenarios (`none` for absense of value for optional types, `empty` for arbitrary types where we want to support the notion of absent value). And because up to now we kept those two scenarios apart (in the language as well as in the intepreter), probably we should stick to that for a better design reasons, making no exception for the `empty`/`none` literals. But maybe @markusvoelter can correct me on this part if I'm wrong (referring also to your words in today's MPS talk that the whole opt<> types is probably not really a practical concept 😉).

Also after looking again on this PR, I have some additional thought that I want to clarify here. The PR introduces another special empty literal `emptyInput` just for the case where we want to match absent values inside of data table w/o the need to provide explicit type (which is then derived from the context). So we essentially add a special `empty` literal for data tables, whereas `empty` literal itself should be constrained not to be available in the data tables to make it correct. This doesn't feel like a clean concept for me. 

My question is - why not instead designing existing `empty` literal to support optional type (and therefore also using it in data tables) and introducing an EmptyType for type system computations like we have done this for Datev project? @markusvoelter this question rather goes to you.

markusvoelter:
Here are my thoughts as I understand things:
* traditionally, KernelF has had option types. `none` is the object the option holds if nothing is in it.
* Later, based on project experience, I added "Nix Support", where a value can represent the notion of nothing or emptyness. It has nothing to do with options. This is where `empty` is used, it's the KernelF equivalent of Datev's `nix`.
* As soon as I get around to it, I want to remove options completely.
* If we really want avoid having both `none` and `empty`, we can make `none` implement the marker interface for "nothing" that `empty` currently implements. But I am not sure this is necessary. I don't see any user of KernelF mix option types and allowing empty values, even before we remove options completely.

wsafonov:
> * If we really want avoid having both `none` and `empty`, we can make `none` implement the marker interface for "nothing" that `empty` currently implements. But I am not sure this is necessary. I don't see any user of KernelF mix option types and allowing empty values, even before we remove options completely.

I actually consider it a good idea to keep `none` and `empty` apart to emphasize that we have here different concepts of dealing with value absense. Then in my opinion there would be no need to introduce a common interface for them, as their handling in the language and in the interpreter would anyway remain different.

The only flaw I see in connection with this PR is that we now want to introduce another dedicated `empty` literal for data tables to support empty thing specification without a type. And therefore it might be worth considering a refactoring of the existing `empty` literal to make it work similar to `none` literal w/o a type. @markusvoelter what do think about that?

AlexeiQ:
I agree to @wsafonov  last comment. `None` belongs to the option-concept and keeping it separate from `empty` makes it easier to stay in that concept (Whether @markusvoelter will remove it or not). 

At Sigi, I recently encountered as well a case where I'd like to have a simple `empty` (or null) w/o a type:
We can define variables with a specified type. And we have statements to assign values to that variables. For list-variables we want to "reset" the list and therefore assign an `empty` to list-variable.  Since the type of the variable is already specified, it is not user-friendly to make the user specify the type again each time.

Thus, I second an `empty` where specifying a type is optional.

coolya:
@wsafonov I removed the marker interface from the NoneLiteral and renamed the variable. I think this is good to merge now.

wsafonov:
> @wsafonov I removed the marker interface from the NoneLiteral and renamed the variable. I think this is good to merge now.

I'd like to discuss this part a bit with Markus before, hopefully he will get a chance tomorrow.

markusvoelter (approved):
* Ok, so we have a marker interface for runtime nix. Datev's "nix" should then 
  probably implement it, even though it doesn't do anything except marking, right?
* The table empty implements this and evaluates to runtime DefaultNix
* There's no relationship to option's none (which I like)
* "emptyInput" is syntactically long/verbose. Why not just "Empty", or if it is 
  table-specific something like "--"?

Right? Then I like it.

wsafonov:
We agreed that we refactor empty literal to support optional type, so it can be used instead of `emptyInput` in tables and in any other places where we want to specify empty value in already typed context. We only leave `IEmptyLiteral` interface so we can write checks or do analysis independent of a specific empty literal implementation (can be customer specific).

Corresponding changes will be merged from `datev-staging` branch, so we can close this PR.

PR #433, opened 2021-01-27
expr.test: improve scoping performance

coolya:
Replaced `nodesIncludingImported` and `where( x.InstanceOf)` combination
with a direct call to `SModelOperations.nodesIncludingImported` in the
TestSuite scope implementation. This change allows MPS to cache the model
and concept tuple and improves scoping performance significant. In one
project it reduced the model checking time by 30%.

MPS doesn't support passing an expression to the concept part of
nodesIncludingImported but only allows direct references to concepts and
calling the underlying API directly was required.


wsafonov:
> MPS doesn't support passing an expression to the concept part of
nodesIncludingImported but only allows direct references to concepts and
calling the underlying API directly was required.

This seems to be a common performance pitfall as most people would not even think about using the direct call to filter for specific concept. @coolya please upvote https://youtrack.jetbrains.com/issue/MPS-32743 and mention this possible performance bottleneck there to put some priority on that ticket 😉

wsafonov (approved):

PR #436, opened 2021-02-08
small fixes in KernelF

lhartl:
- On assert-that test cases, there is the option `go to constraint`. This was broken and is fixed  for all cases in which a constraint fail leads to a failing interpreter test

- `enum-literal.isIn()` lead to an exception, because an empty `IsInTarget`  could not be processed. The case that the `IsInTarget` is empty is now handled. In addition, an empty `IsInTarget` is not allowed anymore, because it is not meaningful

- The type inference os the `IntervalsOp` now returns a list<DateType> instead of a list<NumberType> which is
consistent with the interpreter.

This PR does not contain breaking changes


wsafonov (approved):
LGTM

lhartl:
Thank you

PR #441, opened 2021-02-17
expr.units: do not migrate extensions units

joao-silveira:
And now for mps2020.2


coolya:
Srsly just migrate your project to 2020.3 yet. :D

coolya (approved):

PR #443, opened 2021-02-24
delete all empty generators

lhartl:
With MPS 2020.3, empty Generators in Languages cause modelchecker errors in projects that depend on these Languages and use generation plans. Therefore all empty generators are deleted with this pr.


wsafonov (changes requested):
Looks like you have only removed generators from the languages, but did not delete the files (generator module + template model). Let's avoid dead files in the project ;)

lhartl:
Thank you Wladi, of course we don't want to keep dead files. Sorry that I missed that, I thought checking the `delete files` box in MPS when deleting the generators would do the job. Now I deleted the files manually.

wsafonov:
>  I thought checking the delete files box in MPS when deleting the generators would do the job

That should actually do that. If you are sure, that you've definitely selected checkboxes, that could be also a new MPS bug... 😞

lhartl:
>If you are sure, that you've definitely selected checkboxes, that could be also a new MPS bug... 

I am sure that I did this most of the times and I had to manually delete the files for every generator I touched. I'll double check that and if I can reproduce it, I'll open a bug ticket at Jetbrains.

wsafonov (approved):
Looks good now

wsafonov:
Btw. I can reproduce the issue with non-removed files, gonna open a ticket for that.

lhartl:
> Btw. I can reproduce the issue with non-removed files, gonna open a ticket for that.

Great, thank you! And Thank you for the review!

PR #455, opened 2021-03-08
KernelF: Change list.first/last, so they return none when list is empty

AlexeiQ:
A call on `first`/`last` on an empty list resulted in a thrown exception. Because the typesystem already handles these cases with `Optional`, the interpreter and generator now also return `none` in these cases, without an exception


AlexeiQ:
The interpreter and generator return `none` when `first`/`last` is called on an empty list, same as the typesystem, instead of an exception.

AlexeiQ:
@wsafonov and Me, discovered, that the BangOperator is behaving differently between Interpreter and Generator in when the exception is thrown. I will address this in an additional PR this week.

lhartl:
lgtm

lhartl (approved):

PR #457, opened 2021-03-09
KernelF: Change interpretation of '&&' / '||' to consider the conditi…

AlexeiQ:
The interpreter of KernelF does now consider the conditional aspect of '&&' / '||' and doesn't always evaluate both sides of these expressions.

In case you have `empty && rightHandSide` and want to receive the `rightHandSide`, you need to adapt your NixSupport-Plugin implementation to return `false` instead of `nix` for the leftHandSide NixSupport-Evaluation.

### DIFF
--- a/code/languages/org.iets3.opensource/solutions/org.iets3.core.expr.simpleTypes.interpreter/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/solutions/org.iets3.core.expr.simpleTypes.interpreter/models/plugin.mps
@@ -208,6 +208,10 @@
       </concept>
     </language>
     <language id="fd392034-7849-419d-9071-12563d152375" name="jetbrains.mps.baseLanguage.closures">
+      <concept id="1199542442495" name="jetbrains.mps.baseLanguage.closures.structure.FunctionType" flags="in" index="1ajhzC">
+        <child id="1199542457201" name="resultType" index="1ajl9A" />
+        <child id="1199542501692" name="parameterType" index="1ajw0F" />
+      </concept>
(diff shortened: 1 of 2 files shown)


AlexeiQ:
The interpreter of KernelF does now consider the conditional aspect of '&&' / '||' and doesn't always evaluate both sides of these expressions

wsafonov (changes requested):
Frankly I don't like how it's now implemented because obviously simple operations like `&&` and `||` get that heavy and hard to read/understand interpreter logic. Also we now explicitly depend on if an operand evaluates to a INixValue and how this would affect the final result of the overall operation which I assumed to be completely encapsulated/hidden in the NixHandler. This now introduces an exception. But I don't have any better solution right now. 

What I would at least enhance:
-  Invoking NixSupport for the short-circuit case for a binary operation, but with only one argument value would be against the expectations. I would provide explicit `true` or `false` value, depending on the operand.
- please also give a more meaningful name to the `run` variable as result

markusvoelter:
What would these true and false values do? Remember we can't find out by evaluating the right side.

AlexeiQ:
Fixed:  "please also give a more meaningful name to the run variable as result "

To the first point I don't see any advantage using the 2-Argument-NixSupport with default-values for the arguments, that shouldn't be evaluated. It's just more code, but not easier to read or understand.

AlexeiQ:
@wsafonov All requested changes are implemented

wsafonov:
Sorry, I should have checked tests as well. We have currently:

```
assert nix && false equals empty[C] [no result found]
assert false && nix equals false [C] [no result found]
```

which means that `&&` is not symmetrical. Probably this is not exactly what we want, right? 😉

wsafonov (approved):

Issue #459, opened 2021-03-11
Bugfix async Editor Highlightning


dbinkele:

dbinkele:
The logic for updating ISolvable error messages in async mode was buggy. Sometimes UI-updates were missed. This  has been fixed.

PR #460, opened 2021-03-11
Bugfix/async editor highlightning#459

dbinkele:
Each pull request should contain a brief summary of the proposed changes  
that can be easily read by the consumers of project releases.

Use the first comment in the pull request and update it as needed during
the work on pull request to keep the description up-to-date.

Some of the items in the description might include:

- compatibility issues with previous versions if this PR introduces any
breaking changes

>>>>> no issues 

- hints on the included migrations and things to consider when applying
the migration to the own project

>>> no migrations
- reference to a related Github issue

#459


wsafonov (commented):
Did a formal review, looks good to me. Just one comment: in `SolverChecker.update()` you store result value in a `time` variable, the value of which is however treated as count in other places. Is `time` then really a good name for such a variable? Just asking to make the code more readable.

dbinkele:
@wsafonov the variable `time` has been renamed to `changeCount`

wsafonov (approved):

PR #461, opened 2021-03-18
Migration of the context actions to the MPS context actions

slisson:
This moves the actions from the "Context Actions 2" to the "Context Actions" tool window. The goal is to remove the "Context Actions 2" tool from the mbeddr platform.

The actions after the migration are not identical to the ones before the migration. Some were just removed, that don't appear to be useful and are not worth the effort to reimplement them.


wsafonov:
> Some were just removed, that don't appear to be useful

Can you give an example, which one you mean? Do those actions originate from out languages or from MPS?

slisson:
> > Some were just removed, that don't appear to be useful
> 
> Can you give an example, which one you mean? Do those actions originate from out languages or from MPS?

For example the actions for creating "top level" concept instances. I assume users just use the code completion menu instead of the context actions tool.

wsafonov (approved):

PR #464, opened 2021-03-29
Unify AH2 and AH

sergej-koscejev:
There is now only one AH class that is used from both the interpreter and the generator.


wsafonov (approved):

sergej-koscejev:
It was pointed out to me that there's actually another AH class that only contains `isSome` method. It will be removed at some later point.

PR #465, opened 2021-03-29
[breaking] Extract shared runtime from base runtime to reduce runtime dependencies

sergej-koscejev:
Terms used:
* _base runtime_: solution `org.iets3.core.expr.base.runtime`
* _shared runtime_: solution `org.iets3.core.expr.base.shared.runtime`

The new shared runtime contains classes that will be used both from the generated code and by the interpreter or in-MPS code.

Even though the individual classes used by the generated code do not depend on MPS-specific classes, extracting them from the base runtime means that the generated code no longer needs to depend on a module that contain MPS-specific code (e.g. NixSupport with its usage of the node<> type).

The base runtime was taken out of module dependencies of other runtimes.

Breaking change: class INixValue was moved from the base runtime to the shared one so references to it will need to be updated (Ctrl+R to import org.iets3.core.expr.base.shared.runtime model, then F5 to update
the reference).

Another potentially breaking change is that the generated code now uses the shared runtime and its classes instead of the base runtime (e.g. SharedMinMaxHelper rather than MinMaxHelper). After it is regenerated, it will need the shared runtime JAR to execute (and should no longer need the base runtime JAR).

### DIFF
--- a/code/languages/org.iets3.opensource/.mps/modules.xml
+++ b/code/languages/org.iets3.opensource/.mps/modules.xml
@@ -78,6 +78,7 @@
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.base.collections.stubs/org.iets3.core.expr.base.collections.stubs.msd" folder="expr.lang-core" />
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.base.interpreter/org.iets3.core.expr.base.interpreter.msd" folder="expr.lang-core" />
       <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.base.runtime/org.iets3.core.expr.base.runtime.msd" folder="expr.lang-core" />
+      <modulePath path="$PROJECT_DIR$/solutions/org.iets3.core.expr.base.shared.runtime/org.iets3.core.expr.base.shared.runtime.msd" folder="expr.lang-core" />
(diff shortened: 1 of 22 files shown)


wsafonov (approved):
LGTM, we would just need to communicate the breaking change properly to the customer projects.

sergej-koscejev:
> LGTM, we would just need to communicate the breaking change properly to the customer projects.

Who is "we" and what exactly is "properly"?

wsafonov:
> Who is "we" and what exactly is "properly"?

Those in itemis who are involved in projects using iets3.os. But we would anyway soon have a proper change log with release notes which would simplify this task.

sergej-koscejev:
@wsafonov I want to merge this PR this week. Do you want me to increase some version number somewhere or can I just go ahead and merge?

wsafonov:
@sergej-koscejev we don't apply any semantic versioning (yet), so it's only important that we can clearly track all changes in PRs between the specific builds. If @lhartl is also ok with the breaking changes so far, you can just go ahead and merge.

lhartl:
Looks good to me. 
For communication: I already posted a heads up in the KernelF channel when I noticed this PR first.

lhartl (approved):

wsafonov:
@sergej-koscejev is it still blocked by anything or can we merge it soon?

sergej-koscejev:
@wsafonov you can merge it, I just haven't had time to do it myself yet.

PR #466, opened 2021-03-30
[breaking] Temporal runtime no longer depends on MPS collections and closures

sergej-koscejev:
Instances of collections and closures language in temporal runtime were rewritten using plain Java, to remove the dependency on MPS collections and closures runtimes.

Breaking changes:
* `TemporalValue#slices()` no longer returns a base language collection but a `java.util.List` so the collection language operations like `add` or `sortBy` are not applicable to it anymore.
* `TemporalValue#intervals()` likewise.

PR #470, opened 2021-04-01
[breaking?] Publish KernelF runtime + remove MPS dependency from published artifacts

sergej-koscejev:
KernelF runtimes are published as separate artifacts, with dependencies.

Also contains changes described in #491 (removed MPS dependency from published artifacts), which could be potentially breaking, therefore we mark this PR also as breaking.


sergej-koscejev:
This contains #466, merge that before this one.

sergej-koscejev:
The list of runtime modules that are combined does not include `*.genjava.rt` because it's not clear whether it should or not. Comments?

wsafonov:
Could you add some comments on the rationale for combining several jars into one artifact? Is it just convenience of not having to provide 4 jars on the classpath?

sergej-koscejev:
> Could you add some comments on the rationale for combining several jars into one artifact? Is it just convenience of not having to provide 4 jars on the classpath?

Yes, that's the main reason. I'm not sure myself if it's a good enough reason.

wsafonov:
> Yes, that's the main reason. I'm not sure myself whether it's a good enough reason.

That exactly what I'm asking myself. In case we would have numerous (>5) runtimes that we would want to use frequently in multiple scenarios, then yes, it would make perfect sense. Right now though the only scenario is for external generation. And creating a special publication for that feels a bit too much. Rather I would expect that such kind of automation would be created on the consumer side.

But if you say there is no optimal way to do it elsewhere and it would save you a lot of time and make things really easy, well let's just consider that as good enough reason for now.

> The list of runtime modules that are combined does not include *.genjava.rt because it's not clear whether it should or not. Comments?

If that runtime is only used in the generation, I would keep it like that. But anyway, right now you are the only consumer of the new combined runtime artifact, so you can decide, if you need it there or not.

sergej-koscejev:
Also note that I adjusted the dependencies of other artifacts here, like in #491.

wsafonov (approved):
LGTM now, thanks for cleaning up!

Issue #463, opened 2021-03-24
Add Toolview for inspecting ISolvables running asynchronously


dbinkele:
As it is possible to run ISolvables in  the background in asynchronous manner, there is a need to track those. We need to provide a view where running and queued ISolvables can be inspected.

PR #472, opened 2021-04-12
Feature/toolviewIsolvablesRunning (#463)

dbinkele:
This PR implements the tool requested in #463.

**Summary:**
As it is possible to run an ISolvable in asynchronous mode (via AsyncSolverTaskExecutor) it is desirable to have an overview of the tasks in the background This pull request will add such a view. Currently it is only testable together with the iets3.core branch 'AddAsyncSolverFMandFMC#682'.  The attached screenshots give an impression.

image
 
image

image
Clicking ISolvablse shows a pop-up which lists the most recent 20 tasks which are running, queued cancelled or terminated.

Probably there is a better descriotion than 'ISolvables' ... we are open to proposals.


wsafonov:
In general looks ok to me, just two thoughts:
- probably using a color scheme for so many states like Creating, Queued, Running, Terminated, Canceled is not really the perfect choice from the usability point of view - it would be hard to guess and to remember the meaning of each color. So for a proper solution there should be at least a tooltip on each state icon with its name or actually specific icons.
- for adding status bar widgets you are using deprecated IntelliJ API scheduled for removal. The proper approach would be using the `StatusBarWidgetFactory`. Just saying it cause when we would migrate to a new MPS version where this API would be dropped, we would need to disable this status bar widget until a proper factory implementation would be provided.

dbinkele:
@wsafonov cancelled-> rejected, terminated -> done.

wsafonov:
@dbinkele another thought I had is regarding the state `creating` - does it really make sense from the user perspective? Would any task be shown in that state at any time? I would expect for a new task to be either running, queued or rejected, that should be sufficient. In fact, strictly speaking `creating` is not a task state, as no task exists yet, right? And so for simplification reasons (also due to the icons issue) I would suggest to get rid of it. Or do you actually see a reasonable use case for it?

dbinkele:
> @dbinkele another thought I had is regarding the state `creating` - does it really make sense from the user perspective? Would any task be shown in that state at any time? I would expect for a new task to be either running, queued or rejected, that should be sufficient. In fact, strictly speaking `creating` is not a task state, as no task exists yet, right? And so for simplification reasons (also due to the icons issue) I would suggest to get rid of it. Or do you actually see a reasonable use case for it?

@wsafonov In the solverTask case 'creating' corresponds to a generator-run. This happens prior to the running SolverTask and actually creates the SolverTask. From a user perspective this does not differ much. But it would give the developer a hint where a problem might occurr.

wsafonov (approved):
Let's give a go, LGTM.

PR #474, opened 2021-04-19
Fix EmptyValue editor, so specification of a type is possible again

AlexeiQ:
Fix EmptyValue editor, so specification of a type for the EmptyValue is possible again

Bug: On creation of an EmptyValue instance the cursor is positioned at "empt|y" and cannot move to "empty|",
thus it's not possible to add a type to the EmptyValue. Going to the right results in a jump to the next cell


AlexeiQ:
It seems, that the editor cannot handle `punctuation-right: true` on a cell, if the next cell is a `grammercells.OptionalCell`. Do we need to check this out in the grammercells, or is this a known behaviour?

slisson:
Before I look at the details, did you try the style "last-position-allowed" (or similar)?

AlexeiQ:
I've used a simple condition on the `punctuation-right` to only set it, if the optional is `true`. But `last-position-allowed` works as well, so I'll use that, to stay with a common solution. Thx very much @slisson !

lhartl (approved):
Looks good to me

PR #476, opened 2021-04-29
Explicitly use node.getPresentation for string conversion

wsafonov:
In the last MPS version (2020.2) there was a change that impacted how nodes are converted to strings ([MPS-32042](https://youtrack.jetbrains.com/issue/MPS-32042)). Due to this change for some nodes the string+node concatenation could produce different result not readable from the user point of view (contains `(instance of <concept>)` instead of custom node presentation). To fix this, we need to always use node.getPresentation() explicitly in such cases. This PR adds required calls for all cases of previously implicit conversion of nodes to string.


lhartl (approved):
Looks good to me. Thank yu for the clean up.

PR #477, opened 2021-05-03
run migrations from migration-assistant

AlexeiQ:
migrations: run re-save project/ update versions in descriptors and reload modules from disk in the build scripts


lhartl (changes requested):
The changes in the build script look a bit weird to me, especially the introduced dependencies for the genall.advanced devikit don't make sense imho. When I run reload modules on disk, everything is basically changed back to what is there on master. 

Also there is a strange dependencyVersion introduced in the analysis base language to org.modelix.model.api. Where does this come from? 

Is there a reason for doing these changes?

wsafonov:
> Also there is a strange dependencyVersion introduced in the analysis base language to org.modelix.model.api. Where does this come from?

This is added as a transitive dependency due to an unused dependency in analysis base that I have already removed in #476 which has been just merged :)

> The changes in the build script look a bit weird to me

Yepp, they are indeed as it's a bug in MPS with devkit dependencies management in the build script. If you run reload twice and basically everything is back to the original state (or only the order of dependencies has been changed), you can safely skip any of those changes to avoid switching it back and forth on master.

wsafonov:
It would be best if you rebase on latest master and then check the migrations again. Changes in the build probably can be skipped as described above.

wsafonov (approved):
I have rebased on master and cleaned-up the build script changes, so that only effective changes are in the commit

lhartl (approved):

AlexeiQ:
Thx for the help and reviews here!

Issue #480, opened 2021-05-06
Shortcut for "Run manually" doesn't work if root node is a ICanRunCheckManually


kbirken:
The `RunManually` infrastructure provides three actions for running the check on a node, a root node and a container, respectively. The keymap `runEval` ([nodeURL](http://127.0.0.1:63320/node?ref=r%3Acdcdec44-a636-42c8-b599-c146eb2ca77d%28org.iets3.core.plugin.plugin%29%2F4816256765116125442)) assignes the same shortcut to each of them, and the `isApplicable` methods decide which one will be executed. 

However this logic overlaps for nodes and root nodes, such that if a root node is itself a `ICanRunCheckManually`, no check will be executed (as the root node will be checked first if the shortcut is hit, but its `execute` method doesn't find itself as ancestor. 

Expected: If the shortcut is hit, the `runManuallyOnNode` action should be triggered.

alexanderpann:
Closing because it was fixed in https://github.com/IETS3/iets3.opensource/pull/481 and follup commits.

PR #481, opened 2021-05-06
Allow RunManually shortcut also if root node is a ICanRunCheckManually

kbirken:
If a root node is a `ICanRunCheckManually`, the Cmd-Alt-Enter shortcut doesn't work currently (issue #480). This is because the `runAllManuallyOnRoot` action will grab the shortcut, but its `execute` will not handle it correctly. 

There are two possible options to solve this:
- option 1: fix `runAllManuallyOnRoot` so that it not only handles checks for the root's descendants, but also for itself
- option 2: fix the shortcut logic so that `runAllManuallyOnRoot` will not grab the shortcut, but leave it to `runManuallyOnNode`

This PR solves the problem by doing option 2. I chose to avoid option 1 because it seems so far it was perfectly fine if root nodes only check there children and not themselves, and I decided to stick with that behavior.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.plugin/models/org/iets3/core/plugin/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.plugin/models/org/iets3/core/plugin/plugin.mps
@@ -180,6 +180,9 @@
       <concept id="1079359253375" name="jetbrains.mps.baseLanguage.structure.ParenthesizedExpression" flags="nn" index="1eOMI4">
         <child id="1079359253376" name="expression" index="1eOMHV" />
       </concept>
+      <concept id="1081516740877" name="jetbrains.mps.baseLanguage.structure.NotExpression" flags="nn" index="3fqX7Q">
+        <child id="1081516765348" name="expression" index="3fr31v" />
+      </concept>
       <concept id="1204053956946" name="jetbrains.mps.baseLanguage.structure.IMethodCall" flags="ng" index="1ndlxa">
         <reference id="1068499141037" name="baseMethodDeclaration" index="37wK5l" />
         <child id="1068499141038" name="actualArgument" index="37wK5m" />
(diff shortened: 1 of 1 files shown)


kbirken:
Ensure that the keyboard shortcut  for _Run/Evaluate/Check Manually_ works even on a root node which can be checked (i.e., is a `ICanRunCheckManually `). For details see #480. There is no breaking change involved.

wsafonov:
Would it be really the expected behavior to execute only roots descendants if the root itself is also `ICanRunCheckManually` and can be checked? To me it feels that when a user is on a checkable node and hits the shortcut, he/she would want to see all nodes in the tree checked including the node itself. At least I can't see any reason why excluding the selected node would be a reasonable expectation. Or do we have examples that would rather support that expectation?

kbirken:
> Would it be really the expected behavior to execute only roots descendants if the root itself is also `ICanRunCheckManually` and can be checked? To me it feels that when a user is on a checkable node and hits the shortcut, he/she would want to see all nodes in the tree checked including the node itself. At least I can't see any reason why excluding the selected node would be a reasonable expectation. Or do we have examples that would rather support that expectation?

At least we did not have a use case for this so far, so I chose to fix the issue without overgeneralizing the implementation. But I am not stubbornly fixed to option 2. In order to implement option 1 we would have to include the root node itself in its `execute` method, but also add `&& checkable.parent!=null` to `runManuallyOnNode.isApplicable`'s condition, in order to avoid that both actions could be applied in the same use case (it would be bad if we end up with non-determinism here). What do you think?

kbirken:
The latest commit switches to option 1. Pls. have another look.

wsafonov:
Shouldn't it be `this.selected` in the condition? Because this way if only the root would be `ICanRunCheckManually` we would loose the possibility to click somewhere under it and run the solver for the root with the shortcut which would use `runManuallyOnNode` action then.

kbirken:
Pls. check again.

wsafonov (approved):
Looks good now.

PR #482, opened 2021-05-20
added navigation support to solvable node from monitor (#463)

wsafonov:
This PR extends the solvable monitor widget with the ability to navigate on mouse click to the ISolvable node behind the solver task to show it in the editor. This is especially helpful in case of many tasks shown in the monitor widget.


dbinkele:
@wsafonov I have added a simple null check to the listener as I was getting exceptions when no task that have run were available.

dbinkele (approved):

lhartl (approved):

PR #483, opened 2021-05-25
Feature/dectree improvements

markusvoelter:
I have made a bunch of syntax and usability improvements of the decision tree.


lhartl (approved):
Looks good to me. 
It would be nice though to have an example of the dectree. I assume you removed the test because there is no Java generator for the dectrees. Since there is an Interpreter, please put the test in test.in.expr.os/todo@tests. For this model only the interpreter (but not the generator) is not executed.

PR #484, opened 2021-05-27
Fix wrong generation for AltOption with BlockExpression.

AlexeiQ:
Bug: When an `AlternativesExpression` is used with a `BlockExpression` in its `AltOption`, the generator always generates `return 5;`.
Bugfix: Add a node macro to use the specified `BlockExpression` as actual generation result.


AlexeiQ:
I added also tests. The interpreter could handle this use case before and thus is not concerned.

PR #486, opened 2021-06-14
Ran pending migrations from mbeddr-platform

arimer:
After upgrading to the newest iets3 artifact in a customer project, we are facing some MPS warnings regarding not migrated depenencies: 
image

This PR just rans all migrations proposed by the migration assistant:
image


wsafonov (approved):
LGTM, thanks for submitting!

PR #487, opened 2021-06-21
Fix exception in generated code when modulo is used with floatingpoint

AlexeiQ:
Bug: When the modulo expression is used with a FloatingPoint, which is rounded, the generated code runs into an cast exception, trying to cast `BigDecimal` to `BigInteger`.
Solution: Handled the transformation from `BigDecimal` to `BigInteger` in the right way, instead of a simple cast.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
@@ -3,6 +3,7 @@
   <persistence version="9" />
   <languages>
     <use id="7a5dda62-9140-4668-ab76-d5ed1746f2b2" name="jetbrains.mps.lang.typesystem" version="5" />
+    <use id="b401a680-8325-4110-8fd3-84331ff25bef" name="jetbrains.mps.lang.generator" version="3" />
     <devkit ref="a2eb3a43-fcc2-4200-80dc-c60110c4862d(jetbrains.mps.devkit.templates)" />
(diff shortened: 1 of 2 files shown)


AlexeiQ:
When the modulo expression (`ModExpression`) is used with a FloatingPoint, which is rounded (`ConvertPrecisionNumberExpression`), the generated code runs into an exception 
 `java.lang.ClassCastException: class java.math.BigDecimal cannot be cast to class java.math.BigInteger ` .   
The problem is, that the `BigDecimal` which is returned from the rounding is simply cast to `BigInteger`, which is not possible.  
I've implemented a solution, but it feels not optimal, because
1) the choice whether a reduction rule is used for `BigInteger` or `BigDecimal` is determined on whether the concept for rounding is used and not on the type of the value ( eg; `BigInteger` or `BigDecimal`).
2) the transformation from `BigDecimal` to `BigInteger` could be simplified.

A question I couldn't answer yet is: Why is the exception happening on the left side of `mod`, but not on the right side (since both sides use the same code in the generator)

I would gratefully appreciate any hints or help on how to improve this solution.

lhartl:
I could not have a look at the code yet, but from your description of the problem, I came up with a question/idea: How about changing the generator of `ConvertPrecisionNumberExpression ` so that if a Number is rounded to an Integer, a `BigInteger`. 
This would need some investigation whether the change to the `ConvertPrecisionNumberExpression` breaks something.

AlexeiQ:
I had the same idea at some point, but then though, that people probably would continue a calculation with the rounded value and want to keep the `BigDecimal` without additional converting. Additionally, I think the conversion should be explicit and not implicit, thus giving people a choice.

AlexeiQ:
I had another look, and am satisfied with this solution now. But I've created another MR #498 with an alternative implementation.

wsafonov:
Closing in favour of alternative fix #498

PR #488, opened 2021-06-24
build: added publication to github packages

arimer:
This PR adds a publication to github packages for all artifacts.
In case we are building from CI and running on master/maintenance or a "mps" branch , TC will deploy all corresponding artifacts to github pages in addition to the nexus repository.

I've taken some inspiration from the PR provided for the mbeddr.platform publication: https://github.com/mbeddr/mbeddr.core/pull/2180

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,6 +71,7 @@ def mbeddrVersion = "2020.3+"
 // the name is enough no trailing "." is required, also the plain name from git can
 // be used here. No need to convert "/" the script will take care of that.
 def mbeddrBranch = ""
+def currentBranch = ""
 
 
 if (mbeddrBranch != null && !mbeddrBranch.trim().isEmpty()) {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -362,4 +379,13 @@ githubRelease.doFirst {
     }
 }
 
-githubRelease.dependsOn packageDistroWithDependencies
+if(currentBranch == "master" || currentBranch.startsWith("maintenance") || currentBranch.startsWith("mps")) {
+    /* this is pretty much a workaround so we don't need to change anything in Teamcity. Teamcity calls the  publishAllPublicationsToMavenRepository
```
sergej-koscejev, line 383: This comment doesn't seem to apply to the iets3.opensource builds, they call `gradle publish` which should publish to all defined repositories.

arimer: I see. So just restricting the repository block to the correct branch, should be enough to allow publications for master/maintaines only.
Thx. for the hint.

sergej-koscejev (changes requested):
Will approve after the unnecessary block is removed (or is confirmed to be in fact necessary).

build.gradle
arimer: You were right. TC is currently just running `build publish`. So we dont need this addition block here.

arimer:
@sergej-koscejev removed  unnecessary block.

sergej-koscejev (approved):

PR #491, opened 2021-06-25
[breaking?] Remove MPS dependency from the published POMs, add tests → opensource dependency

sergej-koscejev:
While it's a good idea to know which version of MPS a given version
uses, this is currently solved by the version number. Having MPS in
dependencies causes problems when downstream projects want to use Maven
to unpack all their dependencies to use as project libraries in MPS.

A version of MPS would get unpacked among other dependencies such as
mbeddr, and the dependency on MPS would have to be manually excluded to
prevent this.

On the other hand, the tests artifact was missing the dependency on
the opensource artifact, this commit adds it.

BREAKING? This could potentially be a breaking change if downstream projects expect opensource and tests to have MPS as the dependency.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -279,14 +279,6 @@ publishing {
                     dependencyNode.appendNode('version', it.moduleVersion)
                     dependencyNode.appendNode('type', it.moduleArtifacts[0].type)
                 }
-                configurations.mps.resolvedConfiguration.firstLevelModuleDependencies.each{
-                    def dependencyNode = dependenciesNode.appendNode('dependency')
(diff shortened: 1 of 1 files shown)


wsafonov:
I don't remember why MPS was added as a dependency long ago, but this could have happened indeed due to a non-MPS-based version of the published artefact at that time. So presumably this should not be breaking. But of course we will only know that after this change arrives somewhere in a downstream project. Still we can cross fingers and try it.

wsafonov:
Closing in favour of #470 which already contains this change with some other refactorings.

PR #494, opened 2021-06-30
[breaking] publication: cleanup artifact publications

### LINKED ISSUE #493: Github Packages Puglication returns status code 422
Our TC publication to github packages is not working properly and we receive:

` status code 422 from server: Unprocessable Entity`

arimer:
It seems like TC is not able to deploy our artifacts to github packages. 
The reason for this might be the artifactID of our `allScripts` artifact. It seems like the maven repositories on ghp do not support upper case IDs.

This PR removes the publication for the `allScript` artefact. This artefact is not really required, because all of its content is already part of the overall [org.iets3.opensource](https://projects.itemis.de/nexus/content/repositories/mbeddr/org/iets3/opensource/) artefact.
~~This PR sets the corresponding IDs to lower case.~~

~~However I am not etirely sure if this will somehow break the backwards compatibility, in case users rely on the initiall artifact ID. 
Maybe @sergej-koscejev might know?~~

FIxes #493

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -279,7 +279,7 @@ publishing {
     publications {
         allScripts(MavenPublication) {
             groupId 'org.iets3.opensource'
-            artifactId 'allScripts'
+            artifactId 'allscripts'
             artifact packageAllScripts
         }
         openSource(MavenPublication) {
--- a/build.gradle
+++ b/build.gradle
@@ -218,12 +218,6 @@ task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
 
 check.dependsOn buildAndRunTests
 
-task packageAllScripts(type: Zip, dependsOn: buildAllScripts) {
-    archiveBaseName = 'org.iets3.opensource.allScripts'
-    from artifactsDir
-    include 'org.iets3.opensource.allScripts.build/**'
-}
-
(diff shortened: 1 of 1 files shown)


arimer:
It looks like the build artifacts are anyway packaged as part of the "main" ` org.iets3.opensource` artifact: 

grafik

So maybe it makes also sense to clean those up entirely.

wsafonov:
> So maybe it makes also sense to clean those up entirely.

Agree, its content is already published in the main artefact, as part of the `org.iets3.build.os` plugin. So it should be safe to remove it. Maybe it could break some downstream projects still using the `allScripts` artefact, but they should be able to adjust their dependencies. Therefore we should mark this change as breaking.

arimer:
> Therefore we should mark this change as breaking.

Is there a common approach how we do this? 
Should we just comment this on the commit message or in the PR?

wsafonov:
> Is there a common approach how we do this?

So far we mark the PR with `[breaking]` prefix. Mentioning in the commit would be also welcome.

arimer:
I'm fine with this. Will adapt the code accordingly.

arimer:
Updated the description and removed publication + packaging tasks.

arimer:
@wsafonov should we give it a try?

wsafonov (approved):
So far no objections, so we clean it up.

PR #496, opened 2021-07-01
MPS 2021.1 migration

wsafonov:
This PR contains 2021.1 migration and a workaround for the build that was required to make the build run without generation issues. 

Due to the new [Java compiler infrastructure used in 2021.1](https://blog.jetbrains.com/mps/2021/05/mps-2021-1-has-been-released/#java-compiler), the project generation in the Ant build failed with compilation errors when compiling module cycles. Specifically, the module cycle between `org.iets3.core.expr.base` language and its runtime solution `org.iets3.core.expr.base.runtime` could not be properly compiled. 

In the first approach I've tried the suggested workaround and configured the old Eclipse Compiler for Java (ECJ) as a fallback solution with `mps.compiler.java=ecjlegacy`, which however didn't help. 

Then I have tried to refactor the module cycle (although it would require moving a lot of classes between the modules, adjusting dependencies and causing breaking changes for the downstream projects). However, even without a module cycle the generation ran into other weird compilation issues so I stopped moving into this direction. The results of this refactoring are on the branch `bugfix/compile-module-cycles`.

Because I noticed during the issue investigation that generating only those two modules alone (with their corresponding required dependencies) surprisingly didn't cause any compilation errors, I have provided a workaroud in a separate pre-build step which runs before the main language build and only includes the generation of the problematic modules. This solves the issue for now and allows to move on with the migration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,7 @@ code/languages/*/*.iws
 code/languages/*/build.properties
 code/languages/*/TEST*.xml
 code/languages/org.iets3.core/build/tmp
-code/languages/org.iets3.opensource/.mps/libraries.xml
+.mps-caches
 
 # -----------------------------------------------------------
 # GRADLE
--- a/build.gradle
+++ b/build.gradle
@@ -63,7 +63,7 @@ def major = "2021"
 def minor = "1"
 
 // Dependency versions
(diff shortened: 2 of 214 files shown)


lhartl (approved):
LGTM. Thank you for taking care of the migration

PR #497, opened 2021-07-02
Use pcollections from Maven Central

sergej-koscejev:
The version that was checked into the repo is presumably a 2.x version, now the build will use version 3.1.4. I went through the commit log of pcollections and it appears that 3.1.4 is fully backwards compatible with 2.x (except for bugs).


wsafonov:
Changes were merged as part of #470

PR #498, opened 2021-07-08
Fix exception in generated code when modulo is used with floatingpoint - alternative [small change]

AlexeiQ:
Bug: When the modulo expression is used with a FloatingPoint, which is rounded, the generated code runs into an cast exception,
trying to cast `BigDecimal` to `BigInteger`
Solution: Change result type of rounding to `BigInteger`, if `precision == 0` (looses its decimal places).


AlexeiQ:
This is an alternative fix for  PR #487 , where the `AH.round(..)` is adapted to return a `BigInteger` type instead of `BigDecimal` when the precision is set to `0`, meaning, there are no decimal places after the rounding.
I've asked on slack if this change would be okay for everyone.

wsafonov (approved):
LGTM

PR #499, opened 2021-07-08
fix firstN/lastN semantics for N greater than list

### LINKED ISSUE #402: firstN/lastN generator improvement
see comments in #394

jonaskraemer:
Resolves #402 

As mentioned in #394, the code generated for `list.firstN/lastN(n)` caused an IndexOutOfBoundsException when `n` was greater than the size of `list`.
With this change, the original list is returned instead in such cases, aligning the semantics of firstN/lastN in the generator with the semantics from the interpreter.

Also, this makes sure `list` is evaluated only once in the generated code for each firstN/lastN instead of up to 4 times, which might matter when `list` is a complex expression.


sergej-koscejev (approved):
Looks good to me.

PR #501, opened 2021-07-13
build: add github packages publication for artifacts

arimer:
This PR adds a ghp for all artifacts:
* org.iets3.opensource
* org.iets3.opensource.tests


wsafonov (approved):
Looks good, let's ship it!

PR #503, opened 2021-07-15
Fix two inconsistencies with Nix-handling in interpreter

kbirken:
This PR fixes two inconsistencies in the interpreter for the following binary operators:

- logical and (`&&`): the special logic for short-circuit-handling was non-symmetric and different from the interpreter for `||`, adapted it
- equality (`==`): the implemented logic was different from the one for non-equality, adapted it

Regarding the short-circuit-logic for `&&` and `||`: I am not sure if this is generic enough, I had some problems implementing ternary logic using this. But as I don't know how other NixHandlers (from customer projects) are using this, I couldn't come up with a better proposal. This is maybe something for a different PR.

### DIFF
--- a/code/languages/org.iets3.opensource/solutions/org.iets3.core.expr.simpleTypes.interpreter/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/solutions/org.iets3.core.expr.simpleTypes.interpreter/models/plugin.mps
@@ -2442,19 +2442,52 @@
                 </node>
               </node>
             </node>
-            <node concept="3clFbJ" id="4EEJFuvbYy0" role="3cqZAp">
-              <node concept="3clFbS" id="4EEJFuvbYy1" role="3clFbx">
-                <node concept="3cpWs6" id="4EEJFuvbYy2" role="3cqZAp">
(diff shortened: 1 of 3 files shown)


kbirken:
Summary: Fix Nix-implementation in interpreter for logical-and and equality.

wsafonov (approved):
Thanks for adding some tests! Although `nix` != `nix` equals `nix` looks pretty weird, that's how we have defined it in the default nix handler. In Datev project this is differently and makes more sense. But I think this is ok for us.

kbirken:
Thanks! On [metacpan](https://metacpan.org/pod/Math::Logic::Ternary) there is a good explanation how to propagate the `nil` values in Kleene ternary logic, which might explain some of the test expectations looking weird:

```
In Kleene ternary logic, nil can be thougt of as a value that can either be true or false, but about which nothing is known at each time of evaluation. If nil values take part in a logical expression, the result can be calculated using Boolean logic by replacing all occurences of nil independently by true and false in turn (which requires two to the power of the number of nil values evaluations in total). If the result is always the same, that value is also the value of the ternary logical expression. Otherwise, its value is nil.
```

PR #504, opened 2021-07-16
Fixes/features from datev-staging branch

wsafonov:
* added `real` comparison tests
* show `failed` status for test suites as soon as at least one test fails, independent of whether there are incomplete ones
* added new `indexOf` operation for ordered collections
* improved concept coverage (global coverage statistics were not properly recorded, only on the instance level that was execution specific)
* changed `int / int` to produce `real` instead of `int`
* removed superflous BACKSPACE editor cell actions


sergej-koscejev (commented):
* short description of IndexOfOp is wrong
* type of IndexOfOp is given as positive integer (0+), but it can return -1 if element is not found.
* I don't understand what the fix of coverage is fixing and why it's done the way it's done
* Generator and interpreter of IndexOfOp look very similar, could we extract the shared code into a shared runtime?

wsafonov:
@markusvoelter can you please address the points regarding the `IndexOfOp`?

Regarding the coverage fix, maybe @coolya can leave a short comment here about what happened there.

markusvoelter:
> type of IndexOfOp is given as positive integer (0+), but it can return -1 if element is not found.

Changed the type to regular `int`

markusvoelter:
> short description of IndexOfOp is wrong

fixed.

markusvoelter:
> Generator and interpreter of IndexOfOp look very similar, could we extract the shared code into a shared runtime?

AFAIK we do not have a runtime solution for the collections language. And I don't think we want to introduce one for this duplication alone, do we?

coolya:
> @markusvoelter can you please address the points regarding the `IndexOfOp`?
> 
> Regarding the coverage fix, maybe @coolya can leave a short comment here about what happened there.

When doing the refactoring I missed the “visistedConcepts” which were not recorded at the global instance but only on the instance that was execution specific. I updated the implementation to to also update the global statistics and to reset them when the global stuff is rested.

sergej-koscejev (approved):

PR #506, opened 2021-07-28
Remove duplicate node substitute action for EnumDeclaration

AlexeiQ:
Remove duplicated auto-completion entries for `EnumDeclaration.qualified` flag and `type` option, when declaring enums and on side-transformations.


wsafonov (changes requested):
If we do this workaround anyway, then you should also remove duplicate qualified flag transformation actions as well. But of course this is not a clean solution - probably we should give it another try with the new grammarcells integration (as soon as it will be available).

wsafonov (approved):

PR #509, opened 2021-08-23
Add information about project structure to readme

alexanderpann:
- Contains explanation for all languages and solutions categorized by virtual packages.
- Relevant concepts are linked and can be opened with the HTTP support plugin.
- All available devkits are listed at the bottom of the page.
- Relevant KernelF documentation is linked whenever it is possible.

You can see the updated readme in [my fork](https://github.com/alexanderpann/iets3.opensource).
What do you think about the structure? Is anything missing? I thought about adding some screenshots (e.g. components, state machines) to the readme but wanted to hear your opinions first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,4 +25,148 @@ For details see [wiki](https://github.com/IETS3/iets3.opensource/wiki/Supported-
 - MPS 2020.3 (branch: [master](https://github.com/IETS3/iets3.opensource/tree/master)) 
 - MPS 2020.2 (branch: [maintenance/mps20202](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20202)) 
 - MPS 2020.1 (branch: [maintenance/mps20201](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20201)) 
-- MPS 2019.3 (branch: [maintenance/mps20193](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20193))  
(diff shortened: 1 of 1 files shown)


arimer:
Although I am not reviewer, I really appritiate the effort in updating the readme! I had a quick look and here are my two cents: 
*  I think the currently supported MPS version is already 2021.1 https://github.com/IETS3/iets3.opensource/blob/master/build.gradle#L66 we should update the readme before merging
* Screenshots are typically outdated quite fast, therefore I would try not to add them to the review, but rather have node URLs as you've already used them.
* I think the project structure itself is fine, what would be nice to have is a high level overview what the repository is actually about. Some short explanation on a really high level would be enough here. Maybe @markusvoelter has some ideas here to help.
* I think we should also add some information about, where the MPS-Project is actaully located on file system level, because sometimes people struggle to find it. At the moment it is under `code/languages/org.iets3.opensource`

alexanderpann:
As @arimer suggests, I added the currently supported version and an explanation on how to open the project in MPS to the readme. I don't feel qualified to write the high-level overview because I was not involved in the project.

lhartl:
I really appreciate your work here! 
Why do you use a fork? its not integrated in the build structure and therefore much harder to merge. As you see, the necessary green build is missing.

alexanderpann:
> I really appreciate your work here! Why do you use a fork? its not integrated in the build structure and therefore much harder to merge. As you see, the necessary green build is missing.

I guess that at the time that I created The PR, I either didn't know or didn't hat the permission to create a branch. I've created a new [PR](https://github.com/IETS3/iets3.opensource/pull/584) with the same content.

PR #512, opened 2021-09-14
Don't apply the checking rule of temporal language if it isn't used

### LINKED ISSUE #353: List equals operation yields typesystem error if temporal language is imported
See `ListEquality` in tests: http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F871690368833936470

If we import `org.iets3.core.expr.temporal` language, list equals operations have a typesystem error:
image

alexanderpann:
- fixes #353 

The reason why the condition is duplicated in the fix is that we need to stop the overriding of the rule and the rule itself. Using a condition in the block "applicable" doesn't work.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.temporal/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.temporal/models/typesystem.mps
@@ -22,7 +22,6 @@
     <import index="mhbf" ref="8865b7a8-5271-43d3-884c-6fd1d9cfdd34/java:org.jetbrains.mps.openapi.model(MPS.OpenAPI/)" />
     <import index="pbu6" ref="r:83e946de-2a7f-4a4c-b3c9-4f671aa7f2db(org.iets3.core.expr.base.behavior)" />
     <import index="xfg9" ref="r:ac28053f-2041-47f6-806b-ecfaca05a64a(org.iets3.core.expr.base.runtime.runtime)" />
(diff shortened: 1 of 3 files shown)


lhartl (changes requested):
With your approach. you check, whether none of the descendants of the `BinaryEqualityExpressions` contains a `ConstantRef`, which again has a `TemporalLiteral` as the value. This is a very specific approach and will solve the problem only for this very specific situation. Additionally, on such a ConstantRef operations can be applied that return a number. For these cases, we also don't want the checking rule to be applied.
 Generally speaking, we don't want the checking rule to be applied if neither the type of the right side nor the left side of the `BinaryEqualityExpression` is a Temporal type.

arimer:
@alexanderpann do your latest commits address the findings of @lhartl and do we still need this fix?
In case we want to keep it, pease change the targetbranch `maintenance/mps20213` and update your PR accordingly.

alexanderpann:
I am closing this PR. It is not reproducible in MPS 2021.3. If we can reproduce it in the future: Extract the content of \[typeof\\\_BinaryExpression\](http://127.0.0.1:63320/node?ref=r%3A80cf2246-750c-4158-9056-a619ebcf894c%28org.iets3.core.expr.base.typesystem%29%2F5115872837157417007) into a method with the annotation \`@InferenceMethod\` and when the left and right side is concrete \[here\](http://127.0.0.1:63320/node?ref=r%3A782cab7d-c30f-4797-991c-cb17d0274086%28org.iets3.core.expr.temporal.typesystem%29%2F1075037996907679863) in the overwriting rule and none of the types is of instance TemporalType, execute the extracted method.

PR #513, opened 2021-09-14
Return EmptyDateRange when dates are reversed in until operation

### LINKED ISSUE #261: An until from later to sooner should return an EmptyDateRange
assert /2019 02 01/.until[/2019 01 01/].isEmpty equals true [0 ms]

Right now it returns nonesense. This is easy to fix using ArbitraryDateRange.make() but is harder in the generated Java code. @lhartl please take a look and talk to @markusvoelter .

alexanderpann:
- fixes #261
- implemented for interpreter and generator

I made EmptyDateRangeValue extend ArbitraryDateRangeValue instead of AbstractDateRangeValue to make the code generator work. This hopefully shouldn't be an issue.


lhartl (changes requested):
The  implementation covers only the Arbitrary DataRangeValues but not the DiscreteDateRange values. Please take them into account as well. 

Please add tests for your implementation

alexanderpann:
@lhartl There is a test but I could add some more. 
Regarding the DiscreteDateRange: You probably mean that the count parameter could be negative and therefore an EmptyDateRange should be returned. There is not a single case in the generator where the user has an influence on this parameter because it is always set to 1. So why change it when it can't happen?

lhartl (approved):

PR #514, opened 2021-09-20
Improve type of box expression

### LINKED ISSUE #170: Type inference of box types incorrect
Context: Initialize a box type into a val, 
val globalcounter = box(10)

Now, the type of globalcounter is locked to number(10) implying any update operation throws a type system error

globalcounter.update(it + 1)

throws the error number(11) (inferred as the type of it+1 is) is not a subtype of number(10)

alexanderpann:
When the expression in the box expression has a specific number type, a generic integer type is used instead. Fixes #170 
An example can be seen in #170. Let me know if you see any additional problems.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.mutable/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.mutable/models/typesystem.mps
@@ -124,6 +124,9 @@
       </concept>
     </language>
     <language id="7a5dda62-9140-4668-ab76-d5ed1746f2b2" name="jetbrains.mps.lang.typesystem">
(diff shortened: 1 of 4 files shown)


AlexeiQ:
Why is a copy of the type `valueType.copy` necessary here (and in the previous implementation)? I'm not questioning your implementation, that looks fine, but would like to know the reason for it/ when to use it.

lhartl:
@alexanderpann the ticket was updated. Please update the PR accordingly.

lhartl (changes requested):
see comment above

alexanderpann:
@arimer I've added the requested warning. This PR could be merged.

arimer:
WIll check, thx.

arimer (approved):
New TS warning as suggested is added and tested 👍🏼

lhartl (approved):

PR #517, opened 2021-09-29
Bugfix/temporal value equals alternative

AlexeiQ:
Improve `TemporalValue.equals()` to compare values with null-value-slices, so it's returning `false` when a TemporalValue-slice has a null-value.

Problem: If two equal temporal values have a slice with a null-value, the equals() result is `false`


AlexeiQ:
We discovered, that this is a bigger problem than initially thought and it needs a more holistic solution, which is coordinated between several customer-projects. Until this is clarified and a possible solution discussion, this PR is frozen, in case we later can build upon this solution.

alexanderpann:
@AlexeiQ Can this PR be closed? Is this case covered by https://github.com/IETS3/iets3.opensource/pull/716?

AlexeiQ:
Yes, this it is and thus this branch/PR is obsolete

PR #520, opened 2021-10-07
Fix numberOfDecimals method in NumberLiteral

### LINKED ISSUE #519: Bug: numberOfDecimals behviour on NumberLiteral returns 1 more decimal place than entered
eg. on 12.85 numberOfDecimals returns 3 rather than 2

The behaviour calculates the number as length - index(delimiter) however actual decimal places start one step to the right of the delimiter so index needs to be incremented or result decremented by 1.

alexanderpann:
- fixes #519


alexanderpann:
> @alexanderpann could you please also add some tests to verify the behaviour? Obvously, we missed that for this method.

@wsafonov done.

mgronover (approved):

mgronover (approved):

PR #523, opened 2021-11-02
Deleted all empty generators

slisson:
When a custom generation plan is defined the model checker shows the error
'model uses language X with 1 generator(s), none of which is engaged with the active generation plan'.

A workaround would be to just include these languages in the generation
plan, but since they don't do anything it's unnecessary noise.

(cherry picked from commit 889d6b95ca399f2885566eed17ab00251908c5c3)


lhartl (changes requested):
Thank you for cleaning the empty generators up. 

The generators are removed in mps, but in the file system, they still exist. Please clean this up.

PR #524, opened 2021-11-02
Deleted all empty generators

slisson:
When a custom generation plan is defined the model checker shows the error
'model uses language X with 1 generator(s), none of which is engaged with the active generation plan'.

A workaround would be to just include these languages in the generation
plan, but since they don't do anything it's unnecessary noise.


wsafonov (changes requested):
Strange, we actually did this clean-up previously for all generators in commit 33f39267, but obviously some merge didn't work well somewhere and readded it. Anyway, due to [MPS-33093](https://youtrack.jetbrains.com/issue/MPS-33093) it's same issue as before with MPS 2020.3 - you need additionally to delete the generator files manually.

Also I plan to merge all maintenance branches into master for customer project migration, probably already this week. So this would land on master anyway soon - we can drop the other PR #523 unless you need the change on master asap.

slisson:
I checked that the model files are deleted, but they didn't exist.

wsafonov:
> I checked that the model files are deleted, but they didn't exist.

You're right, for some reason only the generator definition reappeared.

wsafonov (approved):

PR #525, opened 2021-11-03
datetime: add time concepts

coolya:
Add support for time similar to day. Time is "time on the wall clock" with no timezone info attached to it. 

Operations are similar to the existing data implementation but without the support for ranges. e.g. there is no support for `hour[22].contains(\22:32:00\)`. 

The generator is implemented to closely stick to the patterns of the date implementation. 

For technical details and design decisions take a look at the individual commit messages. 

This feature is additive and is a none breaking change.


lhartl (approved):
Looks good to me.
This feels like a nice addition to the datetime language. 
Nice detailed Tests :)

PR #527, opened 2021-11-18
Improve generator for sumOp to consider the collection base type

AlexeiQ:
Improve generator for sumOp to consider the collection base type

Problem: The generator of sumOp always returned the result as a BigDecimal,
which is not obvious and a different behaviour compared to the interpreter.

Solution: The generator considers the collection base type. In case it's a NumberType,
it depends on the numbers precision whether the result will be a BigInteger or BigDecimal.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
@@ -6,7 +6,6 @@
     <devkit ref="a2eb3a43-fcc2-4200-80dc-c60110c4862d(jetbrains.mps.devkit.templates)" />
   </languages>
   <imports>
-    <import index="hxvb" ref="r:ebd1eee1-ef85-4edc-907f-bc08f2f6b482(org.iets3.core.expr.genjava.simpleTypes.structure)" />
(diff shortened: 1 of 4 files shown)


AlexeiQ:
Because it's not possible to retrieve the generic type T of the PCollection<T> type in
CollectionHelper.sum() (see javas type erasure), the different cases are handled via
separate reduction-rules.

If there's a way to create tests for this change, please let me know.

lhartl (approved):
looks good to me. 

I had a look at the dependencies you mentioned. It also seems a bit weird to me, but when I reset your changes to the build script and execute the intention again, I come to exact same result. 
KernelF is relatively well tested, so I would rely on this for now. Since this only happens for devkits, maybe mps handles them in a different way now and you are just the first in the repo to run into this. I'll try to find out whether they changed something with one of the bugfix versions.

wsafonov:
> I've use the Reload Modules From Disk intention on the changed build-script. It updated multiple Devkits which seems weird, especially because language jetbrains.mps.baseLanguage.unitTest was added. Do you know whether it was just missing in previous commits? Shall I remove these changes?

You probably mean changes in multiple genjava devkits. This happens due to a MPS bug, which causes unstable resolving of the devkit dependencies - they bounce between two possible states. The best way to avoid unneeded changes in the build script is currently to execute the intention `Reload Modules From Disk` once again - you would see that the weird changes have disappeared, merely the order of the dependencies has been changed. If this is the case, you can safely revert such changes to avoid merge conflicts in the build script.

AlexeiQ:
Followed your advice @wsafonov. 
With the second intention-execution only the now commited small change is left (besides the differently ordered deps in two other solutions, which I reverted as suggested). This change was also available in the previously commited state.
Thx for the hint!

And thx to both of you for the review :)

PR #528, opened 2021-11-30
typedefs with units

lhartl:
It is possible to add a unit to a TypeDef. When such a TypeDef is used (TypeDefType) it is again possible to add a unit to the TypeDefType. Especially for units, this does not make sense, so this PR fixes the problem. We don't consider all possible tags with this fix, but only units, because there might be use cases for nesting other tags.


code/languages/org.iets3.opensource/languages/org.iets3.core.expr.typetags.units/org.iets3.core.expr.typetags.units.mpl
```diff
@@ -71,7 +70,7 @@
     <language slang="l:d8f591ec-4d86-4af2-9f92-a9e93c803ffa:jetbrains.mps.lang.scopes" version="0" />
     <language slang="l:0eddeefa-c2d6-4437-bc2c-de50fd4ce470:jetbrains.mps.lang.script" version="1" />
     <language slang="l:13744753-c81f-424a-9c1b-cf8943bf4e86:jetbrains.mps.lang.sharedConcepts" version="0" />
-    <language slang="l:7866978e-a0f0-4cc7-81bc-4d213d9375e1:jetbrains.mps.lang.smodel" version="17" />
+    <language slang="l:7866978e-a0f0-4cc7-81bc-4d213d9375e1:jetbrains.mps.lang.smodel" version="18" />
```
arimer, line 73: I am surprised about those changes. Did you execute any migrations?

arimer (approved):
I think the check itself is fine. 

- [ ] However we should add a simple local var for this expr: `ur.ancestor<concept = TaggedType>` and dont need to traverse the AST multiple times.

PR #529, opened 2021-12-14
Add JBR download via mps-gradle-plugin

### LINKED ISSUE #435: Decouple specific runtime JDK version for CL builds from local JDK version
**Problem/Drawback**
In the current state, developers are required to either have JDK 11 installed or configured to be able to perform a CL build of _iets.opensource_. See: https://github.com/IETS3/iets3.opensource/blob/master/build.gradle#L27

Due to the fact that JDK 11 is only required for the CL build it is often cumbersome (especially on windows) to switch to the required JDK version for "just" building the project from sources. 

**Improvement**
It would be great to loose this restriction for the CL build to allow building iets.opensource with any JDK supported by gradle (1.8+). To do so, the following would be required:

- [ ] download the required JDK via gradle for CL build
- [ ] set downoaded JDK as required JDK_HOME for build
- [ ] set JDK for MPS generated Ant files for tests

Notes:
- [ ] This should also be done for iets3.core

nkoester:
`download-jbr` was added post version 1.4 to the `mps-gradle-plugin`, hence it needed to be updated to a newer version. Checks for Java version is removed as we always use the correct version now.

This PR fixes #435

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         maven { url 'https://projects.itemis.de/nexus/content/repositories/mbeddr' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.2.175.cc60dc8'
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.6.281.3790039'
     }
 }
 
@@ -21,28 +21,15 @@ plugins {
     id 'co.riiid.gradle' version '0.4.2'
 }
 
-// Detect jdk location, required to start ant with tools.jar on classpath otherwise javac and tests will fail
-def jdk_home
+apply plugin: 'download-jbr'
 
-if (ext.has('java11_home')) {
-    jdk_home = ext.get('java11_home')
-} else if (System.getenv('JB_JAVA11_HOME') != null) {
-    jdk_home = System.getenv('JB_JAVA11_HOME')
-} else {
-    def expected = JavaVersion.VERSION_11
-    if (JavaVersion.current() != expected) {
-        throw new GradleException("This build script requires Java 11 but you are currently using ${JavaVersion.current()}.\nWhat you can do:\n"
-                + "  * Use project property java11_home to point to the Java 11 JDK.\n"
-                + "  * Use environment variable JB_JAVA11_HOME to point to the Java 11 JDK\n"
-                + "  * Run Gradle using Java 11")
-    }
-    jdk_home = System.getProperty('java.home')
+// configure jbr download
+downloadJbr {
+    // jbrVersion = '11_0_6-b520.66'
+    // jbrVersion = '11_0_9_1-b1145.77'
+    jbrVersion = '11_0_10-b1145.96'
 }
-
-// Check JDK location
-if (!new File(jdk_home, "lib").exists()) {
-    throw new GradleException("Unable to locate JDK home folder. Detected folder is: $jdk_home")
-}
+def jdk_home = "./build/jbrDownload/jbr"
 
 logger.info 'Using JDK at {}', jdk_home
 
@@ -161,6 +148,7 @@ repositories {
     mavenCentral()
 }
 
+
 task resolveMps(type: Sync) {
     dependsOn configurations.mps
     from {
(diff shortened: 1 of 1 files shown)


nkoester:
I am unsure about the third item mentioned in #435:

> - [ ] set JDK for MPS generated Ant files for tests

Can someone verify that this is also handled by my implementation?

arimer:
Your impl. works already with the generated ant xml files because of this line: https://github.com/IETS3/iets3.opensource/blob/master/build.gradle#L216. The CI (linux!) will already you the downloaded JDK/JBR for execution. 
However this will not work on mac/win I guess.

I think you would need to handle the different OS types explicitely to place the downloaded JDK in the corresponding folder.

coolya:
> Your impl. works already with the generated ant xml files because of this line: https://github.com/IETS3/iets3.opensource/blob/master/build.gradle#L216. The CI (linux!) will already use the downloaded JDK/JBR for execution. However this will not work on mac/win I guess.
> 
> I think you would need to handle the different OS types explicitely to place the downloaded JDK in the corresponding folder.

The is essentially what the plugin is doing. So no you don't need to do anything in you build script. You could optimise things a little because the download jar task exposes the location of the jar as a output you can then use to get the path but that's not super important.

coolya:
The path to the downloaded java executable is exposed `javaExecutable` on the task with the name `downloadJbr`.

arimer:
I am not entirely sure, but I think that the `download-jbr` is fetching already the right JBR version matching your OS.
After doing so, you need to configure the JBR depending on the OS where gradle is running:

For inspiration you can have a look here.

```
// precond. `download-jbr` ran already

if(Os.isFamily(Os.FAMILY_MAC)) {
    //weird layout for mac
    jdk_home = new File(yourJBRDir, "jbr/Contents/Home")
} else if (Os.isFamily(Os.FAMILY_WINDOWS)) {
    jdk_home = new File(yourJBRDir, "jbr")
} else if (Os.isFamily(Os.FAMILY_UNIX)) {
    jdk_home = new File(yourJBRDir, "jbr")
}
```
After configuring `jdk_home` properly, the build should run on every OS independing of the local java installation.

arimer:
Saw @coolya  comments too late 😅

nkoester:
This PR is now ready (thanks to @arimer  and @coolya for your help!).

Some more changes to the build file were needed but all in all nothing was broken. Generally a full cleanup and migration to Kotlin could be considered - but that is a story for a different day.

I do not intend to backport this change to previous MPS versions. Thus a fairly recent Java is required to build now.

@wsafonov @lhartl please let me know what you think and if there are things which unclear and need further attention from me.

build.gradle
```diff
@@ -207,15 +187,19 @@ if (gradle.startParameter.logLevel.toString() != "LIFECYCLE") {
     defaultScriptArgs.put('mps.ant.log', gradle.startParameter.logLevel.toString().toLowerCase())
 }
 
-def defaultScriptClasspath = project.configurations.junitAnt.fileCollection { true } +
-        project.files("$ext.jdk_home/lib/tools.jar")
```
arimer, line 211: I assume that due to the newer java version, we dont need to care about the tools.jar any longer?

build.gradle
```diff
@@ -393,21 +377,6 @@ publishing {
     }
 }
 
-
-task generateLibrariesXml(type: GenerateLibrariesXml) {
```
arimer, line 397: What is the reason for removing this task?

arimer:
Thanks for the effort! 💪🏼

build.gradle
nkoester: Yes, as far as I understand it.

build.gradle
nkoester: The `generateLibrariesXml` was never used in the task tree. The readme also does not refer to it. 

Generally speaking, this way of adding the dependencies is outdated anyways. Nevertheless, it might be that I deleted too much here and this was out of context for this PR. Maybe someone else (@coolya ?) can chip in on this?

coolya: It's a useless task anyway the libraries.xml is check in since 2021.1. If anyone needs to point the libraries to the sources they do that on their local machine. The task type is going to get removed from the plugin in version 2.x as well.

coolya (approved):

arimer:
This PR is replaced by: https://github.com/IETS3/iets3.opensource/pull/599

PR #532, opened 2021-12-19
Use BigInteger in interpreter of UpToTarget

alexanderpann:
The generator works with BigInteger, the interpreter does not and as a result cast-exceptions can occur.


lhartl (approved):
Looks good to me

PR #533, opened 2021-12-19
Fix EmptyMessageContent constraint error

alexanderpann:
- adds generated names to this nodes, otherwise a constraint error happens
- the name can't be null because then there is a constraint error for duplicate names ('null')


lhartl (approved):
Looks good to me

PR #534, opened 2021-12-20
Don't use generic Object type for result of collection operations

alexanderpann:
This pull request fixes cases where collection operations fall back to the Object type when calling collect(Collectors.toList()) in the generator and the original type is not available.

Example: Some operations (distinct, lastN) return list&lt;Object&gt; instead of list&lt;Days&gt; and cast errors occur.
```
qualified enum Days { A, B, C }                                                                                                                                                                 
                                                                                                                                                                                                
fun days() : list<Days> { 
  val ds = list(Days:A) 
  ds.distinct.distinct.lastN(10).firstN(10).reverse.withAll(ds.lastN(1)).with(Days:A).without(Days:B).pick[list(1)].where(|it == Days:A|) 
}
```


lhartl (approved):
Looks good to me

PR #535, opened 2021-12-20
Only look at RecordChangeTarget nodes in same model in mapping script

alexanderpann:
Is there a reason for` model.nodesIncludingImported(RecordChangeTarget`) instead of `model.nodes(RecordChangeTarget)`? I changed it because otherwise this exception happens:
`IllegalModelAccessError: You can write model only inside write actions`


lhartl (approved):
looks good to me. 
I think in this case we don't need the ` nodesIncludingImported()`

PR #536, opened 2021-12-21
Override baseType() in CollectionType

alexanderpann:
The original request was for implementing baseType() In ListType, but it seems to make sense to implement in CollectionType. There are two different baseType() methods, only one was implemented through inheritance.


lhartl (approved):
looks good to me

PR #537, opened 2021-12-21
Change layout of map literal(vertical editor)

alexanderpann:
Original request: 
_Can the list of KeyValuePairs in the MapLiteral editor be changed from identical to VerticalGrid? A map is usually useful with many elements - and there the current editor with a line or non-flush vertical termination just does not look nice._

I changed the indent layout to look like a vertical layout. I didn't like the original request because the separators would all be on a separate lines which doesn't look good. @lhartl what do you think of this change in general?

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.collections/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.collections/models/editor.mps
@@ -34,6 +34,7 @@
       </concept>
       <concept id="1106270549637" name="jetbrains.mps.lang.editor.structure.CellLayout_Horizontal" flags="nn" index="2iRfu4" />
       <concept id="1237303669825" name="jetbrains.mps.lang.editor.structure.CellLayout_Indent" flags="nn" index="l2Vlx" />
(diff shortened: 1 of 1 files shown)


lhartl:
Tbh, I don't like the new implementation. For me, it puts every key-value pair on a new line. I made a before/ after comparison. Based on that, I can not understand the request you got, because there are sufficient line breaks for me: 

Before(old implementation) 
Screenshot 2021-12-22 at 11 43 21

After (new implementation that comes with this PR): 
Screenshot 2021-12-22 at 11 42 23

alexanderpann:
@lhartl I think I misunderstood the request. I am guessing it is about the alignment when the key value pairs are not the same length, which you can't see in your example:
```
(100 -> 2 ,  2 -> 3,
 1   -> 2 ,  3 -> 4,
 3   -> 40,  1 -> 10)
```

I don't see an easy way to achieve this in a non hacky way.

coolya:
We could us a similar logic as for nested binary expressions. Where we switch to vertical layout when one expression is very long. Might be possible to do something similar here but not sure if it really satisfies the original request.

alexanderpann:
Closing because a different solution was found where a change to the platform is not needed.

PR #538, opened 2021-12-21
Insert helper code back into the SumOp-generator

AlexeiQ:
Insert code, which was outsourced into a helper, back into the generator for `SumOp` to remove runtime dependencies to MPS

The outsourced helper code creates dependencies to MPS in the runtime, which shall be avoided to maintain independence of MPS for the generated java code (and thus increase performance)


AlexeiQ:
I followed the build-script-dependency-trick from PR #527 to reduce changes.

lhartl (approved):
looks good to me

PR #540, opened 2021-12-23
Minor coverage fixes

alexanderpann:
- fixes empty cells in IMultiDecTab
- fixes lambda expression
- fixes short lambda expression
- ignores documentation comments

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
@@ -16790,27 +16790,41 @@
                 <node concept="1bVj0M" id="3$EgSSy1cEm" role="23t8la">
                   <node concept="3clFbS" id="3$EgSSy1cEn" role="1bW5cS">
                     <node concept="3clFbF" id="3$EgSSy1cEo" role="3cqZAp">
-                      <node concept="1Wc70l" id="3$EgSSy1cEp" role="3clFbG">
-                        <node concept="3fqX7Q" id="3$EgSSy1cEq" role="3uHU7w">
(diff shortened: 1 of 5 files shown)


lhartl (dismissed):
looks good to me

wsafonov (changes requested):
For both kinds of lambda expressions, `LambdaExpression` and `ShortLambdaExpression`, I don't think simply adding all descendants to the coverage is the right approach. One can think of conditional parts inside the lambda expression which would not be executed but still will be reported as covered. This is not correct in my opinion.

alexanderpann:
@wsafonov test cover function:
```
fun filterList(liste: list<struktur>) = liste.where(|outer: struktur => outer.wert > 0 
&& liste.map(|value: struktur => value.wert + outer.wert|).isNotEmpty 
&& liste.map(|it.wert + outer.wert + liste.map(|it.wert|).sum + if liste.isEmpty then 1 else 0|).isNotEmpty|)
```
It is called with a non-empty list and the coverage shows that the last 0 expression can never be reached and everything else is covered as it should be.  Let me know if you think I missed a test case.

wsafonov:
@alexanderpann regarding your test, one can of course imagine some more complex cases with conditional execution, but as far the basic stuff works, this should be ok for us.

I just have one question regarding the new code in `LambdaValue`:
```java
node<> originalNode = mapping.keys.findFirst({~key => mapping[key] == it; }); 
if (DefaultCoverageAnalyzer.hasNodeCoverage(it)) { 
   coverage.coverValue(originalNode, null); 
}
``` 
Shouldn't `isNodeCovered` be used here instead? Exactly same as in the code before.

wsafonov (dismissed):

wsafonov:
@alexanderpann can you resolve the conflicts? I'll merge then right thereafter

wsafonov (approved):

PR #542, opened 2022-01-10
Fix transform explicit lambdas into short lambdas

alexanderpann:
- replace only references to the original parameter
- disallow this intention for nested short lambda expressions because the **it** expression would be ambiguous
- fixes usage of **it** expression in nested lambda expressions
- let **it** parameter navigate to corresponding short lambda expression. It doesn't make sense to navigate to next Container (e.g. map) because it might refer to a different lambda expression (e.g. an outer short lambda expression)


wsafonov (changes requested):
In `transformIntoShortVersion` filtering out expressions with `ShortLambdaExpression` would be probably too restrictive, as this is a valid scenario. Instead, what would not work is having a `LambdaArgRef` inside a nested `ShortLambdaExpression` as it would wrongly refer to the argument of the nested `ShortLambdaExpression` after replacing it with `ShortLambdaItExpression`.

So I guess, we should exclude only this type of scenario instead of all with nested `ShortLambdaExpression`.

wsafonov (approved):

PR #543, opened 2022-01-13
Show failing items of manual test results in gutter

alexanderpann:
- Display error status of `AbstractTestItems` in error stripe in the same way type system errors are shown. 
Reason: failing test items are hard to find in long files without visual indication. 
- The results are only shown when the editor is currently opened


wsafonov (changes requested):
As disscussed today, we can adopt this approach for marking failed tests. However I have one concern regarding performance impact - calling the UI operation in the `AbstractTestItem.runManually()` operation might cause bad performance impact in scenarios where many items are tested altogether in a batch run, for example for a large test suite. Why not putting the code to the `ICanRunCheckManually.runManuallyWithUIUpdate()` method that is explicitly indended for UI updates?

wsafonov:
@alexanderpann in order to avoid code duplication, it makes sense to call super version of the `runManuallyWithUIUpdate()` method from the interface and just add your additional logic regarding the markings of the failed test results.

wsafonov (approved):

PR #546, opened 2022-02-10
Bugfix/lambdas in tracer

AlexeiQ:
Fix Tracer (Interpreter-Debugger) for (Short)LambdaExpressions

Problem: Traces for (Short)LambdaExpressions were shown broken and without links to the model, so clicking a trace to jump to the model wouldn't work.

Note: `LambdaValue.lambde` property changed from public to private


AlexeiQ:
I've change the property `LambdaValue.lambda` from public to private, without getter/setter, since it's used only internally. Should it probably be protected, in case this class is going to be extended? Though I doubt it.

lhartl (approved):
lgtm

Issue #550, opened 2022-03-01
Migrate 'Override basetype() in Collectiontype' to 2021.1


dbinkele:
in maintenance/MPS20203. the method 'basetype()' in 'Collectiontype' has been overriden. This changed the behavior drastically, as now the basetype is returned instead of CollectionType. 

P.S. This cause 57 tests in iets3.core (maintenance/MPS2020.3) to fail, which is now fixed. In order to fix this also in master (2021.1). the current change has to be made.

wsafonov:
Change is already in master.

PR #551, opened 2022-03-01
Override baseType() in CollectionType

dbinkele:
See #550 for details


wsafonov:
Change is already in master.

PR #554, opened 2022-03-15
string ops in the interpreter can now deal with nix

markusvoelter:
Updated some of the string operations to be able to deal with nix.


wsafonov (changes requested):
- in wrapped processor code, instead of accessing original value we should use NixSupport.getString(index), as it was meant to be used with the NixSupport.
- all operations should consistently take the path with invoking NixSupport I guess, otherwise we don't have exchangeable nix handling in those cases.
- please also change the first comment in the MR by putting short description there - obviously, the template text there is not clear enough, I will format it accordingly ;-)

markusvoelter:
3) done; sorry, hadn't even realized there was a template.
1) fixed
2) fixed

wsafonov (changes requested):
- could you extend existing string operation tests and also test those nix handling cases, just to make sure we have everying correct in place? 

Some cosmetics:
- to simplify the code you can use explicit `s.getString(int)` instead of custom cast `(string) s.get(int)`
- in the case where we need create BigInteger for int you can use BigInteger.valueOf(int) instead of detour with converting int to string

arimer:
PR is no longer maintained. Closing.

PR #556, opened 2022-03-24
Make test execution preferences overridable in workspace settings

alexanderpann:
This PR adds new actions to the run menu that allows changing the test execution mode (interpreter, generator) also per workspace (persisted in workspace.xml instead of model).
Reason: The original TestExecutionConfig might be changed by a user and checked into version control. Now all other user get the new preference when they check out the project.

After this change the workspace setting is preferred over the TestExecutionConfig when it is set by the user.
The combined preference should then be queried by calling`TestExecutionSettings.getExecutionModePreference` (for example in TestSuites).


AlexeiQ:
There are again these "scrambled" build-script-changes. If you re-execute "Reload Modules from Disk" the real changes will be shown. For more infos, see [Wladis comment](https://github.com/IETS3/iets3.opensource/pull/527#issuecomment-995832203)

lhartl:
I think this solution is nicer than the one we have. 👍 

We should consider removing the old one completelyI think. Since I don't know which projects might use the old one and how they use it, I am ok with keeping it for now.

lhartl (approved):

PR #558, opened 2022-04-04
Feature/improve value inspector

markusvoelter:
Some syntactic improvements in value inspectors


wsafonov:
Looks good to me, just one question: why did you remove and readd the `InlineAll` intention under `inlineAll2` name?

markusvoelter:
Because there was a problem with this intention - so it looked. In my desparation, I reimplemented it. But the problem was a different one.

wsafonov:
> Because there was a problem with this intention - so it looked. In my desparation, I reimplemented it. But the problem was a different one.

Does it make sense for me to have a look what could be that problem? Because the new intention looks exactly the same as the old one, except for child filter. If not, we can then at least give it the old name, cause having only single one `inlineAll2` intention looks strange for later reviews.

markusvoelter:
No. It was a version incompatibility.

markusvoelter:
If you care about the "2" in the name, you can remove it.

mgronover:
> If you care about the "2" in the name, you can remove it.

I removed it.

mgronover:
This PR is closed (target branch is not supported any more) and will be continued by #1122

PR #560, opened 2022-04-07
Don't allow slash as first character in CorrdCellRef

### LINKED ISSUE #559: Spreadsheets: adding sheet with '/' does not work
Can't enter `$S2/ B1` in the editor. Seems like the optional grammar cell does not work in this place?

image

alexanderpann:
Fixes #559


sergej-koscejev:
This does not fix #559. The problem is if you go to http://127.0.0.1:63320/node?ref=r%3Ac26f3707-1ecb-413e-ba84-cd641318e760%28playground.sheets%29%2F3922243018846595131 and try to type this same assertion from scratch on another line - it won't work.

alexanderpann:
You are both right. At the time, it seemed to work, but with the current master it certainly doesn't. The problem is related to the optional cell, but I couldn't find a nice solution where I don't introduce an unnecessary space in the editor. 
example: $ AnotherExampleSheet/A0

PR #562, opened 2022-04-07
Exclude commented elements from the referencable scope

wsafonov:
This PR ensures that the scope of visible, referencable elements doesn't include commented out elements - referencing such elements should produce an out-of-scope error.

I have also deliberately changed the query for visible elements to look only in the top-level content instead of filtering all descendants for those which are not in a comment annotation. It seems to me that including only the top-level contents should be the correct approach. I don't really know why the visible scope was previously implemented using all descendants version. @markusvoelter if you have any opinion or objection on that, please comment.


markusvoelter:
The descendants was used because at some point we had Sections that could contain elements.

wsafonov:
> The descendants was used because at some point we had Sections that could contain elements.

Are those sections are still around? Cause I'm only aware of section marker that don't have own content. Then the assumption remains that only top-level content could be visible from a chunk.

lhartl:
Are the sections you refer to @markusvoelter the sections in Lodas?

alexanderpann:
I am closing this PR as it was already merged as far as I can tell.

PR #563, opened 2022-04-07
JavaGen: Invoke generated closure of BlockExpression with compact invoke function expression

alexanderpann:
Reason: The implementation of `InvokeFunctionOperation` -> invoke() calculates the return type of the closure literal differently than the short invoke operation -> (). In fact, the calculation is wrong:
Example from base language:
Result: Object (inner lambda return type is calculated as nulltype)
```
{ => 
  if (true) { 
    return { => return "hi"; }.invoke(); 
  } else { 
    return null; 
  } 
}.invoke()
```

Result: String (correct)
```
{ => 
  if (true) { 
    return { => return "hi"; }();
  } else { 
    return null; 
  } 
}.invoke()
```

I've only changed the generator of the BlockExpression because this is the only case where the wrong calculation happens.


lhartl (approved):
That was definitely a bug. 
looks good to me

PR #564, opened 2022-04-25
Computation trace: Don't always show LambdaExpression

alexanderpann:
The lambda expression was shown as the child of every other node because the flag for also show was set. There was another fix in [mbeddr.core](https://github.com/mbeddr/mbeddr.core/pull/2216) to make this fix possible.

How to reproduce and check: Execute a test case with lambda expressions ([example](http://127.0.0.1:63320/node?ref=r%3Af6b93d14-1af1-4f84-a11b-cbe2d8c5efff%28test.in.expr.os.lambda%40tests%29%2F8293738266734930343)). Right click first assert statement -> Show trace and Click Expand All button. The ShortLambdaExpression shouldn't be displayed ~10 times anymore.

Screenshot from other project:

withShortLambda

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.lambda/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.lambda/models/plugin.mps
@@ -1022,9 +1022,7 @@
                 <node concept="37vLTw" id="6ITtBskYEku" role="37wK5m">
                   <ref role="3cqZAo" node="6ITtBskYEke" resolve="originalLE" />
                 </node>
-                <node concept="3clFbT" id="5d4Vabvs8X9" role="37wK5m">
-                  <property role="3clFbU" value="true" />
-                </node>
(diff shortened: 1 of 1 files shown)


lhartl:
Is the mbeddr fix available for iets3.os currently? The branch you linked for the fix in mbeddr went to master. 
mbeddr master is already migrated to mps 2021.2. (see: https://github.com/mbeddr/mbeddr.core/blob/master/build.gradle line 21/22) 
This results in version numbers for mbeddr starting with 2021.2.xxx
Screenshot 2022-04-25 at 13 24 33

iets3.os uses mbeddr versions 2021.1.+ (see https://github.com/IETS3/iets3.opensource/blob/master/build.gradle line 68) meaning the latest 2021.1 build. So your changes are not available in iets3.os I think. Did you merge your mbeddr changes to the 2021.1. maintenance branche of mbeddr as well?

alexanderpann:
Last question was answered internally (answer: yes).

lhartl (approved):
LGTM

PR #565, opened 2022-04-25
Set cartesian product of tuple element types as super types of TupleType

alexanderpann:
Screenshot 2022-04-25 at 11 24 38

The supertype calculation for the list fails because the tuple types of the first and second literal are considered as different because of the first number (100 vs. 100.4). The only common super type  is `TupleType`.

To understand the issue, let's compare the typeof implementation of the kernelF `ListLiteral` with the baselanguage `ArrayLiteral`. In the `ArrayLiteral` the element type get's calculated through an inequation statement: `infer elementType :>=: typeof(item);`. In the ListLiteral it gets calculated through a call to PTF.computerSupertype with [this](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F3237442461102096418) implementation for simple types. As a result all the other typesystem mechanics that affect typesystem equations  are not applied and no common supertype can be found except the default implementation for tuples: `new node<TupleType>()`. We can't implement this rule the same way it is done in baselanguage because we use our own `Type` concept which seems to be incompatible with the type variables as far as I can tell. 

As a workaround, I have implemented the super type of the `TupleType` as the cartesian product of all the types of the containing items. This way the list can also find a common supertype.
Example:

Screenshot 2022-04-25 at 11 40 47

Screenshot 2022-04-25 at 11 28 34

This way a common supertype can be found for the example: [real, string, boolean] because the common type of 100 and 100.4 can be infered as `real`.

Let me know what you think of this implementation or if you think that the calculation is too costly.


lhartl (approved):
LGTM
Since you changed the TS, please add tests for it

dbinkele (approved):

PR #569, opened 2022-05-16
Improve expression and temporal op editors

alexanderpann:
This PR adds autocompletion support for operation expression in the core, collections and temporal language and fixes and issue with optional cell in the temporal language. 

I've changed the editors of the temporal language so that the operation context can be overriden by the optional expression (e.g. time) and also deleted when the cursor is at the end of the context. The ability to override the context was lost after a change in the optional cell in  grammar cells where we fixed a bug where side transformations where erroneously added to neighboring cells that shouldn't have a transformation (in this case a transformation to the right and **left** side of the context cell).

There are other languages that also have the problem that the alias doesn't support code completion. Please let me know if my approach is the right one before I fix all instances.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
@@ -127,11 +127,15 @@
       <concept id="1164833692343" name="jetbrains.mps.lang.editor.structure.CellMenuPart_PropertyValues" flags="ng" index="PvTIS">
         <child id="1164833692344" name="valuesFunction" index="PvTIR" />
       </concept>
+      <concept id="1078938745671" name="jetbrains.mps.lang.editor.structure.EditorComponentDeclaration" flags="ig" index="PKFIW" />
       <concept id="1078939183254" name="jetbrains.mps.lang.editor.structure.CellModel_Component" flags="sg" stub="3162947552742194261" index="PMmxH">
         <reference id="1078939183255" name="editorComponent" index="PMmxG" />
       </concept>
       <concept id="1235728439575" name="jetbrains.mps.lang.editor.structure.BaseLineCell" flags="ln" index="2R9Tw8" />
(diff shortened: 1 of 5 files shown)


lhartl:
I am not 100% sure yet what the problem was and what you achieved with this fix. 
Is it maybe something similar to the editor experience of the BinaryExpression you want to achieve here? 
Binary expression alias handling: http://127.0.0.1:63320/node?ref=r%3A8405f486-53b5-4fe6-af3e-7f68358bd631%28org.iets3.core.expr.base.editor%29%2F1247721522496321673

lhartl:
What was wrong with this test case? Why was it flaky? Since when is it flaky? I never noticed this. Is there no option to fix the underlying problem instead of deleting the test?

alexanderpann:
> What was wrong with this test case? Why was it flaky? Since when is it flaky? I never noticed this. Is there no option to fix the underlying problem instead of deleting the test?

I introduced it in the tuple PR. The editor shows a type system error for one test statement but after pressing F5 the correct type gets calculated. The only time it worked was when I merged the PR. I also muted the test on TC so that other branches don't fail.

alexanderpann:
> I am not 100% sure yet what the problem was and what you achieved with this fix. Is it maybe something similar to the editor experience of the BinaryExpression you want to achieve here? Binary expression alias handling: http://127.0.0.1:63320/node?ref=r%3A8405f486-53b5-4fe6-af3e-7f68358bd631%28org.iets3.core.expr.base.editor%29%2F1247721522496321673

- I Changed the editors because there was a bugfix for GrammarCells (optional cells) that changed the behavior of the SideTransformations. I changed the editors so that the old behavior is restored.

- Other languages: I changed the editors for operations (e.g., list operations) so that you can use code completion. Constants like list.size could be changed with CC before, but list.map could not. You had to delete the operator before you can use another operator. I also tried again to implement this as with BinaryExpression with grammar.substitute grammar.constant (subconcept.concetAlias), but there are editors where readonly model accessor cells are also used, which do not work in combination with grammar cells.

lhartl (approved):
Ah, ok. 
Lgtm

PR #570, opened 2022-05-23
Show lambda transform intentions also in child nodes

alexanderpann:
The intention is now also available in child nodes because otherwise the intention only works, for example when the cursor is placed after the second pipe (|) symbol.


lhartl (approved):
Since both intentions are already limited to the concrete Concept LambdaEcpression or ShortLambdaExpression, I don't think you need the child filter in this case. 
please check on this before merging

alexanderpann:
@lhartl It is needed for nested lambda expressions.

PR #573, opened 2022-06-13
Units: Rename quantity metre to length

### LINKED ISSUE #572: Wrong quantity (metre) in Library SIUnits
In org.iets3.core.expr.typetags.units.si.units.SIUnits
the _quantity metre_ should be _quantity length_ . 
See e.g. https://en.wikipedia.org/wiki/International_System_of_Units

alexanderpann:
Fixes #572.


lhartl (approved):
Lgtm

PR #574, opened 2022-06-28
CoordCellRef: Fix adding the optional finder

### LINKED ISSUE #559: Spreadsheets: adding sheet with '/' does not work

alexanderpann:
This PR replaces the not working optional finder cell of the CoordCellRef editor with a transformation menu so that the finder can be added through the code completion menu of the cell instead of typing "/" before the cell.
I changed it because I couldn't get the optional cell working. The missing spaces in the editor (punctuation-x: true) prevent the side transformation to be executed. Fixes #559.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.repl/models/org/iets3/core/expr/repl/constraints.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.repl/models/org/iets3/core/expr/repl/constraints.mps
@@ -2,6 +2,7 @@
 <model ref="r:098cd5d2-6719-4a9b-9976-ee98e642cc00(org.iets3.core.expr.repl.constraints)">
   <persistence version="9" />
   <languages>
+    <use id="3f4bc5f5-c6c1-4a28-8b10-c83066ffa4a1" name="jetbrains.mps.lang.constraints" version="6" />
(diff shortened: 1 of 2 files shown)


brucetrask:
@alexanderpann and @sergej-koscejev  Thanks very much for the fix. I tried it and it works great. Should we create an issue for the broken optional cell in the grammar cells language? I don't know where that language is or know much about it but I heard it is something useful I should be using.  Is it also part of iets3.opensource?

brucetrask:
@alexanderpann and @sergej-koscejev  Ok. Thanks. One last follow up just to make sure I am understanding (I am a bit new to MPS). So, could I use the optional grammar cell facility in this sheets use case or is it that use case which is not compatible with the grammar cells I have with 2021.1? In other words, you mentioned you couldn't get the optional cell working originally. Is that because the use in the sheets is not possible or something else? Just wanted to understand.

arimer:
@alexanderpann I am not sure here. Where the questions/findings of @wsafonov already addressed in this PR or are they missing?

alexanderpann:
IMO, I've addressed all points.

alexanderpann:
@arimer I've rebased the PR for 2021.3. It could be merged.

PR #575, opened 2022-06-28
bugfix: override renderReadable() where missing

pavel-nedvedicky:
To get correct representation of expressions as string, renderReadable() must be overriden for concept.
Implement this method for 3 concepts.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -301,6 +301,18 @@ publishing {
                 }
             }
         }
+        maven {
+                name = "itemisCloud"
+                url = version.toString().endsWith("-SNAPSHOT")
+                        ? uri("https://artifacts.itemis.cloud/repository/maven-mps-snapshots/")
+                        : uri("https://artifacts.itemis.cloud/repository/maven-mps-releases/")
+                if (project.hasProperty("artifacts.itemis.cloud.user") && project.hasProperty("artifacts.itemis.cloud.pw")) {
+                    credentials {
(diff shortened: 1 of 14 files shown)


lhartl:
The build is not running green yet, so we dont consider this PR ready for review. Since the build seems to have hung up, please push a no-op commit (e.g. add an empty line somewhere). When the build is running through, we can review the PR.

wsafonov (changes requested):
Thanks for the contribution! Just one minor change request to make the code cleaner: in `AbstractMinMaxExpression.renderReadable()` there is no need to convert to list. You can also use join operation for the sequence of strings with a custom delimiter. This gives not only consistent code style, but more important avoids list object creation (which can cause some unwanted performance losses in scenarios with large number of method calls).

pavel-nedvedicky:
@wsafonov thank you for the feedback. Changes implemented, hopefully the build will be triggered now.

pavel-nedvedicky:
@lhartl is there something I can do to trigger the build? Seems it won't run again. 
Could it be caused by PR from forked repo?

lhartl:
yes, this is due to the fork. Is there a reason why you prefer to use the fork? In your case I would not recommend to fork because your project uses the regular iets3.opensource as a dependency. 
If it's ok for you, please use a regular branch.

pavel-nedvedicky:
Closing due to not running build.

PR #577, opened 2022-06-29
Fix makeRecord interpreter off-by-one error

sergej-koscejev:
There's an off-by-one error in the interpreter for makeRecord caused by incrementing an index too early.


brucetrask:
Thanks @sergej-koscejev . Your fix works for me.

wsafonov (approved):

PR #578, opened 2022-06-29
Fix not working conditional editors

### LINKED ISSUE #576: type in enum declarations can't be set anymore
In an enum declaration it used to be possible to add a type by typing a < after the name. This doesn't work anymore. There is also no intention for this.

Example:

Screenshot 2022-06-29 at 08 47 18

most likely this is due to the fix in grammar cells (I did not investigate though)

alexanderpann:
I've fixed #576 but also checked all other optional cells and fixed some other issues that are related to these cells. More info can be found in the commit messages.


lhartl (approved):
Lgtm

lhartl:
Thanks a lot @alexanderpann !!!

PR #585, opened 2022-07-14
bugfix: override renderReadable() where missing

pavel-nedvedicky:
Replaces https://github.com/IETS3/iets3.opensource/pull/575 which didn't run the build

To get correct representation of expressions as string, renderReadable() must be overriden for concept.
Implement this method for 3 concepts.


pavel-nedvedicky:
@wsafonov build finally running, could you review please? Should be OK as we implemented proposed changes on previous PR. Thanks

wsafonov (approved):
Looks good now.

PR #586, opened 2022-07-20
Temporal language: Fix NullPointer in reduce method by introducing a default constraint type

alexanderpann:
Return an optional default value that is calculated from a type constraint on the temporal literal.
This constraint has to be set when no slices are provided. Previously, the reduce method returned null and failed with a `NullPointerException`.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.temporal/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.temporal/models/editor.mps
@@ -88,6 +88,7 @@
       </concept>
     </language>
     <language id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage">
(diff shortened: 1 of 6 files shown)


wsafonov (changes requested):
I have some minor points:
- in `typeof_TemporalLiteral`, instead of creating a new list of nodes for type computation, why not handle the special case with an empty slices list and provided type constraint in a if clause? This should make the logic more explicit and avoid superfluous list creation during the type computation. Mavbe peanuts here, but still unnecessary effort.
- the type constraint is only shown in the editor if there are no slices. And it's not checked for type-compatibility with the slices. It might theoretically happen that the added slice values are not type-compatible with the specified type constraint, but there will be no error shown and the type constraint will be used unwitting for the user to set the default value of the temporal value, which might lead to a wrong behaviour at some point. Idealy we should add the required type check and also always show the type constraint in the editor if it's actually set giving the user the chance to remove it.

PR #587, opened 2022-07-25
Improve IRunManuallyWithUIUpdate logic

alexanderpann:
I introduce an additional method `performAdditionalEditorUpdate` in `ICanRunCheckManually` where concepts can contribute further editor updates such as displaying the test result in the error stripe.
Also the editor is updated now only once after all checks are done.
This is a big performance improvement for huge test suites such as [collections](http://127.0.0.1:63320/node?ref=r%3A2309ed17-e7b4-45b5-b25e-2c0f3ea87e8b%28test.in.expr.os.collections%40tests%29%2F7740953487930555503).


wsafonov (approved):
LGTM, hopefully we got it right now from the performance aspect of possible exection scenarios.

PR #588, opened 2022-07-26
Fix substitution menu of various tables + small fixes for LookupTable

### LINKED ISSUE #539: Pasting in tables can throw exceptions
Table editors with substitute node functions are affected (e.g.DataRow, TableRow...). In most cases, the code assumes that the substituted node is an Expression but when a node of another type is pasted, a cast exception happens for the cell root concept.
The implementations of this method look a bit suspect to me in general because a new value is always added, even if a value exists for this cell. This bug was probably never found because the code is not called in cases when an expression is substituted for another expression. Especially in cases where a wrapper node is used as a value holder this exception can be observed.

alexanderpann:
This PR fixes #539.

I've changed the substitution menu, so that they work, when an expression is paste into a table cell. The problem is that an expression is pasted to a cell in the table and these concepts are of course incompatible (a cell contains an expression).
I've also fixed the `LookupTable`, so that I can at least test my changes.


alexanderpann:
A quick note about the used custom factory cell: the fix for the hiding of the data holding cells is not possible in the tables languages because it might break something in other projects.

wsafonov (changes requested):
As discussed today in a call, changes for DecTab and MultiDecTab miss adjustments of the selectable style for the tables and have therefore to be checked again.

wsafonov (approved):
Looks good now.

PR #589, opened 2022-08-03
added smart reference interface to Unit concept and implemented get description text

baran-goru:
`org.iets.core.expr.typetags.units` : implemented smart reference to the Unit concept with dimension information in the description text, so you can see the dimension information int the completion menu:
2022-08-03 15_28_35-com cso enabler core – tech cariad arche tutorial_tech cariad arche tutorial fea


wsafonov (approved):
LGTM

PR #591, opened 2022-08-08
Migrate to MPS 2021.2

alexanderpann:
Observed reported changes/bugs:
- sub list operation checks for valid range ([MPS-34628](https://youtrack.jetbrains.com/issue/MPS-34628/fromIndex5002-toIndex2))
- Running tests with the generator fail because there is an error in generation. Workaround: Build model yourself and then run tests (very annoying; [MPS-34445](https://youtrack.jetbrains.com/issue/MPS-34445/Error-during-executing-provider-MakejetbrainsmpsexecutionconfigurationspluginSolutionpluginMakeNodePointersBeforeTask)). 
- Messages view for Make is not cleared on the next make/build ([MPS-34402](https://youtrack.jetbrains.com/issue/MPS-34402/Messages-view-for-Make-is-not-cleared-on-the-next-makebuild)).
- An action must not request write-action during actions update([MPS-34373](https://youtrack.jetbrains.com/issue/MPS-34373)). Solution: Only use read action when updating an action (e.g. updating the text in a toggle action).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,10 @@ We take care of porting changes done in older supported MPS versions to the newe
 For details see [wiki](https://github.com/IETS3/iets3.opensource/wiki/Supported-MPS-Versions).
 
 ### What MPS versions are currently supported? 
-- MPS 2021.1 (branch: [master](https://github.com/IETS3/iets3.opensource/tree/master)) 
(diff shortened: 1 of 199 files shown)


kbirken:
@wsafonov It would be good if you could review this migration PR soon...

wsafonov (approved):
Looks good, let's ship it!

PR #592, opened 2022-08-10
Add a check for duplicated enum literals for primitive and enum literals

alexanderpann:


arimer:
I had just a quick look and here some comments from my side: 

- how about adding tests for the newly interoduced check? This test case seems to be a proper place: [enums](http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F3760701216320903292)
- Does it really make sense to enable that check for enumliterals of boolean type? 
- the impl. is quite expensive. Because we are traversing all literals multiple times in multiple for-loos and to compare pair-wise. How about we simplify this a little bit by just using hashsets where we store the string representation of the primitive literals and check for duplicates? At least for the primitive literals part this should be quite ease.

arimer:
@alexanderpann  FYI: I've added some tests covering the current impl. of the uniqueness check.

arimer (dismissed):
Changes look good to me 👍🏼 

Just a tiny remark. 
- This test [withDuplicatedIntValues](http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F1127195381204429581) you've updated is identical to the first one `withDuplicatedNumberValues`. So we could either remove it or maybe change the types to floating points. 

This [TestAddElementOperation](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959032e%28jetbrains.mps.baseLanguage.collections.structure%29%2F5699792037748043353) is new to me but comes in handy.  Thx for sharing :) 

What is the excact semantic of this operation? I could not find it in the jetbrains user guide.

alexanderpann:
@arimer The `testAdd` method calls the method `add` of the used Java collection and returns true if adding was successful. How it works depends on the used collection class: [documentation for HashSet add method](https://docs.oracle.com/javase/8/docs/api/java/util/HashSet.html#add-E-)

arimer (approved):

PR #595, opened 2022-08-31
Generator fix for #586 + some more tests

alexanderpann:
This is the same fix as #586, but this time for the generator and not for the interpreter.


wsafonov (approved):
LGTM

PR #596, opened 2022-08-31
build: removed publication to old itemis nexus and added dep. to new itemis cloud

arimer:
- In the scope of migriting away from https://projects.itemis.de/nexus we now remove the publication to the old repository.
- Instead we now publish all artifacts to https://artifacts.itemis.cloud/
- In addition the version of the mps-gradle-plugin was upgraded to the latest MPS 2020.3 version which is available on the new cloud.


arimer:
Sorry, I am closing this PR again. I will target the oldest maintenance branch first and then merge from old -> new ones to get the changes.

PR #597, opened 2022-08-31
build: replaced old itemis nexus by the new itemis cloud nexus

arimer:
- In the scope of migriting away from https://projects.itemis.de/nexus we now remove the publication to the old repository.
- Instead we now publish all artifacts to https://artifacts.itemis.cloud/
- now only new nexus is used for dependency resolution
- In addition the version of the mps-gradle-plugin was updated to the next available on the new cloud.
- mavenLocal() repo was removed from dependency resolution as a problematic one

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,11 +6,10 @@ import java.time.LocalDateTime
 //will pull the groovy classes/types from nexus to the classpath
 buildscript {
     repositories {
-        maven { url 'https://projects.itemis.de/nexus/content/repositories/mbeddr' }
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.2.175.cc60dc8'
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.2.177.46341c0'
     }
 }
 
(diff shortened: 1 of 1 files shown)


wsafonov:
This looks good, but let's wait with merge of these changes due to possible need of current builds in the downstream iets3.core for some customers that don't have yet proper configuration for the new Nexus. Just in case something would need to be fixed in iets3.os for them as well and to avoid making things complicated then. 

We could probably also remove the old Nexus repos from the dependency resolution in this PR later as well.

arimer:
Ok for me. I change that PR into draft to not merge it by accident.

alexanderpann:
There is just one issue with this approach: the master build fails now at the publishing stage to the old nexus (readonly). As far as I can tell, as a result, the artifact also doesn't get published to the new nexus server.

arimer:
@wsafonov @alexanderpann I think it time to have a look at this  one.

arimer:
Tbh. I am not sure about the target branch here. Should we start with the oldest and then cherry-pick to the more recent ones?

wsafonov:
> Tbh. I am not sure about the target branch here. Should we start with the oldest and then cherry-pick to the more recent ones?

I would suggest, we start with the oldest maintenance branch and do then a cascaded merge to newer branches. And we also remove the old nexus urls in this PR completely. Time to say goodbye ;)

wsafonov (approved):

PR #598, opened 2022-09-01
Run migrations

alexanderpann:
Run pending migrations


arimer (approved):
Thanks for applying the pending migrations 👍🏼

alexanderpann:
@arimer Unfortunately not, I haven't found the cause.

PR #599, opened 2022-09-06
build: Add JBR download via mps-gradle-plugin

arimer:
This PR picked up the work of https://github.com/IETS3/iets3.opensource/pull/529 and rebased it onto master.
All insides could be find in the original one. 

**Sumup**
It decouples the gradle build from any Java version installed on the host machine. The defined java version is downloaded via the mps-gradle-plugin and used during execution.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,14 +4,13 @@ import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
 
-
 //will pull the groovy classes/types from nexus to the classpath
 buildscript {
     repositories {
         maven { url 'https://projects.itemis.de/nexus/content/repositories/mbeddr' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.2.175.cc60dc8'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -21,32 +20,33 @@ plugins {
...
-    jdk_home = System.getenv('JB_JAVA11_HOME')
-} else {
-    def expected = JavaVersion.VERSION_11
-    if (JavaVersion.current() != expected) {
-        throw new GradleException("This build script requires Java 11 but you are currently using ${JavaVersion.current()}.\nWhat you can do:\n"
-                + "  * Use project property java11_home to point to the Java 11 JDK.\n"
-                + "  * Use environment variable JB_JAVA11_HOME to point to the Java 11 JDK\n"
-                + "  * Run Gradle using Java 11")
+repositories {
+    mavenLocal()
```
sergej-koscejev, line 29: Let's use this opportunity to remove `mavenLocal()` here, it shouldn't be listed as a repository. See https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:case-for-maven-local

sergej-koscejev (approved):

build.gradle
arimer: Done.

PR #605, opened 2022-09-21
[minor] Collections (interpreter): make eval arg class check less strict

alexanderpann:
Runtime exceptions can occur because of this check.


wsafonov (approved):
LGTM.

PR #607, opened 2022-10-07
[minor] Temporal: Make type constraint in temporal literal optional 

alexanderpann:
The type constraint that disappears is too confusing for the user. Now it works similar to the type constraint in list literals.


wsafonov (changes requested):
Looks good, just small enhacement idea for the typesystem: because we cast `typeConstraint` to `Type`, it would be more straightforward to use instanceOf(Type) condition in the if clause which would replace both checks for being not null and not abstract

wsafonov (changes requested):
typeConstraint.isNotNull check is of course also not required anymore :)

alexanderpann:
> typeConstraint.isNotNull check is of course also not required anymore :)

Yes I noticed. It seems that I am not good at reading instructions ;)

PR #608, opened 2022-10-07
[minor] Expr base: Refactor MessageValue

alexanderpann:
The getter was missing for the kind field.


wsafonov (changes requested):
Please use refactoring for renaming because it would also update resolveInfos in the used references. Otherwise those changes would be done automatically later by MPS on models save and appear later in different unrtelated MRs.

wsafonov (approved):

wsafonov:
> > resolveInfos
> 
> Done. I didn't think about the resolveInfos.

Yeah, that is typically forgotten and actually not a hard requirement. But then you get suddenly those spurious changes later in models that you haven't touched and start wondering where that comes from...

PR #612, opened 2022-10-13
Make EnumLiteral documentable

alexanderpann:
The additional workaround in the commit is needed because of a MPS layout bug (not celllayout). Video of the bug:

https://user-images.githubusercontent.com/88385944/195582276-2de1087e-1674-4df4-acae-7d8ca7c3e1a6.mov

It is probably related to  the`if` cell in the EnumDeclaration where a switch of layout between an indent layout and an vertical grid layout happens. Only the EnumerLiteral that caused the change gets relayouted but not the other EnumerLiterals.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/editor.mps
@@ -10,6 +10,7 @@
     <use id="3bdedd09-792a-4e15-a4db-83970df3ee86" name="de.itemis.mps.editor.collapsible" version="-1" />
     <use id="1919c723-b60b-4592-9318-9ce96d91da44" name="de.itemis.mps.editor.celllayout" version="-1" />
     <use id="120e1c9d-4e27-4478-b2af-b2c3bd3850b0" name="com.mbeddr.mpsutil.editor.querylist" version="0" />
+    <use id="52733268-be24-4f5f-ab84-a73b7c0c03b0" name="de.slisson.mps.richtext.customcell" version="0" />
     <devkit ref="fbc25dd2-5da4-483a-8b19-70928e1b62d7(jetbrains.mps.devkit.general-purpose)" />
   </languages>
   <imports>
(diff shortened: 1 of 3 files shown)


arimer:
I think the metamodel changes in general look fine, however the layout still looks weird in case we use typed enumerations with default values: 

grafik

wsafonov (changes requested):
Frankly after looking at the current editor definition of the `EnumDeclaration`, I'm asking myself why this has been done that way with the whole redundant definition of the editor. The only reason we need the second branch is due to the grid layout we want to have for literal values. The outer part should behave equally. This means that using indent layout for the outer part and only applying if condition for the inner literals collection should be enough. This also seems to solve layouting issue for me. At least I could not reproduce the broken layout behaviour with the aforementioned editor structure (however not sure if it was reproducable for me at all). Regardless of its impact on the layout behaviour, it makes sense to simplify the editor.

alexanderpann:
@wsafonov do you mind pushing your changes? I don't quit get what you changed and how you didn't end with the problematic layout in the screenshot.

wsafonov:
> @wsafonov do you mind pushing your changes? I don't quit get what you changed and how you didn't end up with the problematic layout in the screenshot.

I have pushed the suggested change on branch `feature/enumdecl-editor-cleanup`. But again, for me the broken layout behaviour is actually not reproducible on the current `maintenance/mps20211` branch (I even tried adding empty lines after the enum as with your example):

https://user-images.githubusercontent.com/8157816/200914714-796666bc-a58e-49ce-8de6-68f18262e24e.mp4

Maybe this is platform-specific?

alexanderpann:
I don't know how this problem could be platform-specific. I've tried your branch, but we still have the issue shown in the screenshot with typed enumerations with default values where the editor looks wrong for the first enum.

wsafonov:
> I don't know how this problem could be platform-specific. I've tried your branch, but we still have the issue shown in the screenshot with typed enumerations with default values where the editor looks wrong for the first enum.

Ok, but that would be a different issue with the way how the documentation annotation editor works and not the one with the unstable layout. Does the original issue shown in the video still occurs on that branch for you?

alexanderpann:
> Does the original issue shown in the video still occurs on that branch for you?

No.

arimer:
@alexanderpann what is missing in this PR? Is this PR already alligned with the branch `feature/enumdecl-editor-cleanup` proposed by @wsafonov ?

alexanderpann:
Nothing.  I can't re-request a review somehow.

arimer (approved):
In general the documentation feature looks fine to me. 
I think it is a nice improvement 👍🏼

wsafonov (approved):
@alexanderpann so it contains now the cleaned up version of the editor from the `feature/enumdecl-editor-cleanup` branch that is not needed anymore, right? Looks good then.

PR #613, opened 2022-10-14
Feature/stringvalidation

markusvoelter:
I have implemented a string validation DSL for Big D from Nuremberg. Includes interpreter and generator. 

Tests here: http://127.0.0.1:63320/node?ref=r%3A4498a3f1-ffdc-46cb-b225-63bf7159e69e%28test.in.expr.os.stringvaliation%40tests%29%2F1922523186439131488

New Modules: 

image


wsafonov (changes requested):
- `org.iets3.core.expr.genjava.stringvalidation` generator language should go under `expr.genjava` virtual folder
- `XCannotBeCheckKind` should be along with other CheckKinds into the `check.kind` virtual package
- `PositionIndicatorFirst` / `PositionIndicatorLast` concepts both define non-used `value` property
- `FailCheck` concept should be along with other OccurenceBasedChecks under `check.occurence` package
- `IfExistsOccurenceConstraint` concept name is not consistent with its `each` alias - concept name should be better aligned with that alias, also the Java runtime class `XIfExistsOccurenceConstraint`.
- `ValidateStringExpr` editor looks a bit awkward in a dot expression with that additional `|` at the end before the dot, but I guess that because there is no proper caret support for the custom bracket cell, right? Maybe we could fall back to using differently-sized bracket characters here instead of the the custom bracket cell. We should also try adding caret support for the custom bracket cell, but this would be out of scope here.
- also for more consistent concept names, let's rename `StringResultOkOp` -> `ValidateStringResultOkOp`, `StringResultMessagesOp` -> `ValidateStringResultErrorsOp`
- to keep the languages cleaner, let's drop the empty generator of the stringvalidation language, as it won't be required
- for the `StringResultMessagesOp` generator (Ex2ExStringVal switch), instead of generating a closure, using a direct expression like `Empty.vector().plusAll(COPY_SRC)` as in the interpreter should also work
- in `Constants` class there is a duplication of adding `!` as a special char - maybe there was some other char supposed to be added instead?

Apart from that all looks good 🙂

markusvoelter:
Oh, and I tried to implement your suggestion regarding plusAll and the avoidance of the closure. Doesn't work, can't get the type checker to shut up. I have reverted it to the closure.

wsafonov:
> Actually the name is correct. If each thing exists, then the next
> constraint is checked. I am struggling with the alias.

Ok, after having another look I take my suggestion with `ForEach` for the occurence constraint concept name back - it actually could be misleading to think about this pattern in terms of each occurence. Because then `[each digit [must occur more than 2 times]]`  from the common sense could be interpreted as "whenever a digit occurs, it must occur more than 2 times" and non-appearing of any digit in "aaaaa" is ok, as there are no digits and hence the pattern is not invalid. Anyway, I understood the struggling with the concept alias. As long as it's fine for the customer, we can merge it.

wsafonov (approved):

PR #617, opened 2022-10-25
Make duplicate check in enum declarations customizable

### LINKED ISSUE #615: Breaking change in EnumLiterals associated value
This summer has been introduced I change (I think in this commit https://github.com/IETS3/iets3.opensource/commit/9302cc810e667b4db1e8fad6d678cd2799072cfd by @alexanderpann ) to values associated to EnumLiterals, preventing to associate the same value to more than one literal.

For example, this test would fail with the current implementation of IETS3 while it used to pass:
Screenshot 2022-10-18 at 15 38 18

We have actually used enums in this way for years, and for our use-cases this is actually useful.

Would you reconsider going back to the previous behavior?
Alternatively, would it be possible to disable it by using some flag on enum?

I would be happy to provide a PR for this, in case this can be useful

alexanderpann:
- There is now an extension point typesystemCustomization that can change the behavior.
- Fixes #615
- The default behavior is to check for duplicates


arimer (approved):
Changes look good now!
Thanks a lot for the refactorings 👍🏼

PR #618, opened 2022-11-07
Bugfix/minor fixes 20211

alexanderpann:
This PR collects minor bug fixes where a separate PR is overkill. It can be merged anytime.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <model ref="r:8405f486-53b5-4fe6-af3e-7f68358bd631(org.iets3.core.expr.base.editor)">
   <persistence version="9" />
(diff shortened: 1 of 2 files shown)


arimer (approved):
Looks fine to me. 
Just a minor question: Why exactly das a [EnumLiteralRef](http://127.0.0.1:63320/node?ref=r%3A1d6bd88a-7393-4b32-b0e6-2d8b3094776e%28org.iets3.core.expr.toplevel.editor%29%2F6867966982819126294) needs to be `auto-deletable`?

alexanderpann:
> Looks fine to me. Just a minor question: Why exactly das a [EnumLiteralRef](http://127.0.0.1:63320/node?ref=r%3A1d6bd88a-7393-4b32-b0e6-2d8b3094776e%28org.iets3.core.expr.toplevel.editor%29%2F6867966982819126294) needs to be `auto-deletable`?

There are cases where you press delete and an empty cell with a red cursor is left behind instead of deleting the expression. NumberLiteral, for example, also needs auto-deletable. Property-Cells are auto-deletable by default but in EnumLiteralRef we are using a `R/O model access cell` that's why it is needed I believe.

PR #619, opened 2022-11-09
Interface to hide arbitrary nodes in the editor.

### LINKED ISSUE #614: Provide mechanism to hide nodes in the editor, but delegate the hide logic to other level
The visualization of nodes in editors may depend on some logic that is implemented outside the language definition. 
By providing some interface, the hiding logic can be delegated to other layers but tested in the concept editor where the language is defined.

Use case: 
The preview of nodes in editors may depend on the availability of those nodes according to the selection status in a variability model.
Unselected nodes could be hidden, reducing the cognitive load during artifact analysis.

mfsoliveira:
Contribution:
- Provide an interface that allows for a user to hide a node in the editor, but the logic for hiding it is delegated to client code.

The interface name is ICanHide and is located in org.iets3.core.base.
It provides 3 methods:
- boolean isVisible(), used to query the intended visibility state
- void hide(), used to indicate the node should be hidden
- void show(), to indicate the node should be shown

The boolean state is temporarily saved in a userObject, thus it is not persisted within the model.
The handling of user object is closed to the provided methods so that no additional string objects are created.

The change adds a feature and shall not break compatibility.

This PR closes #614

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
@@ -41,6 +41,10 @@
       <concept id="1225194691553" name="jetbrains.mps.lang.behavior.structure.ThisNodeExpression" flags="nn" index="13iPFW" />
     </language>
     <language id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage">
+      <concept id="1224071154655" name="jetbrains.mps.baseLanguage.structure.AsExpression" flags="nn" index="0kSF2">
+        <child id="1224071154657" name="classifierType" index="0kSFW" />
+        <child id="1224071154656" name="expression" index="0kSFX" />
+      </concept>
       <concept id="1080223426719" name="jetbrains.mps.baseLanguage.structure.OrExpression" flags="nn" index="22lmx$" />
       <concept id="1082485599095" name="jetbrains.mps.baseLanguage.structure.BlockStatement" flags="nn" index="9aQIb">
         <child id="1082485599096" name="statements" index="9aQI4" />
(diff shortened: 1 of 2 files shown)


wsafonov:
Just really minor thing to make the code even more simple :)
Insteaad of having if statement, assign to a Boolean variable `isVisible` and then directly return `isVisible == null || isVisible` (no need to call booleanValue()).

kbirken:
In addition to @wsafonov 's remark here is one optimization: Due to the logic in `isVisible()` visibility is `true` if no user object is attached. This is fine. But because of this the `this.show()` code in the constructor can be removed. This would avoid that user objects are added by default to each implementor of this interface (saves time and space).

Minor remark: There are some typos in the javadoc :-)

mfsoliveira:
@wsafonov Thanks Wladimir. I refactored the isVisible, but as I removed the constructor, I cannot test null || visible. The isVisible shall default to true if no userObject is defined.
@dbinkele Thanks Daniel. I removed the constructor. As the logic in isVisible changed, it got outdated and was no longer needed. I hope there is no more typos.

wsafonov:
@mfsoliveira not quite sure what you meant by removing the constructor, as it's unrelated - I just thought about simplifying the result statement to `(visible == null) || visible;` leveraging the logical short-circuit evaluation.

kbirken:
After a discussion with @wsafonov and @mfsoliveira we decided that a better location for this interface is mpsutil. It does not have any interaction with KernelF-specific things and should thus be put in some more generic layer. On the other hand, it is not mature enough to be put into mps.extensions. 

The corresponding issue is [#2244](https://github.com/mbeddr/mbeddr.core/issues/2244) in mbeddr.core.

kbirken:
Converted to "Draft" state as this PR will eventually not be merged. There will be a replacement PR in mpsutil.

wsafonov:
> Converted to "Draft" state as this PR will eventually not be merged. There will be a replacement PR in mpsutil.

Then let's just close it 🙂 The branch remains there anyway.

PR #624, opened 2022-12-02
Extend DataTableLookUp with optional argument: default row

AlexeiQ:
# What changed
- `DataTableLookUp` is extended with an optional argument `defaultRow`. This argument allows to specify a default row for the lookup, so in case the looked for value was not found in the searched column, the default row will be returned. Unless the column contains an `empty` value, then the row of the `empty` is returned. (If the look up doesn't specify a column for the search, the tables default column is used)
- Tests for `DataTableLookUp` and `DataTable` are cleaned up and improved

## Limitations
- Resolved with [45b6c56](https://github.com/IETS3/iets3.opensource/pull/624/commits/45b6c5656ae3e8d2a4cfcead352a99d15268148c):   
When the `DataTableLookUp` doesn't specify a default column, but uses the tables default column and is about to add the `defaultRow` argument, the code completion menu suggests the `, ` two times. image Currently it's too hard to resolve this:
  - Either remove the menu item by removing the [IDotTarget_TransformationMenu](http://127.0.0.1:63320/node?ref=r%3A8405f486-53b5-4fe6-af3e-7f68358bd631%28org.iets3.core.expr.base.editor%29%2F2853962202203059943). This has an impact on every `IDotTarget`
  - Add a transformation menu to the `DataTableLookUp`, but then other invalid entries are proposed
  - This can even be influenced by MPS changes, when calculations are changed on how the entries for the menu are accumulated

# Todos
- [x] Clarify whether the changed behavior regarding the default is okay (@wsafonov, @mgronover, @AlexeiQ )
- [x] Clarify the `#alias#` versus constant-cell question
- [x] Add release comment
- [x] Remove Draft, add Reviewers and `Prio`-label

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.data/models/org.iets3.core.expr.data.behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.data/models/org.iets3.core.expr.data.behavior.mps
@@ -44,9 +44,21 @@
       <concept id="1082485599095" name="jetbrains.mps.baseLanguage.structure.BlockStatement" flags="nn" index="9aQIb">
         <child id="1082485599096" name="statements" index="9aQI4" />
       </concept>
+      <concept id="1215693861676" name="jetbrains.mps.baseLanguage.structure.BaseAssignmentExpression" flags="nn" index="d038R">
+        <child id="1068498886297" name="rValue" index="37vLTx" />
+        <child id="1068498886295" name="lValue" index="37vLTJ" />
+      </concept>
+      <concept id="4836112446988635817" name="jetbrains.mps.baseLanguage.structure.UndefinedType" flags="in" index="2jxLKc" />
+      <concept id="1173175405605" name="jetbrains.mps.baseLanguage.structure.ArrayAccessExpression" flags="nn" index="AH0OO">
+        <child id="1173175577737" name="index" index="AHEQo" />
+        <child id="1173175590490" name="array" index="AHHXb" />
+      </concept>
       <concept id="1197027756228" name="jetbrains.mps.baseLanguage.structure.DotExpression" flags="nn" index="2OqwBi">
         <child id="1197027771414" name="operand" index="2Oq$k0" />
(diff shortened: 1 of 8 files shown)


AlexeiQ:
The lastest changes look good and are working ([45b6c56](https://github.com/IETS3/iets3.opensource/pull/624/commits/45b6c5656ae3e8d2a4cfcead352a99d15268148c)  + [02f1a08](https://github.com/IETS3/iets3.opensource/pull/624/commits/02f1a0837309653c799efcf16a30d999d88f6583)).
This also resolves the initial limitation. Great Job!

AlexeiQ:
**Question: Do we have an convention when to use `#alias#`-cell and when to use a simple constant-cell?**
In this case the `lookUpBy` and `default` cells are `#alias#`, which means, they are editable, but don't delete the node. Regular constant-cells are not editable and can delete the complete node, when the curser is on that cell.  `lookUpBy` is an `iDotTarget` like `oneOf` or `toList` and those use the constant-cell instead of an alias. Thus we have different behavior and I'm wondering if we have some convention when to use what to have a consistent behavior.

AlexeiQ:
The `default`-part of the `DataTableLookUp` cannot be deleted, when the cursor is placed on the `default` keyword. Deleting from the `=` and from eg. first position of the value works fine. As Discussed, this should be fixed as well.

mgronover:
> The `default`-part of the `DataTableLookUp` cannot be deleted, when the cursor is placed on the `default` keyword. Deleting from the `=` and from eg. first position of the value works fine. As Discussed, this should be fixed as well.

This is fixed now. But when the caret is placed right behind `val1` (s. below), pressing `DEL` has no effect.
image
The user has to move the caret to the right side of the `,` to have the expected delete operation (both `DEL` and `Backspace` are working here).

According to (https://github.com/IETS3/iets3.opensource/pull/624#issuecomment-1382066670) the `#alias#` editor field was exchanged by a constant to make the editor action workable (`#alias#` with `read-only` style does not to call the required actions) .

arimer:
@AlexeiQ, @mgronover in case we need this please change the target branch to at least maintenance/mps20213

PR #626, opened 2022-12-05
Fix `add comment` transformation action in auto completion

AlexeiQ:
# Change-log
The transformation action applyCommentsToIDocumentable to add a comment to an IDocumentable node is improved to only be available depending on IDocumentable.isDocumentationAllowed(). Furthermore it only jumps over IEmpty nodes to reach an IDocumentable node instead of jumping as well over non-IDocumentable nodes.

# Problem
- The `applyCommentsToIDocumentable` contribution menu to add comments to nodes isn't customizable in the sense, that it's not possible to deactivate it for specific concepts, though there is `IDocumentable.isDocumentationAllowed`. It's not enough to just remove the `IDocumentable` interface, because it can be applied implicitly via several other interfaces, due to the interface hierarchy.
- Further more in this case the `execute` method is executed even though `can execute` returns `false`.
- The action adds a comment to the next node via `next-siblings` and jumps over `IEmpty` nodes, which is okay, and non-`IDocumentable` nodes, which is not okay, because  it may seem to the user totally random.
- Additionally, if the targeted next node, i.e. the node that should get a comment, already has a comment, the current node ( = `IEmpty` node) is detached, which is unexpected and weird. 

# What changed
- Added a check, so `IDocumentable.isDocumentationAllowed` is considered and it's possible to specify which concepts can be commented.
- Added additional checks in `text` and `execute` methods in the `applyCommentsToIDocumentable` contribution menu, to circumvent the bug disregarding the `can execute` result
- The search for the target node stops at the next non-`IEmpty` sibling node.
  - If it is an `IDocumentable` and doesn't have a comment, a comment is added to it.
  - If it has a comment, the cursor will jump into that comment. 
  - If the target node is not an `IDocumentable` the transformation-text `//` is not recognized and shows up as an error.

# Todos
- [x] Add release comment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - Computation of the least-common-supertype for expressions with different return types has been fixed. The typesystem now correctly infers a join type (c.f. [original issue](https://github.com/IETS3/iets3.opensource/issues/505))
-- Naming constraint of IValidNamedConcept is customizable [original issue](https://github.com/IETS3/iets3.opensource/pull/631)
+- Naming constraint of IValidNamedConcept is customizable [original request](https://github.com/IETS3/iets3.opensource/pull/631)
+- Made transformation action [applyCommentsToIDocumentable](http://127.0.0.1:63320/node?ref=r%3A80fb0853-eb3b-4e84-aebd-cc7fdb011d97%28org.iets3.core.base.editor%29%2F5981628904839421072) only applicable if documentation is allowed [original request](https://github.com/IETS3/iets3.opensource/pull/626)
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


mgronover:
I would prefer writing a behavior method for IEmpty to check if the next sibling that is not an IEmpty is documentable. 
The contribution menu could use that in all of its parts (`text`, `can execute` and `execute`).

The method itself would simply contain what is done in the can execute part:

image

**Problem of this idea:**
IEditor is not a concept of IETS3 but of mbeddr.

mgronover (approved):

mgronover:
I would prefer writing a behavior method for IEmpty to check if the next sibling that is not an IEmpty is documentable.
The contribution menu could use that in all of its parts (text, can execute and execute).

The method itself would simply contain what is done in the can execute part:

image


Problem of this idea:
IEditor is not a concept of IETS3 but of mbeddr.

mgronover (approved):

arimer (approved):
Changes look fine. The contribution menu behaves better now 👍🏼

PR #627, opened 2022-12-05
Implement missing base types

alexanderpann:
I am not sure about the TaggedType. It also has a base type, but there is a type system check that fails when the method is implemented.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
@@ -5594,6 +5594,24 @@
         <ref role="ehGHo" to="hm2y:6sdnDbSla17" resolve="Expression" />
       </node>
     </node>
+    <node concept="13i0hz" id="2TQFEoQTY_T" role="13h7CS">
+      <property role="TrG5h" value="baseType" />
+      <ref role="13i0hy" node="XhdFKv3UAU" resolve="baseType" />
+      <node concept="3Tm1VV" id="2TQFEoQTY_U" role="1B3o_S" />
+      <node concept="3clFbS" id="2TQFEoQTY_Z" role="3clF47">
+        <node concept="3clFbF" id="2TQFEoQTYG9" role="3cqZAp">
+          <node concept="2OqwBi" id="2TQFEoQTYSw" role="3clFbG">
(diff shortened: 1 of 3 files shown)


lhartl:
which typesystem check fails when the method is implemented for tagged types?

alexanderpann:
https://build.mbeddr.com/buildConfiguration/Iets3_BuildOrgIets3core/333366?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildTestsSection=true

arimer (changes requested):
I just had a closer look to the problem. Chaning the behaviour impl. of `TaggedType.baseType()` would change the semantic of what a base type of TaggedType should be. **Before** the change the baseType of a TaggedType was the TaggedType itself. **After** the change we delegate to its child `TaggedType.baseType`. To be honest, I am not 100% sure what is the right impl. here. In case we change it, our internal API would calculate a different expected base type for TaggedType and might break on runtime, because it is expecting an other type.

This is excactly the problem for the breaking test, because the test is relying on the fact that:
[ur.ancestor<concept = TaggedType>.baseType:TypedefType.baseType()](http://127.0.0.1:63320/node?ref=r%3Abf3cd5a0-eefc-4fd9-b3a6-b57643c9d80c%28org.iets3.core.expr.typetags.units.typesystem%29%2F4020487592981850840)is delgating to [TypeDef.baseType()](http://127.0.0.1:63320/node?ref=r%3Acfd59c48-ecc8-4b0c-8ae8-6d876c46ebbb%28org.iets3.core.expr.toplevel.behavior%29%2F1103723565285356950) which again invokes the (now changed) `baseType()` implementation on an instanec of TaggedType, resulting in a different returned value.

I would argue, in case the want to override the default implemention of [Type.baseType()](http://127.0.0.1:63320/node?ref=r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%29%2F1103723565285353914) of any subconcepts of `Type`, we must ensure that the calculated runtime type is correct. This would mean, that code which expects the (old) default impl. of `baseType()` would need to be updated to handle the (new) different return type.

I dont think we can merge this PR withouth the required changes sketched above.

alexanderpann:
I am closing this PR in favor of #723 and I'll keep the `TaggedType` as it is.

PR #628, opened 2022-12-05
Restore pasting in the empty expression editor

alexanderpann:
Paste certain expressions into an editor where an expression was expected still didn't work.
The wrong paste action was used (the one of EditorCell_Label instead of the generic node paste action).


lhartl (approved):
lgtm

PR #629, opened 2022-12-13
Make IRecordDeclaration equality check overridable

alexanderpann:
This way, an extending record declaration can use a custom equality check for special cases where two record should be treated as equal (e.g., two records have the same structure but have different version numbers).


lhartl (approved):
lgtm
great that you adjusted the EqualsHelper as well

PR #630, opened 2022-12-13
Support hexadecimal number literals and hexadecimal number ranges

alexanderpann:
The hexadecimal numbers work similar to the ones in baselanguage. They are considered integers and also support negative values. Internally, they are converted to decimal whenever they are used, that means that the typesystem also converts them to decimal. They are only displayed in hexadecimal form in the editor and the left side of the typesystem debug tool.
Hex values are also now supported in NumberRangeSpecs e.g., `[-0xff | 0xff]`.

Additional fixes:
- `NumberRangeSpec` now validates through the `NumberLiteral` class.
- Precision expression for integers are now not only supported in the generator but also in the interpreter.

Example:

Screenshot 2022-12-13 at 12 41 33

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/plugin.mps
@@ -1153,6 +1153,19 @@
         </node>
       </node>
     </node>
+    <node concept="2tJIrI" id="4NJAaUg11Bg" role="jymVt" />
+    <node concept="3clFb_" id="4NJAaUg1hYw" role="jymVt">
+      <property role="TrG5h" value="allowHexadecimalNumbers" />
+      <node concept="3clFbS" id="4NJAaUg1hYz" role="3clF47">
+        <node concept="3clFbF" id="4NJAaUg1ie9" role="3cqZAp">
+          <node concept="3clFbT" id="4NJAaUg1ie8" role="3clFbG">
+            <property role="3clFbU" value="true" />
+          </node>
+        </node>
+      </node>
+      <node concept="3Tm1VV" id="4NJAaUg1fSe" role="1B3o_S" />
+      <node concept="10P_77" id="4NJAaUg1hXd" role="3clF45" />
+    </node>
     <node concept="2tJIrI" id="2q0DACtJfgT" role="jymVt" />
     <node concept="3Tm1VV" id="2Qbt$1tSnqi" role="1B3o_S" />
   </node>
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.simpleTypes/generator/template/org.iets3.core.expr.genjava.simpleTypes@generator.mps
@@ -3,6 +3,7 @@
   <persistence version="9" />
   <languages>
     <use id="7a5dda62-9140-4668-ab76-d5ed1746f2b2" name="jetbrains.mps.lang.typesystem" version="5" />
+    <use id="b401a680-8325-4110-8fd3-84331ff25bef" name="jetbrains.mps.lang.generator" version="4" />
     <devkit ref="a2eb3a43-fcc2-4200-80dc-c60110c4862d(jetbrains.mps.devkit.templates)" />
   </languages>
   <imports>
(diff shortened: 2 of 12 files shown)


lhartl:
@alexanderpann is there an easy way to exclude this from projects which don't work with hex numbers? This seems rather specific for me. Based on your description, the hex numbers are integrated in the usual number types. I'd suggest to have this in a separate concept, ideally in a separate language and plugin and ship this as optional add on. 
Since this feature was not requested in multiple projects in over 5 years of KernelF, I think it's dangerous to "force" this in the customer projects by adding it to the existing number types/ Languages.

alexanderpann:
@lhartl In baselanguage it is indeed an extra literal that extends IntegerLiteral.  I don't see it that way in KernelF as it is just another way of entering an integer number (in base 16 instead of base 10 to be precise). We don't even want to generate different code for it. The value is automatically converted to base 10 when needed (the typesystem also sees it as a normal integer), as it is done in getCompileTimeConstantValue of the `HexIntegerLiteral` in baselanguage. Do you really want to change 100s of places to support the `HexIntegerLiteral` instead of seeing it a special notation of the `IntegerLiteral`?
We could make the hex literals opt-in through the extension point [primitiveTypeMapper](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F1085994689350268573) or another extension point if you prefer it that way. As long as you don't use them, nothing will break. Every change in this PR is meant to be backward compatible.

arimer (changes requested):
I think there are still some things to clarify before we can merge.

**Findings and questions**
1. how about we use an elseif() instead of an else + nested if statement [here](http://127.0.0.1:63320/node?ref=r%3Aac5f749f-6179-4d4f-ad24-ad9edbd8077b%28org.iets3.core.expr.simpleTypes.behavior%29%2F2754665542854598744)?
2. could you maybe explain[ this particular code part](http://127.0.0.1:63320/node?ref=r%3Aac5f749f-6179-4d4f-ad24-ad9edbd8077b%28org.iets3.core.expr.simpleTypes.behavior%29%2F1021978185991894596). Why do we need to handle hexadecimals in the `valueWithDotInsteadOfComma()` function at all? Is it even possible that hex numberscontaining comma/dot is used?
3. Maybe we could also drop a comment [here](http://127.0.0.1:63320/node?ref=r%3Aac5f749f-6179-4d4f-ad24-ad9edbd8077b%28org.iets3.core.expr.simpleTypes.behavior%29%2F8700435542243260314) explaining that you need to get rid of the `(-)?0x` prefix before parsing
4. Should it be possible to combine/mix hex. and dec. values in the ranges like `number[10|0x64]`? In case I do it in the  [TestDefautlValues](http://127.0.0.1:63320/node?ref=r%3A136a38b9-9b4d-4cdb-82ce-de7f17f13658%28test.in.expr.os.numbers%40tests%29%2F6932772747670725918) test I see an error: 
5. grafik
6. The [getMin()](http://127.0.0.1:63320/node?ref=r%3Aac5f749f-6179-4d4f-ad24-ad9edbd8077b%28org.iets3.core.expr.simpleTypes.behavior%29%2F1021978185995354424)and [getMax()](http://127.0.0.1:63320/node?ref=r%3Aac5f749f-6179-4d4f-ad24-ad9edbd8077b%28org.iets3.core.expr.simpleTypes.behavior%29%2F1021978185995364680) impl. in `NumberRageSpec` looks a little bit weird, because u need to create a newly, not needed node of `NumberLiteral` which is only used to call `isValidNumber()`. Would it be possible to make the ` isValidNumber()` static or is it too much of a rework?
7. There might be a problem with the **backwards-compatibility** of the newly added hex values in `NumberRangeSpec`. In case customers/users start using hexademimal literals in their models,  but their implementation is still using direct property access via the [min](http://127.0.0.1:63320/node?ref=r%3A6d93ddb1-b0b0-4eee-8079-51303666672a%28org.iets3.core.expr.simpleTypes.structure%29%2F1330041117646892911) and [max](http://127.0.0.1:63320/node?ref=r%3A6d93ddb1-b0b0-4eee-8079-51303666672a%28org.iets3.core.expr.simpleTypes.structure%29%2F1330041117646892912), errors will appear in case they expect decimals, because they will receive the hex values. Adding getters for min/max which delegate to your `getMinimum() / getMaximum()` implementation would fix this, but at the same time would diactivate the entire hex-value feature. I dont see a proper solution here :(
8. The logic of the `isValid` impl. of the max property is now different. Initially we considered a value valid in case it was parseable, now we dont do this any longer: 
9. grafik
10. How about we stick to the old behaviour and do the following `Double.parseDouble(new node<NumberLiteral>().set(propertyValue).valueWithDotInsteadOfComma());` to keep the old implementation?
11. No need to use the beh. methods for [setMax()/setMin()](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F1021978185995697783) in `SimpleTypesPrimitiveTypeMapper`. The overriden setters should take care of it.
12. Same is true for the changes in [unary](http://127.0.0.1:63320/node?ref=r%3A050f6d52-a81b-4b31-9a1c-531c1a04708e%28org.iets3.core.expr.simpleTypes.typesystem%29%2F1021978185995758211). In case get/set methods are overriden properly we could in 7) we could keep the old impl.
13. In `MultiDecTabPasteHelper`, how about we return the same `testLiteral` [here](http://127.0.0.1:63320/node?ref=r%3Aecf4180a-6e34-47e0-90d1-b073b95f6f24%28org.iets3.core.expr.util.intentions%29%2F2754665542857860690) instead of creating an other new node?
14. In ExprSimpleTypesInterpreter this [change](http://127.0.0.1:63320/node?ref=r%3A78e88ebb-2d27-4b89-867f-623c50619338%28org.iets3.core.expr.simpleTypes.interpreter.plugin%29%2F8700435542242517680) looks unintentional. Why not keep the old cast?!grafik

I am not sure how and if we can overcome the problem listed in 7). We need to make sure to make users aware of this before merging. 
Also figuring out how much effort it would be for current users/clients to update their implementation would be beneficial before merging. Maybe @wsafonov or @lhartl can give an overview here.

arimer:
The implementation itself is nice and using the constraints to allow additional values (hex ones) is a good non breaking approach. 
However the main problem I see here is that as soon as users are allowed to use the new hex values, their old and not updated implementation (in calculations, generators etc.) would break. So introducing this feature "explicetly" and on purpose would be really important here. Meaning we would need to give customers the posibility to:

1. update their impl. using min/max first
2. then give them the possibility to "switch on" the new feature

IMHO shipping it "under the hood" is not a good approach and might bring more trouble then value.

HeikoBecker:
> @HeikoBecker any progresson this one?

No, this PR is currently blocked.

arimer:
@HeikoBecker, thx for the update. I converted it to a draft PR. In case you continue working on it, feel free to change the status.

HeikoBecker:
Closing this PR in agreement with @alexanderpann since it is now quite outdated.

alexanderpann:
Let's close it. I'll reimplement it. I already have a better idea how to not make this a breaking change.

PR #631, opened 2022-12-14
Feature/apostrophe can be disallowed in IValidNamedConcept implementations

mgronover:
### Requirement for this PR
A customer requires to disallow an apostrophe within a name of IValidNamedConcept implementations.
IValidNamedConcept was modified to allow/disallow an apostrophe
at the second place (and following) within its name. 

### Changes:
- Refactoring of the constraint (condition is now given by a behavior method) to have wider test possibilities
- Test cases for legal and illegal names were added.  
- New concepts implementing IValidNamedConcept were introduced which are used in the test cases. These override the different allow___() methods to have all different cases.

### Compatibility
- fully given: the default behavior is kept

### Migrations
- not needed


AlexeiQ:
Modified `IValidNamedConcept` to allow/ disallow the usage of an apostrophe in the name, at the second place (and following). It's allowed per default, so the default behavior didn't change.

AlexeiQ:
Congratulations to your first PR!
The implementation looks good, thus I've only minor suggestions. The testing is, like you mentioned, extensive and we should simplify it a bit. It's a good idea to move the validity-check into the behavior and split it up, so it's easier and possible at all to change the validity of names 👍 

Here are some specific remarks:
1. `IValidNamedConcept.isLegalName()`: A more consistent name with the interface would be `isValidName()`
2. As far as I can see the inner `[` and `]` in both parts of the regex are totally obsolete. They are not characters, that can be used in the name but rather an regex-group inside a regex-group. I don't see any benefit, since they also don't improve the readability/understandability of the regex and say we can remove them.
3. Regarding the tests, I think we can simplify/ squash them a bit
    1. We can squash lots of the node-checks, e.g. the last 4 into one valid, since we don't gain anything by having multiple node-checks. Furthermore it seems to me, that the node-checks and the test-methods below are redundant and we can remove one of those parts (or pieces of both)
    2. I also think we don't need to test everything in all combinations, but can simply test each option independently with positive/negative tests and if necessary for both regex-parts. Though the testing-infrastructure in the utils-section looks impressive :) 
    3. The `TestValidNamedNodes` test should be located in the solution `test.ts.expr.os` and not in `test.in.expr.os`, since this solution is for interpreter/gen-tests
4. The new TestDSL must be included into the tests-buildscript, so it will be build with the tests and is available with the tests. 

Let's have a look at 3. and 4. together.

AlexeiQ (changes requested):
Congratulations to your first PR!
The implementation looks good, thus I've only minor suggestions. The testing is, like you mentioned, extensive and we should simplify it a bit. It's a good idea to move the validity-check into the behavior and split it up, so it's easier and possible at all to change the validity of names 👍 

Here are some specific remarks:
1. `IValidNamedConcept.isLegalName()`: A more consistent name with the interface would be `isValidName()`
2. As far as I can see the inner `[` and `]` in both parts of the regex are totally obsolete. They are not characters, that can be used in the name but rather an regex-group inside a regex-group. I don't see any benefit, since they also don't improve the readability/understandability of the regex and say we can remove them.
3. Regarding the tests, I think we can simplify/ squash them a bit
    1. We can squash lots of the node-checks, e.g. the last 4 into one valid, since we don't gain anything by having multiple node-checks. Furthermore it seems to me, that the node-checks and the test-methods below are redundant and we can remove one of those parts (or pieces of both)
    2. I also think we don't need to test everything in all combinations, but can simply test each option independently with positive/negative tests and if necessary for both regex-parts. Though the testing-infrastructure in the utils-section looks impressive :) 
    3. The `TestValidNamedNodes` test should be located in the solution `test.ts.expr.os` and not in `test.in.expr.os`, since this solution is for interpreter/gen-tests
4. The new TestDSL must be included into the tests-buildscript, so it will be build with the tests and is available with the tests. 

Let's have a look at 3. and 4. together.

AlexeiQ:
Points 1) + 2) look good
For 3.3): Commit d9ad6a0 changed the vcs.xml. Why did this change happen?

mgronover:
> Points 1) + 2) look good For 3.3): Commit [d9ad6a0](https://github.com/IETS3/iets3.opensource/commit/d9ad6a04c68e7b85910bd2de4aef5465934304b8) changed the vcs.xml. Why did this change happen?

Good question, i do not know. To me it's a kind of MPS magic, i never touched that file manually.

alexanderpann:
@mgronover 
```
 <component name="VcsDirectoryMappings">
    <mapping directory="$iets3.github.opensource.home$" vcs="Git" />
  </component>
```
Maybe you don't have the path variable set.
```
 <component name="GithubSharedProjectSettings">
    <option name="branchProtectionPatterns">
      <list>
        <option value="master" />
        <option value="maintenance\/.*" />
      </list>
    </option>
  </component>
```
Do you possibly have your GitHub Account added in preferences->Version Control->GitHub?
This could be the reasons for the changes in vcs.xml.

mgronover:
@alexanderpann 

1. Yes `iets3.github.opensource.home` is set.
2. You  mean i should simply remove the "Michael Gronover" entry (you wrote it "could" be the reason):
image

alexanderpann:
@mgronover 2. The file comes from the IntelliJ platform and it is not documented when it is created, thats why I wrote "could". You don't have to remove the entry. If you keep it, just make sure not to commit it.  You rarely have to commit *.xml files, except maybe [build-allScripts.xml](https://github.com/IETS3/iets3.opensource/blob/master/build/scripts/build-allScripts.xml) in rare cases.

AlexeiQ (dismissed):
The latest change looks good and all my feedback is addressed 👍 . Approve!
But someone has to check my commits as well.

arimer (commented):
@mgronover , @AlexeiQ In general I like the idea making that constraint customizable. 

However I have still some minor remarks:

- At the moment it is not possible to start a writing a name with a `paragraph` or an `apostroph`. Although a custom concept would allow both cases, the default regex would look like this: 

```
[a-zA-Z$_äöüßÄÖÜ][a-zA-Z0-9$_äöüßÄÖÜ§']*
```
Is this expected behaviour or are we just missing those cases?

- I also think that this [comment](http://127.0.0.1:63320/node?ref=r%3A02d078a1-d0db-43fc-a66a-8505d53851a4%28org.iets3.core.base.constraints%29%2F3632959622637106050) is not helpful. IMHO the code is soo simple, we do not need to comment it.

mgronover:
1. Starting a name with an apostrophe is not really common (very rarely used), so it can shouldn't be a default case. The same reason is applicable for paragraph I think. So we can assume that this should be the expected behaviour. If someone really do need another one, it is required to override these methods.

2. Yes, it can be removed.

3. :+1:

arimer:
@mgronover regarding 1: I would actually argue that starting somehting with a paragraph is the regular usecase, at least for german law.

```
Example: § 26 Vorstand und Vertretung
```
But I also agree, as long as there is no need for supporting this, we can postpone it.

arimer:
- rebased on current [maintenance/mps20211](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20211)
- fixed findings
- updated changelog

arimer (approved):
Nice contribution 👍🏼

arimer:
I do not know exactly why we missed the existence of the [extension point: identifierConfigurator](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F4394073590212312598) and went with the approach of having general gehaviour methods. However it feels like we should choose one approach over the other and do not mix them.

PR #632, opened 2022-12-20
Core expr: make NumberType.setRange virtual

alexanderpann:
This change is needed so that the method can be overridden.


lhartl (approved):
Thank you! 
During the review, I noticed some more necessary adaptions to allow us to properly override the NumberType. I'll create  a follow up ticket. 
This PR solves our current problem, so please already merge this.

PR #633, opened 2022-12-22
[minor] Typesystem: handle cases where computeSupertype returns an error

alexanderpann:
Error message: `Can't cast node: 4621384398004576888, concept: jetbrains.mps.lang.typesystem.structure.RuntimeErrorType to concept: org.iets3.core.expr.base.structure.Type`

The original problem is that we calculate the super type in some rules such as the ImmutableListLiteral, IfExpression etc. and then we cast the returned node (`:Type`). The implementation of the super type method for units can return nodes of type `RuntimeErrorType` that have to be handeled separately. I've checked all instances where this method or a similar method is used and opted for throwing a type system error instead of  using a safe cast (as)  and just swallowing the error message. To me, it looks safer to not calculate a type in those cases instead of returning null and have types like List<null>.
In some cases, `RuntimeErrorType` is assigned directly as the type of the node. In those cases, a separate [checking rule](http://127.0.0.1:63320/node?ref=r%3A42a3bc53-29b1-44d6-9767-7c921cef7ba0%28org.iets3.core.expr.typetags.typesystem%29%2F7718893141589766030) is responsible for showing the error.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/typesystem.mps
@@ -182,6 +182,9 @@
         <child id="1144231399730" name="condition" index="1Dwp0S" />
         <child id="1144231408325" name="iteration" index="1Dwrff" />
       </concept>
+      <concept id="5497648299878491908" name="jetbrains.mps.baseLanguage.structure.BaseVariableReference" flags="nn" index="1M0zk4">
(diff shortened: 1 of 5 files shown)


alexanderpann:
Yes, this was done on purpose. I remember that the error was better visible that way.

wsafonov:
> Yes, this was done on purpose. I remember that the error was better visible that way.

Ok, then it's fine for me if it was an actual improvement.

wsafonov (approved):

PR #634, opened 2023-01-10
Make expression and top level content substitution menu customizable

alexanderpann:
Both expressions and top level content have an implicit substitution menu (include menu for the subconcepts). The Substitution menu unfortunately only supports contributions, that means no concepts can be removed (this was possible in early MPS versions).
Not all concepts of KernelF are needed all the time (e.g. in test suites), so it must be possible to filter them from the substitution menu. Own concepts can be filtered through can be child/parent/ancestor/root constraints. 
I've recreated the menu for the subconcepts with optional filtering of the concepts through a new extension point `substitutionCustomization`:
Screenshot 2023-01-09 at 11 51 38

An alternative solution would to extract the default menus into a separate plugin which doesn't have to be installed. Then the  menus could be populated through default substitute menu contributions. I assume that filtering is still easier as many  concepts are indeed need and only a handful concepts have to be filtered.

Are there alternative built-in approaches that I am missing?

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
@@ -14,6 +14,7 @@
     <use id="120e1c9d-4e27-4478-b2af-b2c3bd3850b0" name="com.mbeddr.mpsutil.editor.querylist" version="0" />
     <use id="b33d119e-196d-4497-977c-5c167b21fe33" name="com.mbeddr.mpsutil.framecell" version="0" />
     <use id="b1ab8c10-c118-4755-bf2a-cebab35cf533" name="jetbrains.mps.lang.editor.tooltips" version="0" />
+    <use id="c0080a47-7e37-4558-bee9-9ae18e690549" name="jetbrains.mps.lang.extension" version="2" />
     <devkit ref="fbc25dd2-5da4-483a-8b19-70928e1b62d7(jetbrains.mps.devkit.general-purpose)" />
   </languages>
   <imports>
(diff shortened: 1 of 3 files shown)


HeikoBecker:
I have merged the branch locally with `maintenance/mps20211` and there were no build errors.

The code itself looks okay to me, however I have a conceptual concern:

Adding the extension point without a default implementation and using all instances for filtering (see screenshot) may lead to confusion in customer code. I would favor if we were to use an approach like the NixSupport, where one instance is responsible for the filtering.
image

arimer (commented):
I am not fully familiar with all details of extension points in MPS, however I still have some remarks:

@alexanderpann 
- How does the extension point behaves if there is no default implementation of the provided interface [SubstitutionCustomizer](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F4852640530887511198)? 
- According ot the [documentation](https://www.jetbrains.com/help/mps/2021.1/extension-support.html) , an `Extension` using the provided `ExtensionPoint` can be shipped along side with the extension. Can't we impl. an extensions containing the default behaviour (as before your change)?
- How can users contribute a custom impl. of `SubstitutionCustomizer`? Where does the concrete implementatoin of that interface needs to be located, so that MPS picks it up?

HeikoBecker:
> * How does the extension point behaves if there is no default implementation of the provided interface [SubstitutionCustomizer](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F4852640530887511198)?
> * According ot the [documentation](https://www.jetbrains.com/help/mps/2021.1/extension-support.html) , an `Extension` using the provided `ExtensionPoint` can be shipped along side with the extension. Can't we impl. an extensions containing the default behaviour (as before your change)?
> * How can users contribute a custom impl. of `SubstitutionCustomizer`? Where does the concrete implementatoin of that interface needs to be located, so that MPS picks it up?

I discussed this with @mgronover and @AlexeiQ yesterday and we shared your concerns. As far as I understood, we agreed that it would be better to have the code more in line with how the NixSupport is implemented because this will mandate a) giving a default implementation (that does not filter at all) and b) having a clear defined method for overriding this filter. 
As it stands, the current setting would only allow "subtractive" filtering, if a filter is defined that removes all instances of `LogicalAndOperator`, this could not be undone.

AlexeiQ:
This idea doesn't work out to filter out concepts that are not owned from rootnodes that are not owned, ie. introduce a blacklist for not own conwepts inside of not owned (root)concepts.
Flaws:
- we would have to introduce a substitute-menu everywhere such a filter could be used, potentially for every concept
- blacklisted concepts are not shown in the auto-completion, but can reside inside the model without any shown error, if they were introduced before the filter or "hacked in" eg. via console.

The state-of-the-art handling is to extend the rootnode you need (if you don't own it) and introduce constraints there for a blacklist. This way all instances of blacklisted concepts are shown as errors and not suggested anymore.

PR #637, opened 2023-01-19
Add a type to the field of FieldSetter

alexanderpann:
With this change, the type of the referenced field can be displayed in the builder with SHIFT+CTRL+P.


arimer:
@alexanderpann small side quetsion: besides the convinience of having the type beeing calculated and visivle via  SHIFT+CTRL+P, is there any other benefit of having this change?

arimer (approved):
Regardless the remark in the comments, the change is trivial and can be merged from my point of view.

alexanderpann:
No, only convenience. In general, I think it is a good idea to have the option to check the type because type declaration (e.g. in constants -> var test) are most of the time optionally in KernelF.

PR #641, opened 2023-01-26
Extend NixSupport with a constructor supporting list values

HeikoBecker:
This PR adds a new constructor to the NixSupport class that supports initialising it with a sequence of values.

Currently there is no check applied that the list might not be empty.


alexanderpann (approved):
LGTM

PR #644, opened 2023-02-13
Make the function runSolver virtual to overwrite it where necessary

HeikoBecker:
This PR makes the function `runSolver` in the Behavior of the `ISolvable` concept virtual such that it can be overwritten in dependencies.

**Known Issues**

If the function is implicitly overwritten in sub-concepts, the reflective editor must be used to add the overriding information manually. This is only required if the function should properly overwrite the `runSolver`  method from the `ISolvable`. 

_No changes are required if the two functions can be kept separate._


arimer (approved):
`runSolver()` is virtual now 👍🏼

PR #647, opened 2023-02-14
Set different priorities for the conditional editors

alexanderpann:
I'm not sure if there should be a specific order. I've used priorities from the 200er range because the 100er range is used by mbeddr. The original error message comes from mps itself:
`
 ERROR - actEditorHintsSpecificRegistry - Additional editor class org.iets3.core.expr.tracing.editor.BaseConcept_conditionalEditor_Editor_Tracing is applicable to the current context (de.slisson.mps.conditionalEditor.hints.editor.conditionalEditorHints.conditionalEditor_doNotUseThisHint, de.slisson.mps.conditionalEditor.hints.editor.conditionalEditorHints.conditionalEditor). Skipping this editor, using class org.iets3.core.expr.tracing.editor.BaseConcept_conditionalEditor_Editor_Coloring.`

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/editor.mps
@@ -3525,7 +3525,7 @@
     </node>
   </node>
   <node concept="RtYIR" id="1WlYLwX1DOp">
-    <property role="Rtri_" value="100" />
+    <property role="Rtri_" value="200" />
     <property role="3NULOk" value="ModelCoverage_Coloring" />
     <property role="3GE5qa" value="coverage" />
     <ref role="1XX52x" to="tpck:gw2VY9q" resolve="BaseConcept" />
@@ -5733,7 +5733,7 @@
     </node>
   </node>
   <node concept="RtYIR" id="3TIaSh$92nT">
-    <property role="Rtri_" value="100" />
+    <property role="Rtri_" value="210" />
     <property role="3GE5qa" value="coverageHighlighter" />
     <property role="3NULOk" value="coverageAnnotation" />
     <ref role="1XX52x" to="tpck:gw2VY9q" resolve="BaseConcept" />
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tracing/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tracing/models/editor.mps
@@ -333,7 +333,7 @@
     </language>
   </registry>
   <node concept="RtYIR" id="1rUbSenML5">
-    <property role="Rtri_" value="100" />
+    <property role="Rtri_" value="240" />
     <property role="3NULOk" value="Tracing" />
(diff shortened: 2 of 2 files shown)


arimer (approved):
Changing the prios. to get rid of the error message seens fine to me 👍🏼

wsafonov (changes requested):
So all the conditional editors would have now a distinct priority which puts them into that specific order of applying/rendering, right? I would only prefer to keep the step between the single priorities higher (200, 300, etc...) to differentiate between them more easily.

wsafonov:
And does it then really fix the error? Because to me the editors would be still applicable in the same context, but just with a different priority/order.

alexanderpann:
I'll probably have to test that, but the different priorities should mean that only the editor with the highest priority is selected. In the past, the editors had the same priority, so MPS selected the first editor that it found and showed the error for the other editors and skipped those editors. I am not sure what the right step size and order is. I think we should really think about this because we have conditional editors for base concept. If we set the values too high people will fail to override the editors. So I also think that the priorities should be documented somewhere. I've written this before: I've copied the pattern from the mbeddr conditional editors.

wsafonov:
> I'll probably have to test that, but the different priorities should mean that only the editor with the highest priority is selected. In the past, the editors had the same priority, so MPS selected the first editor that it found and showed the error for the other editors and skipped those editors.

That would mean that globally only one conditional editor could be applied to a specific node, right? That doesn't make any sense. To my understanding the priority only decides upon the application/rendering order of the conditional editors. So if there are several applicable conditional editors (based on their condition), with the same priority you would just get some randon order of those editors applied to the same node. With different priorities you could control which editor comes first, basically being then the innermost editor in the nested structure of conditional editors.

> I am not sure what the right step size and order is.

There is not "right" step size - it just seems easier to memorize large round numbers then 230, 240... But if we used this pattern in mbeddr already, fine to leave it like this. 

Regarding the order - I'm not sure as well. I would say, it could be really sometimes a matter of taste, which conditional editor should be applied first. It depends very much on a specific combination of those conditional editors. But on the other side, I don't think that many of those conditional editors would be available simultaneously in the same context based on their conditions. Maybe only tracing editors. But then it's only about finding the right order between the tracing editors. For other editors it should be ok to still use the default priority.

Getting back to the error. I actually don't quite understand what's happening there, because again as I said, in my understanding applying several conditional editors in the same context should work fine. And the priority is not affecting that. But the error is rather related to the editor hints. So maybe it's caused somehow by the hints usage. Would be curious to know if the error would disappear with this change.

alexanderpann:
@wsafonov I think I've figured out what's going on: the editor with the highest priority is in fact selected, but as you already pointed out, the other editors are also applied because all the conditional editors use next-editor to also apply the other editors. The conditional editor language overrides the comparator that checks which editor should be used and also disables the logging for the class that logs the mentioned error. We already found the right editor because of the priority, so we don't care that there are other editors with the same size of context hints available. I think that the error can only happen on startup when the application plugin of the conditional language is not applied in time and the normal comparator and error logging level are used. I'm not sure if we can even fix that.

alexanderpann:
Closing this PR because it won't fix the original exception.

PR #654, opened 2023-03-15
Expression eval helper methods without coverage

alexanderpann:
This PR adds complementary helper methods to IETS3ExprEvalHelper, so you can call the methods without the default coverage analyzer. In some cases, e.g. when invoking the other methods too often or maybe from the generator, the default (global) coverage analyzer needs too much memory and the coverage information is not even needed.


arimer (approved):
No semantic or breaking changes introduced. Just restructuring private API and additions to the public one. Looks good to me 👍🏼 

One side notice: In the long termn we might want to use **error log statements** instead of [system.err.printlns](http://127.0.0.1:63320/node?ref=r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%29%2F1887965683634660627). But this is out of scope for this RP.

PR #656, opened 2023-03-27
[minor] stringvalidator: fix cannot be/must be for predecessor/successor

alexanderpann:
```
test case predecessorSuccessor [success] {                                                  
  assert validate "12345"each str[3]predecessor cannot be str[1]|.ok equals true [C] [71 ms]
  assert validate "12345"each str[3]successor cannot be str[5]|.ok equals true [C] [0 ms]   
  assert validate "12345"each str[3]predecessor must be str[2]|.ok equals true [C] [1 ms]   
  assert validate "12345"each str[3]successor must be str[4]|.ok equals true [C] [0 ms]     
}                                                                                           
```

The first two asserts failed in the past.


arimer:
The changes iteself look fine. But could you maybe explain the initial bug / problem? 

Without knowing the DSL it is quite hard to comprehend why this test was broken:
> assert validate "12345"each str[3]predecessor cannot be str[1]|.ok equals true [C] [71 ms]

Or how your change influences that behaviour.

alexanderpann:
I've got this test case from the customer that requested the language itself. The bug itself was missed because there was no good test for it. The implementation for "cannot be" in combination with predecessor/successor was just missing.

PR #657, opened 2023-03-30
Unify end cell of multitab, builder expression and validate string

alexanderpann:
This was a customer request. All three concepts used different custom cells to denote the end of the editor. I've attached some screenshots that showcase how the newly introduced end cell looks like. It is modeled after the end cell of table rows. If you are unhappy with the look (e.g. the thickness, color), please amend the commit yourself. There is also no way to show the cursor in a good way, the same way it is done for the table rows.
Screenshot 2023-03-30 at 08 31 09

Screenshot 2023-03-30 at 08 38 44

Screenshot 2023-03-30 at 08 37 59

Screenshot 2023-03-30 at 08 45 15

The vertical line gets the highlight color when it is selected:

Screenshot 2023-03-30 at 08 51 30

Screenshot 2023-03-30 at 08 52 28


mgronover (approved):
Future enhancements required:

If the caret is positioned next to the right side of the row of a MultiDecTab,
the behavior for DEL is suspicious: the row to the left is deleted, which is usually not the intended behavior.
We should think about having a two step deletion here.

According to the DarkTheme: I do not see the Open-/CloseBracketCell.

arimer:
In general, having a common way to define the "end" of an editor cell, is good idea 👍🏼 but I think we need some more polishing.


**Some remarks:**
- Just out of curiosity: Wasn't there a way to make use of a constant cell (instead of impl. a cell in baselanguage) where you could apply the styles and actions and extract that cell into an editor component to make it re-usable?
- I think we are missing some side transformations an that End-Cell. Image you have a `ValidateStringExpression` like this: 
grafik
At the moment there is no easy way to use any additional expression (DotExpressions, +, - ...) when you place the cursor on the `EndCell`. Meaning it is not possible to construct further complex expressions starting from `ValidateStringExpression`. I am not sure how the behaviour was before that, but to me it seems weird
-

Issue #659, opened 2023-04-03
Type checking equality operator on set literals fails


HeikoBecker:
![image](https://user-images.githubusercontent.com/3237993/229427479-31949075-316b-47b7-a3e0-ea6675c3ea20.png)
In the screenshot above, `testSet2` has a type error: `Error: incompatible types: set<string> and set<string>`.

Expected behavior: No type-checking error is reported

HeikoBecker:
Fixed in #660

PR #660, opened 2023-04-03
Make collectiontype implement IComplexTypeSupportsEquals (#659)

HeikoBecker:
This fixes a bug in the typesystem where equalitites of collections were not typechecked correctly.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.collections/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.collections/models/behavior.mps
@@ -21,6 +21,7 @@
     <import index="33ny" ref="6354ebe7-c22a-4a0f-ac54-50b52ab9b065/java:java.util(JDK/)" />
     <import index="5qo5" ref="r:6d93ddb1-b0b0-4eee-8079-51303666672a(org.iets3.core.expr.simpleTypes.structure)" />
     <import index="xfg9" ref="r:ac28053f-2041-47f6-806b-ecfaca05a64a(org.iets3.core.expr.base.runtime.runtime)" />
+    <import index="guwi" ref="6354ebe7-c22a-4a0f-ac54-50b52ab9b065/java:java.io(JDK/)" implicit="true" />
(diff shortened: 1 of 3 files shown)


HeikoBecker:
The last commits have extended the new functionality for collections. There was some breakage with the test, as the new support for equality typing on collections made some cases that would previously fail in the semantics fail in the type checking. I have updated the tests accordingly.

arimer (changes requested):
The impl. of _isComparableTo()_ looks fine. However, I think your [comment](http://127.0.0.1:63320/node?ref=r%3A2c0153cb-f6d9-49f3-b0fe-e4f726698ef0%28org.iets3.core.expr.collections.behavior%29%2F4454357034015195699) is missleading:

> // We want to allow e.g. sets to be comparable but not lists with sets
 
According to the impl. you could still compare two lists of basetype set with each other and would have a valid result. For example:  `list<set<boolean>>.isComparableTo(list<set<number>>)` would result in true, which makes sense.

I also think it would not harm to have setup some tests checking the new functionality before mergin this PR.

HeikoBecker:
> The impl. of _isComparableTo()_ looks fine. However, I think your [comment](http://127.0.0.1:63320/node?ref=r%3A2c0153cb-f6d9-49f3-b0fe-e4f726698ef0%28org.iets3.core.expr.collections.behavior%29%2F4454357034015195699) is missleading:
> 
> > // We want to allow e.g. sets to be comparable but not lists with sets
> 
I have changed the comment to better reflect the actual implementation
 
> I also think it would not harm to have setup some tests checking the new functionality before merging this PR.

Indeed, the latest commit added some more tests.

arimer (dismissed):
Thanks for the changes 👍🏼 

The test case is also nice.

Just a minor remark here: In case you want to make sure that a certain case is error free (like [here](http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F1986909083020056108)) you could explicitly ask MPS to do this via [CheckNodeForErrorMessagesOperation](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c89590388%28jetbrains.mps.lang.test.structure%29%2F1215607067978). 
So you could write a test, where you expect a case to be error free like this: 

grafik

PR #661, opened 2023-04-14
Prevent tagged types from wrapping

HeikoBecker:
Give the same treatment to tagged  types as done in #658 

before:

image

after:

image


arimer:
I guess the target branch is wrong :)

Issue #505, opened 2021-07-26
The type inference underlying the alt-expressions doesn't work correctly


kbirken:
The type inference of the alt-expression doesn't work as expected. Two examples:

<img width="628" alt="image" src="https://user-images.githubusercontent.com/1193045/127029537-21740ca5-3c55-44c8-a139-1f40530afa41.png">

<img width="594" alt="image" src="https://user-images.githubusercontent.com/1193045/127029656-3fbcce08-e868-49ca-b18e-5fcd4d13e3fe.png">

Expected: The result should be a join-type (but I didn't find a specification for this in the KernelF-paper). 

Gimmick: It would be nice if the underlying infrastructure (`TypingHelper`) would support an option to return just a single type and issue errors if the list of input expressions contains different types.

kbirken:
The DecTab and some other KernelF constructs suffer from the same problems.

lhartl:
[JoinType](url)

lhartl:
@kbirken can you list the problematic parts of kernelF when you stumble over them please? I think these are nice tickets for people to pick up during ttt.

heikob2:
#### Observations
The Problem is located in the `SimpleTypesPrimitiveTypeMapper::computerSupertype()` in the _toplevel-else-block_:
- There the jetbrains _SubtypingManager_ `SubtypingManagerNew` calls `leastCommonSupertypes()`, which returns an empty set if **2 different types** are provided and the correct **join** is returned as type.
- If there are **3 different types** provided an arbitrary type is returned (cause the parameter is a hash-set). The result then is the **wrong type**, which is exposed.

#### Todos
- [ ] fix it
- [ ] ensure that the types in the join-type are distinct

mgronover:
I choose fixing it by using the following  work around:

The least common super type of a set of types is determined by the method:
`typechecker.getSubTypingManager().leastCommonSupertypes(setOfTypes, boolStrongSubtyping)`

If `setOfTypes` contains the three iets3 types : `{StringType, BooleanType, NumberType}` , the derivation of the least common super type is done wrong. It operates correctly if there are only two types (as mentioned [above](https://github.com/IETS3/iets3.opensource/issues/505#issuecomment-889278536)). So I broke the set down and called that function repeatedly until the least common type was determined.

Additionally I recursively flattened the join types and optional types before using this set of types to get the super type of them. If there was an option type (or a NoneType is one of the types within the set) this will result into an surrounded option type.

[...]

PR #662, opened 2023-04-20
Typesystem: fixed wrong determination of least common super type (#505)

mgronover:
Ticket: #505

The least common super type of a set of types is determined by the method:
typechecker.getSubTypingManager().leastCommonSupertypes(setOfTypes, boolStrongSubtyping)

If setOfTypes e. g. contains the three iets3 types : {StringType, BooleanType, NumberType} , the derivation of the least common super type is done wrong. It operates correctly if there are only two types (as mentioned https://github.com/IETS3/iets3.opensource/issues/505#issuecomment-889278536). So I broke the set down and called that function repeatedly until the least common type was determined.
By the way, the list of types you see above is only an example. Any types should be allowed! 

Another point is that NumberTypes are only reduced to a common Numbertype (the least required super type) if setOfTypes *only contains* Numbertypes. This unnecessary constraint is now removed.  

Additionally I recursively flattened the join types and optional types that are found within setOfTypes before determining the super type of their element types. So the resulting types becomes the super type of the element types of this special types. Afterwards this the resulting types are surround with the join and/or optional type to get only one super type.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/constraints.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/constraints.mps
@@ -818,6 +818,30 @@
       </node>
     </node>
   </node>
+  <node concept="1M2fIO" id="5ipapt35kk6">
+    <property role="3GE5qa" value="join" />
+    <ref role="1M2myG" to="hm2y:7VuYlCQZ3ll" resolve="JoinType" />
+    <node concept="9S07l" id="5ipapt35kk7" role="9Vyp8">
+      <node concept="3clFbS" id="5ipapt35kk8" role="2VODD2">
+        <node concept="3clFbF" id="5ipapt35krh" role="3cqZAp">
+          <node concept="2OqwBi" id="5ipapt35lty" role="3clFbG">
+            <node concept="2OqwBi" id="5ipapt35kAG" role="2Oq$k0">
+              <node concept="nLn13" id="5ipapt35krg" role="2Oq$k0" />
+              <node concept="2Xjw5R" id="5ipapt35kMz" role="2OqNvi">
+                <node concept="1xMEDy" id="5ipapt35kM_" role="1xVPHs">
+                  <node concept="chp4Y" id="5ipapt35kV_" role="ri$Ld">
+                    <ref role="cht4Q" to="hm2y:5ipapt35kjG" resolve="IJoinTypeContext" />
+                  </node>
+                </node>
+                <node concept="1xIGOp" id="5ipapt35ldI" role="1xVPHs" />
+              </node>
+            </node>
+            <node concept="3x8VRR" id="5ipapt35lKq" role="2OqNvi" />
+          </node>
+        </node>
+      </node>
+    </node>
+  </node>
   <node concept="1M2fIO" id="6BCTLIjCrHS">
     <ref role="1M2myG" to="hm2y:6BCTLIjCra2" resolve="IControlAdvancedFeatures_old" />
   </node>
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/constraints.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/constraints.mps
@@ -818,30 +818,6 @@
       </node>
     </node>
   </node>
-  <node concept="1M2fIO" id="5ipapt35kk6">
-    <property role="3GE5qa" value="join" />
-    <ref role="1M2myG" to="hm2y:7VuYlCQZ3ll" resolve="JoinType" />
-    <node concept="9S07l" id="5ipapt35kk7" role="9Vyp8">
-      <node concept="3clFbS" id="5ipapt35kk8" role="2VODD2">
-        <node concept="3clFbF" id="5ipapt35krh" role="3cqZAp">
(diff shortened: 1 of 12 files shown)


mgronover:
There is another further enhancement possible:

If the set of types from that the least common super type is to be determined contains more than one numerical type with different ranges and precisions, these types should have a least common numerical super type by themselves.
That numerical super type is an expansion of these attributes (greatest precision and smallest common range). 

Currently this is only done if all given types are numerical.

mgronover:
Also fixed the issue with the numerical types.
So here is resulting type for the variable `v`:
image

arimer:
FYI: I've triggered the rebuild, because the last status was not visible in github.

arimer:
FYI: I've triggered the rebuild, because the last status was not visible in github.

arimer (commented):
Some remarks/questions so far to the: SimpleTypesPrimitiveTypeMapper

**SimpleTypesPrimitiveTypeMapper**
- Can you maybe in simple terms explain what the expected behavoiur of `computerSupertype()` should be? When do we return a joinType and when do we calculate a common super type?
- maybe a stupid question, but does the `SimplePrimitiveTypeMapper.computerSupertype()` should only cover NumberType/StringType/Boolean or other primitive types as well? 
grafik
- I think having a test for that method would be beneficial. We could maybe use [TypingHelper.computeRegularSuperType()](http://127.0.0.1:63320/node?ref=r%3A80cf2246-750c-4158-9056-a619ebcf894c%28org.iets3.core.expr.base.typesystem%29%2F1206081519719692184) for testing purposes
- lets rename compute**r**Supertype() -> computeSupertype()
- make ComputeSupertypeOfNumberTypes() lower case
- if ComputeSupertypeOfNumberTypes() only operates on lists of number types, why not just declare the input parameter as `nlist<NumberType> types`? Then you don't need to filter [here](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F2124603155693245348)
- also instead of the `where`  [here](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F2124603155693486254) you could just do a `types.ofConcept<NumberType>` and calculate directy a `sequence<node<NumberType>>`
- I think we can safely rely on the [SetType](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959032e%28jetbrains.mps.baseLanguage.collections.structure%29%2F1226511727824) // [HashSet](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c8959032e%28jetbrains.mps.baseLanguage.collections.structure%29%2F1226516258405) provided by the `jetbrains.mps.baseLanguage.collections` instead of using the pure java `Set // HashSet`
- can you explain to me, why we ignore the first entry [here](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F643800908087552243)? Also, why do we [clear](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F643800908087904647) the list after adding 2x entries?
- I think the impl. is still not complete. In case your list contains multiple, duplicated types like this: `new arraylist<node<>>{<StringType()>, <StringType()>, <BooleanType()>, <NumberType()>}` your calculated joinType will also contain those "duplicated" string types: `join<string, string, boolean, number[-∞|∞]{0}>`. Should the joinType really contain 2x stringtype?

Review of TypingHelper is still ongoing :)

mgronover:
Thanks for the review!

I tried to find good answers to all of your intentions. Here is what I did:

- I adjusted the explanation in this MR it hopefully is now clear what the expected behavoiur of `computerSupertype()` is (BTW: the method was renamed to `computeSupertype()`).
- I added some tests also to document the behaviour.
- The input parameter of `ComputeSupertypeOfNumberTypes(..)` is now `nlist<NumberType>` types.
- Instead of using the pure java `Set` the `SetType` as an `hashset<node<>>` is used.
- The first entry is not really ignored, but already added to the set before the loop is entered (within the loop the first entry will be the least common supertype so far).
- Did remove the commented out code.
- Test are implemented on the basis of node checks using the `has type` operation (see: http://127.0.0.1:63320/node?ref=r%3A84f8cd40-0164-44eb-acf9-1b79e378eb7a%28test.ts.expr.os.LeastCommonSuperType%40tests%29%2F8720697010560585743)
- Renamed the `flattenJoinTypes()` Method to `checkOptionTypeIsReuiredAndRecursivlyExtractElementTypes()`

TODO:
- change target branch of this MR

arimer:
Side notice: master is alredy migrated to [ MPS 2022.2](https://github.com/IETS3/iets3.opensource/blob/4bc153a16915e41a02a98842083a06a39d414fc0/build.gradle#L59) but your branch is still building with `MPS 2021.3.2`. I would higly recommend to align with master before merging.

mgronover:
According to 
- https://github.com/IETS3/iets3.opensource/pull/662#pullrequestreview-1425946649 and also 
- https://github.com/IETS3/iets3.opensource/pull/662#issuecomment-1598585497 
I created a new merge request based on the branch "maintenance/mps20211".

All review discussion will go on there.

PR #665, opened 2023-04-24
Add a type inference rule for IfElse Section (#664)

### LINKED ISSUE #664: IfElseSection has no type system type
[IfElseSection](http://127.0.0.1:63320/node?ref=r%3A66e07cb4-a4b0-4bf3-a36d-5e9ed1ff1bd3%28org.iets3.core.expr.base.structure%29%2F606861080870797309) does not contain any type system type and therefor access of the `IfElseSection.type` results into `null`.

HeikoBecker:
Also include some tests to ensure that the typesystem will remain consistent.

Closes #664


arimer (approved):
Thanks for adding the missing type. 

One remark on your tests: You could even siplyfiy those by using the right test annotation: [NodeExpectedTypeCheckOperation](http://127.0.0.1:63320/node?ref=r%3A00000000-0000-4000-0000-011c89590388%28jetbrains.mps.lang.test.structure%29%2F4395293866213195846) got you covered :)
grafik

PR #667, opened 2023-05-02
Minor fix for ComponentInstances.

HeikoBecker:
Add ISubstructureContent interface to AbstractComponentInterface. This change fixes a warning in dependent projects.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/structure.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.components.core/models/structure.mps
@@ -383,9 +383,6 @@
         <property role="YLQ7P" value="The link was moved to concept &quot;org.iets3.components.core.structure.AbstractComponentInstanceWithRef&quot;" />
       </node>
     </node>
-    <node concept="PrWs8" id="siw10FiR6l" role="PzmwI">
-      <ref role="PrY4T" node="siw10FiR6c" resolve="ISubstructureContent" />
-    </node>
     <node concept="PrWs8" id="6Z4vEhQZ6sI" role="PzmwI">
       <ref role="PrY4T" to="vs0r:3m8H$lmFM60" resolve="IDocumentable" />
     </node>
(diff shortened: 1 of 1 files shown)


arimer (commented):
I have still some open questions to the fix:

- I am not fully aware of the semantic meaning of the `ISubstructureContent`. Does it also mean, that every concept implementing this interfaces can be used inside the substructure (of a component)? If this is the case, your change would basically now allow the usage of instances of those concepts inside the substructure:
grafik

- Now that you have added this interface to the abstract concept, you can remove it from [here](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F509610514780483989) and [here](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F8209493818901074928)
- What was the excact warning/error message you were facing?

HeikoBecker:
> 1. I am not fully aware of the semantic meaning of the `ISubstructureContent`. Does it also mean, that every concept implementing this interfaces can be used inside the substructure (of a component)? If this is the case, your change would basically now allow the usage of instances of those concepts inside the substructure:

That might very well be, I must admit that I was just trying to fix the error in question (see below, 3.) and noticed that all sub-concepts anyway implement the interface.

> 2. Now that you have added this interface to the abstract concept, you can remove it from [here](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F509610514780483989) and [here](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F8209493818901074928)

Done in latest commit.

> 3. What was the excact warning/error message you were facing?

Some internal (java) code raises this error:

`Error: type node<AbstractComponentInstance> is not a subtype of node<ISubstructureContent>` .

arimer (approved):
Thanks for cleaning up 👍🏼. Also the explanation of the warning/error is clear now.

Issue #668, opened 2023-05-03
Typesystem improvement


mgronover:
Allow an override of the type of a FieldSetter.

arimer:
closed via #669

PR #669, opened 2023-05-03
fixed type inference of FieldSetter to allow override

mgronover:
Ticket: #668 

Customer may want to override the type of a FieldSetter
This can be achieved with this PR.


mgronover:
This commit is cherry picked into the branch "datev-loon-staging-20213" (right now).

arimer (commented):
@mgronover besides allowing the type override of a FieldSetter you also changed the implementation of [typeof_IFunctionCall](
http://127.0.0.1:63320/node?ref=r%3A3b5d2a4d-f539-4854-bc25-c43da4b5202c%28org.iets3.core.expr.lambda.typesystem%29%2F4790956042240522435) was this done on purpose or by accident? 
I would also argue that we should not use a "[node as Type cast](http://127.0.0.1:63320/node?ref=r%3A3b5d2a4d-f539-4854-bc25-c43da4b5202c%28org.iets3.core.expr.lambda.typesystem%29%2F2619776616563037887)" but a `node:Type` cast.

mgronover:
@arimer both points were intended by customer [ticket](https://gitlab-extern.itemis.de/customer_datev/lodas/-/issues/1018). According to the type cast: well it is safe to write node:Type in this case, are there concrete reasons to do so?

alexanderpann:
> I would also argue that we should not use a "[node as Type cast](http://127.0.0.1:63320/node?ref=r%3A3b5d2a4d-f539-4854-bc25-c43da4b5202c%28org.iets3.core.expr.lambda.typesystem%29%2F2619776616563037887)" but a `node:Type` cast.

The `node:Type` is a source of many of our type system exceptions. IMO, we should get rid of it for the reasons mentioned in  #688.

arimer:
IMHO letting the TS fail with with a `NodeCastException` is more helpful, because it raises awareness that the overriden type is not compatible. It is more helpful than just "failing silently" and do not calculate any type.

alexanderpann:
@arimer We probably have to check all those instances and check that we don't fail silently. Maybe some null checks or additional type system checks are necessary. Throwing **fatal errors** for cases where the user provides the wrong type, the editor is incomplete, an abstract concept is used, or an `RuntimeErrorType` is returned because of related errors is **never** an acceptable form of showing errors. All those cases have to be caught by the type system and shown in the editor by the type system, for example, type X is not a subtype of Y or the checks and calculations shouldn't be executed when the error happens earlier.

I really want to avoid **any** exceptions that are shown in the lower right corner. No user should read exception traces.

arimer:
@mgronover in case we merge this, what do we need todo on the stading branch?

arimer (approved):
@mgronover lets 🚀 with the current impl.

PR #671, opened 2023-05-15
Add support for type equality to maps

HeikoBecker:
Also refactor type equality for collections by reusing the isSameAs function. Add tests for map comparisons.


arimer (approved):
Changes look good, just some minor remark:

- is there a specific reason for using [Object.equals()](http://127.0.0.1:63320/node?ref=r%3A2c0153cb-f6d9-49f3-b0fe-e4f726698ef0%28org.iets3.core.expr.collections.behavior%29%2F7258829162985842656)here? Why not using `concept.isExactly()` which is a little bit more self explaining?  
- I like the simplificaiton of `CollectionType.isComparableTo()` 👍🏼

HeikoBecker:
> * is there a specific reason for using [Object.equals()](http://127.0.0.1:63320/node?ref=r%3A2c0153cb-f6d9-49f3-b0fe-e4f726698ef0%28org.iets3.core.expr.collections.behavior%29%2F7258829162985842656)here? Why not using `concept.isExactly()` which is a little bit more self explaining?

Thanks for noticing! Fixed it in the latest commmit.

Issue #672, opened 2023-05-16
NumberType.isSameAs should use NumberType#precision method


alexanderpann:
IMO, if one NumberType has the precision set to 0 and the other Numbertype doesn't have a precision defined and has therefore the default precision 0, they should be considered the same. At least that's how I interpret this method, we don't structurally compare those types.

HeikoBecker:
Closing this issue since it has been resolved with #673

PR #673, opened 2023-05-22
Use precision function in isSameAs on NumberType (#672)

HeikoBecker:
As pointed out in #672 , the number type equality function (`isSameAs`) was a bit too pessimistic in using the direct precision accessor for comparisons. With this PR, the comparison function uses the the `getPrecision` function instead which allows the types `number` and `number{0}` to be treated equally.


arimer (approved):
Thanks for the 🐛-fix and the additional tests 👍🏼

PR #676, opened 2023-06-16
Add a EndCell to tagged expressions (#675)

### LINKED ISSUE #675: Editing a unit expression captures the editor
Start editing a new unit expression, e.g.:
image

Any additional character typed after the `s` symbol will continue the editing of the unit expression. Pressing the enter key cannot move to a new line but adds a unit level multiplication:

image

When manually adding textual delimiters to `TaggedExpresison` and `UnitReference`, I can see focus the end of the expression by clicking in the editor and then enter a new line.

Maybe this can be fixed using a guiding border similar to `de.itemis.mps.math.notations` in mps-extensions?

HeikoBecker:
This uses the end cell from #657 to make it clear where the end of a tagged expression is to fix #675 .


arimer:
Tbh. I don' see much of a difference, neither in the eiditor nor in the behaviour of the editor. Do you maybe have an example (screenshot) what the change actually is doing?

arimer (approved):
Your chages improve the usability of the editor 👍🏼 

Just a remark:
In case you want to add yet another unit with `*` in this cursor position:
grafik
it is still really cumbersome and unintuitive to do that. The only way atm. is to hit ENTER :/

HeikoBecker:
> 

I checked this issue and unfortunately this is not as straight-forward to fix as the new line...

Will merge this PR and then see what I can do about the other issue.

PR #677, opened 2023-06-19
Repair github publication step

HeikoBecker:
The old plugin `co.riiid.gradle` was outdated. This PR replaces it with a [newer plugin](https://github.com/BreadMoirai/github-release-gradle-plugin) which is the same as the one used for MPS-extensions .

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ buildscript {
 plugins {
     id 'base'
     id 'maven-publish'
-    id 'co.riiid.gradle' version '0.4.2'
+    id "com.github.breadmoirai.github-release" version "2.4.1"
 }
 
 ext.dependencyRepositories = [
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -382,21 +382,18 @@ Includes dependencies:
...
     prerelease = true
-    assets = packageDistroWithDependencies.outputs.files.collect {it.path}
-}
-githubRelease.doFirst {
-    // do late body init to avoid executing listMergedPRs during config phase
-    github {
-        body = releaseNotes
-    }
+    releaseAssets = packageDistroWithDependencies.outputs.files.collect {it.path}
+    dryRun = true
```
arimer, line 394: Do you maybe know what the dryrun ins actually doing?

HeikoBecker: It is meant to be used locally for testing. I have set it to false in the latest commit as I forgot to change it in the original commit.

arimer (approved):
The change to the new plugin looks fine 👍🏼 

However some remarks/questions from my side:

- could you please update the camption of the PR to reflact the changes? I think you mostly "fixed the github release publication"
- linking the now used plugin: https://github.com/BreadMoirai/github-release-gradle-plugin in the description would also be benefictial.
- Did you check if the release artiact looks identical to the one produced by the old plugin?

None of those remarks is blocking, this is way I approve in any case :)

HeikoBecker:
> The change to the new plugin looks fine 👍🏼
> 
> However some remarks/questions from my side:
> 
> * could you please update the camption of the PR to reflact the changes? I think you mostly "fixed the github release publication"
> * linking the now used plugin: https://github.com/BreadMoirai/github-release-gradle-plugin in the description would also be benefictial.
> * Did you check if the release artiact looks identical to the one produced by the old plugin?
> 
> None of those remarks is blocking, this is way I approve in any case :)

I have adapted the descriptions accordingly.
Since I only did the dry-run, I only got shown whether or not the build would have worked. However, the plugin is currently also used for MPS-extensions without hickups as far as I could see, so this should be fine to replace.

PR #678, opened 2023-06-26
505 type system improvement

mgronover:
Ticket: https://github.com/IETS3/iets3.opensource/issues/505

The least common super type of a set of types is determined by the method:
typechecker.getSubTypingManager().leastCommonSupertypes(setOfTypes, boolStrongSubtyping)

If setOfTypes e. g. contains the three iets3 types : {StringType, BooleanType, NumberType} , the derivation of the least common super type is done wrong. It operates correctly if there are only two types (as mentioned https://github.com/IETS3/iets3.opensource/issues/505#issuecomment-889278536). So I broke the set down and called that function repeatedly until the least common type was determined.
By the way, the list of types you see above is only an example. Any types should be allowed!

Another point is that NumberTypes are only reduced to a common Numbertype (the least required super type) if setOfTypes only contains Numbertypes. This unnecessary constraint is now removed.

Additionally I recursively flattened the join types and optional types that are found within setOfTypes before determining the super type of their element types. So the resulting types becomes the super type of the element types of this special types. Afterwards this the resulting types are surround with the join and/or optional type to get only one super type.

**Previous review discussion: https://github.com/IETS3/iets3.opensource/pull/662**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,5 +15,5 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 ### Added
 
-- Tuples are now handled within the typesystem.
-	
+- Tuples are now handled within the typesystem. 
+  Insteasd of allowing JoinTypes within tuples we join different tuple types (we do not allow JoinTypes within tuples).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,5 +15,5 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 ### Added
 
-- Tuples are now handled within the typesystem. 
-  Insteasd of allowing JoinTypes within tuples we join different tuple types (we do not allow JoinTypes within tuples).
+- Tuples are now handled within the typesystem.
+   Instead of allowing JoinTypes within tuples we merge different tuple types by JoinTypes.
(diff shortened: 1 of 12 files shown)


arimer:
I think somehting went wrong. No tests are executed on CI: https://build.mbeddr.com/buildConfiguration/Iets3_BuildOrgIets3core/340795?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true

mgronover:
Regarding to https://github.com/IETS3/iets3.opensource/pull/662#pullrequestreview-1488192270.

The only usage of TypeingHelper.computeRegularSupertype() is in a quickfix about specifying the Value Type for an EnumDeclaration. So I don't think that missing the checkOptionTypeIsReuiredAndRecursivlyExtractElementTypes-method  is an issue.

arimer (commented):
Changes look fine to me, however we should add tests to see if the implementation also work properly in combination with TaggedTypes (types with additinal units).
 
- [x] add tests for types with units

HeikoBecker:
We are now maintaining a changelog on the maintenance branch. Can you document the typesystem change in it as part of the PR?

mgronover:
I've added some tests to check if the computation of the least common super type works with units.

arimer (dismissed):
Changes look good 👍🏼 

Do you maybe mind adding the following information to the changelog as well: 

> [We do not allow JoinTypes with in tuples and handle Runtime Errors](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F3212637104698705860)

Having this in the changelog, makes it obvious to communicate this limitation.

mgronover:
I added the following sentence to the CHANGELOG.md file:

`Instead of allowing JoinTypes within tuples we merge different tuple types by JoinTypes.`

PR #683, opened 2023-07-19
IETS3ExprEvalHelper: do not log re-thrown exceptions.

sergej-koscejev:
If the caller decides that the exception should be ignored, it should not appear in the log file. Logging the exception should be left to the caller.


alexanderpann (approved):

sergej-koscejev:
@alexanderpann can you please re-approve? I changed the base branch from 2021.1 to 2021.3.

alexanderpann (approved):

PR #707, opened 2023-08-07
Add error check for TemporalLiterals to identify slices with duplicate date.

AlexeiQ:
`TemporalLiterals` may have multiple slices with different values at different points in time. Slices with the same point in time are forbidden, because it can lead to nondeterministic behaviour. 
Thus a check is introduced to show an error on each slices point in time, if the date is a duplicate, i.e. the date is already used in another slice. The check retrieves the date literals in the slices' point in time and compares them against each other. It can handle date literals and (nested) references to constants defining a date via a date literal. 

If a point in time cannot be unwrapped to a date literal, e.g. the date literal is part of a list, then a warning is shown, that the date literal couldn't be retrieved. This indicates, that the duplicates-check  cannot be applied to this slice and its point in time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,12 @@ Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelo
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
 
-## August 2023
+## September 2023
 
 ### Added
 
-- Temporal Literal shows an error if two slices use the same point in time. The check can handle date literals and (nested) references to constants with a date literal.
-- Temporal Literal shows a warning if the point in time of a slice cannot be unwrapped to a date literal, rendering the duplicates-check mentioned above ineffective.
+- Temporal Literal shows an error if two slices use the same point in time. The check can handle date literals and (nested) references to constants with a date literal. (See [PR 707](https://github.com/IETS3/iets3.opensource/pull/707))
+- Temporal Literal shows a warning if the point in time of a slice cannot be unwrapped to a date literal, rendering the duplicates-check mentioned above ineffective. (See [PR 707](https://github.com/IETS3/iets3.opensource/pull/707))
     
 
 ## July 2023
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,29 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
+Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). 
+The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
+
+
+## August 2023
+
+### Added
+
+- Temporal Literal shows an error if two slices use the same point in time. The check can handle date literals and (nested) references to constants with a date literal.
+- Temporal Literal shows a warning if the point in time of a slice cannot be unwrapped to a date literal, rendering the duplicates-check mentioned above ineffective.
+    
+
+## July 2023
+
+### Fixed
+
+- Computation of the least-common-supertype for expressions with different return types has been fixed. The typesystem now correctly infers a join type (c.f. [original issue](https://github.com/IETS3/iets3.opensource/issues/505))
+- Naming constraint of IValidNamedConcept is customizable [original request](https://github.com/IETS3/iets3.opensource/pull/631)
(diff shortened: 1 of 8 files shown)


arimer (changes requested):
I think we should addrass some findings before the merge:

1. AT the moment you rely on the [render readable impl. ](http://127.0.0.1:63320/node?ref=r%3Aca744960-a59a-476e-a723-f852da3f606c%28org.iets3.core.expr.datetime.behavior%29%2F3885635233757570117)which is "just" the textual representation of the literal. How about we override the  a i[sSameAs()](http://127.0.0.1:63320/node?ref=r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%29%2F8872269265511400470) operation to enable the comparison of two literals instead?
2. your [unwrapToDateLiteral](http://127.0.0.1:63320/node?ref=r%3A782cab7d-c30f-4797-991c-cb17d0274086%28org.iets3.core.expr.temporal.typesystem%29%2F7119724182729197429) method might return null, how about we add an annotation [Nullable](http://127.0.0.1:63320/node?ref=3f233e7f-b8a6-46d2-a57f-795d56775243%2Fjava%3Aorg.jetbrains.annotations%28Annotations%2F%29%2F~Nullable) to indicate this explicitly? 
3. I think we can get rid of that [if-statement](http://127.0.0.1:63320/node?ref=r%3A782cab7d-c30f-4797-991c-cb17d0274086%28org.iets3.core.expr.temporal.typesystem%29%2F7119724182729054776)and aways calculate the `DateLiteral` via the `TemporalHelper`. 
 
 ```
 node<DateLiteral> date = TemporalHelper.unwrapToDateLiteral(slice.pointInTime);
 ```
should work just fine

4. in case `date.isNull` you still continue the calculation of `renderReadable` [here](http://127.0.0.1:63320/node?ref=r%3A782cab7d-c30f-4797-991c-cb17d0274086%28org.iets3.core.expr.temporal.typesystem%29%2F6199815761296886246) , imo the comparison should only be done in case we are dealing with dates (which are not null)
5. Instead of using while + nested if-else statements, consider maybe the usage of a switch. This helps at least to figure out the default case easier. 

 grafik

6. at the moment your implementation only covers `ConstantRefs` and `DateLiterals`. What about other expressions resulting in a date type and provide a date literal? Don't we need to handle functions as well?
 grafik
7. Could you also please link this PR or the original ticket in the updated release notes?

arimer:
@AlexeiQ Any updates on this one?

AlexeiQ:
@1) I've looked into the `isSameAs()` implementation suggestion, but I'm not sure how this should work out. Lets discuss it quickly when you're available, please.
@2) done
@3) done
@4) Good catch! I've moved the comparison into the `else`-condition of `if(date.isNull)`
@5) Switched the if-else-block to concept-switch. But we need to keep the while-loop, to iteratively unwrap the expression into a `DateLiteral`
@6) Yes, there are many other possible expressions that could lead to a `DateLiteral`. These two (directly defined `DateLiteral` and a `ConstantRef` pointing to `DateLiteral`). The idea is to implement customer-cases, that are also probably the most common use-cases and add further cases when they are actually needed instead of considering  every possible combination of expressions. 
@7) linked the PR, since the ticket is in a private repo

I've merged the 2021.3-maintenance-branch into this PR to resolve merge-conflicts.

AlexeiQ:
@arimer regarding 1)
I've added a `isSameAs` implementation to `DateLiteral` but don't see how I can use it properly in the check.
I rely on the `render readable`-method, because it provides me a string, that I can utilize in the set with the `set.contains()` method. `set.contains` doesn't rely on `isSameAs()` though. I cannot put a node directly into the set, because two `DateLiterals` with the same date are still two different nodes and therefore not equal, unless I redefine the `equals()` and `hashCode()`. I don't see how I can override them, since they are not offered from the `expression`-concept.

When or what for exactly is the `isSameAs()`-method used? I don't see where/what for it is used in iets3. Is it just the kernelf pendant to javas equals()?

Nevertheless the `render readable` usage isn't that problematic, even the rendered text is changes, since the same rendering is applied to all dateliterals in the check. Unless the rendering is changed to some to indistinctable representation, which wouldn't make any sense. But if we can rely on something better, I'm happy to do that.

arimer:
@AlexeiQ I've updated your check to make use of the `isSameAs()` implementation. Hope that's fine for you.

arimer (approved):
Thanks for the fixes. Looks fine from my point of view.

PR #705, opened 2023-07-31
Merge 2021.2 into 2021.3

HeikoBecker:

PR #709, opened 2023-09-21
Remove MetaUnit and refactor unit comparisons

HeikoBecker:
If the same unit is defined in different packages, this now leads to an error. The error shows the package in which they are not aligned as well as the root.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,15 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). 
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2023
 
+### Fixed
+
+- Unit comparisons for the **same** unit defined in **different** packages now lead to an error message for compatibility
+
+### Removed
+
+- the MetaUnit node was removed since it was not actively used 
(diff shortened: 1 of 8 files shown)


HeikoBecker:
PR is currently blocked by #705 which adds a changelog to the base branch. The changes in this PR need to be documented there.

alexanderpann (changes requested):
Please add a test that tests this new behavior.

alexanderpann (changes requested):
The rest of the changes look fine. 

- For me, it is important that now the scenario `val a : number<m> = 10 m` also throws an error when the units do not match. This should be another test case.

- You forgot to change the format of the error message as discussed.

HeikoBecker:
The recommended changes should be addressed now.
In the last commit I also fixed a test that broke because references to units now are non-transparent. The used `cm` unit pointed to the wrong file and thus the conversion could not be found anymore.

alexanderpann:
There shouldn't be too many unit definition that we need the model name. How about just \[root name\].\[unit name\]? The full information doesn't look good in the editor and also not in the type preview.

HeikoBecker:
Agreed. I have adapted the strings accordingly.

alexanderpann (dismissed):

HeikoBecker:
@alexanderpann The last commit only documents the changes in this PR in the changelog.

alexanderpann (approved):

PR #710, opened 2023-09-25
Merge state of 2021.3 into master

HeikoBecker:
I have merged in the changes from maintenance/mps20213, tested whether there were any missing migrations, and ran the re-runnable migrations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# Changelog
+
+All notable changes to this project are documented in this file.
+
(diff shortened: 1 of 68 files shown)


alexanderpann (commented):
I've tried to run Force Save All and get changes in build-allScripts.xml and some changed resolve infos. Is this a local problem on my end or why didn't the linters detect the issue?

HeikoBecker:
> I've tried to run Force Save All and get changes in build-allScripts.xml and some changed resolve infos. Is this a local problem on my end or why didn't the linters detect the issue?

I ran Forcee Save all which changed some editor infos for me locally. Can you check again?

Issue #711, opened 2023-09-26
`extensionPoint/IdentifierConfigurator` should be used by `MessageDefiniton` concept 


mgronover:
A customer wants to use  `MessageDefinition` with german umlauts. 
Therefore `MessageDefiniton` should use `extensionPoint/IdentifierConfigurator/` that allows the user to decide if german umlauts and paragraphs are allowed by defining (aka implementing) such an extensionPoint.

alexanderpann:
On a side note: I really question the way we handle special characters in `IValidNamedConcept`. The current handling is awful. We already have an extension point. Why not have a method where you can provide special characters that are allowed? The method should also map the character to an identifier e.g. ö to oe.

mgronover:
That would be much clearer and nicer i fully agree. Do you see a way to stay backward compatible (the methods `allowUmlaute and `allowParagraph` would become obsolete, won't they)?

alexanderpann:
You could read the return type of the existing two methods and add the corresponding character mappings to the map if they do not exist already. For the constraint of `IValidNamedConcept`, you can then maybe also use the mappings instead of the hardcoded characters. For more complicated cases it might be necessary to just have the ability to overwrite the full regex of the constraint.

mgronover:
Okay, I think we should go ahead here first and create a new ticket for that approach.

Reasons:
1. Nobody asked for more flexibility (maybe everyone silently accepted the situation so far).
2. We need tests to ensure the backward compatibility.
3. We will not forget that much more flexible approach and can work on it if there is a need for it.

Do you agree?

PR #712, opened 2023-09-26
Enable usage of MessageHandler and MessageArg with umlauts or paragraphs

mgronover:
Issue: #711 

Implemented  the [according methods](http://127.0.0.1:63320/node?ref=r%3Afc2b8fcf-559d-4f16-ae65-376838b59be6%28org.iets3.core.expr.messages.behavior%29%2F1203587247559183548) for these concept calling the functions of the extensionPoint/identifierConfigurator/ [class](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F4394073590212256166) and extended that by a method to return a [priority](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F1203587247559087196).

So when the extensionPoint is used the object with the [highest priority is taken into account](http://127.0.0.1:63320/node?ref=r%3Aac28053f-2041-47f6-806b-ecfaca05a64a%28org.iets3.core.expr.base.runtime.runtime%29%2F1203587247559093098) (others are ignored). The [DefaultIdentifierConfigurator](http://127.0.0.1:63320/node?ref=r%3A6c6155f0-4bbe-4af5-8c26-244d570e21e4%28org.iets3.core.expr.base.plugin%29%2F1203587247559088809) has the priority number 0.

~~**Breaking changes**~~
~~The extensionPoint got a new method. Any old implementation not having this new method will cause an error.~~


mgronover:
Ah, what a mess, that was happening, when I talked with Alexei on that theme.
So I under-pressed CTRL + Z  at least once and committed that. 
Sorry for that.

AlexeiQ:
@alexanderpann: @mgronover has integrated your feedback. Is it okay this way for you? Please unblock the PR in that case.

I've added an ExtensionPoint test with the help of @HeikoBecker to verify, that the correct EP implementation is taken, if multiple implementations are available.

HeikoBecker (commented):
Thanks @mgronover for updating the changelog and for your work on this.

I have not reviewed the code itself but tested its functionality together with @AlexeiQ and we added some minor improvements in how the extension point handles default values. 

I will leave it to @alexanderpann to approve this change.

AlexeiQ:
@HeikoBecker, @alexanderpann 
Since this PR has a breaking change (the EP has a new method, that needs overwriting), I guess we need to add a feature-flag here, to smooth the brake. I can do this, but it will take me a short while, since I currently don't have the time.

HeikoBecker:
@AlexeiQ Could we not give the default implementation mimicking the behavior from before in the default extension point implementation? That should suffice in my opinion.

mgronover:
The according method causing the breaking no more abstract (i.e. implemented). 
So all classes extending the IdentifierConfigurator will inherit the implementation of the `getPriorityLevel()` method.

arimer:
@HeikoBecker do you want to have a second look on this one?

arimer (changes requested):
@mgronover do you mind updating the changelog entry? Now that your changes is non breaking, the changelog is outdated.

mgronover:
The CHANGELOG.md already is updated.

arimer:
Yeah just saw it. Unfortunatelly I reviewed an older local version. 😓

arimer (approved):
Changes look for now to me. 👍🏼 

However we should discuss in general wich `IValidNamedConcept` should in addition benefit from this extension point mechanism. At the moment only [MessageArg](http://127.0.0.1:63320/node?ref=r%3A1a11ce0d-cf54-4682-9b8a-ab4ee15fc129%28org.iets3.core.expr.messages.structure%29%2F4026566441518474153) and [MessageDefinition](http://127.0.0.1:63320/node?ref=r%3A1a11ce0d-cf54-4682-9b8a-ab4ee15fc129%28org.iets3.core.expr.messages.structure%29%2F4026566441518134939) can be customized via this extension point.

@mgronover could you please add a ticket to not forget agout it.

- [ ] add ticket to collection ideas which IValidNamedConcept name should be customizable

alexanderpann (approved):

PR #713, opened 2023-09-29
Add a simple binding times language

HeikoBecker:
The binding times language extends the typetags with support for constraints on **when** a value will be bound.

With the language one can specify a temporal order in which values will be bound during execution.

Examples and tests illustrate the usage: 
[positive examples](http://127.0.0.1:63320/node?ref=r%3Af68e5d0e-8310-4fe0-9098-1dd49ca98947%28playground.bindingtimes%29%2F2278760416889396542)
[node tests](http://127.0.0.1:63320/node?ref=r%3Aed0dc053-fc6f-4fd7-b7ae-7adafb379a0c%28test.ts.expr.os.bindingtimes%40tests%29%2F2490986482722853162)

The tags are additive and do not influence exisiting interpreters. No interpreter for these tags is necessary to be implemented.


alexanderpann (changes requested):
*   Could you add a reference (maybe a web link) for the terminology that you use?
*   Short descriptions should not start with a capital letter.
*   You can write: `computation class Simulation after: Simulation` Is that on purpose?
*   Replacement rule: Please remove the message info statements or change the log level to debug.
*   I didn't find out if this is actually a problem but there is no check for computation classes that form a cycle e.g. A after B, B after A.
*   Your language is not part of any devkit in case you want to add it, for example to `org.iets3.core.expr.advanced.devkit`.
*   Shouldn't you only be able to refer to stages in the same computing class in the after block of the state?
*   The behavior of multiple initial computation classes and initial stages seems unclear to me. Should this even be allowed? Is the last declaration overwriting the previous declarations?

HeikoBecker:
@alexanderpann I havee updatet the code according to your suggestions.

The concepts from the language are loosely based on Lightweight Modular Staging in Scala: https://dl.acm.org/doi/abs/10.1145/1868294.1868314

As far as I remember, the grouping and the marking as an initial stage was not part of the language described there and could be considered additions.

It is intentional that multiple stages can be declared an initial root, both on the grouping and stages level, since separate computation chains may be modelled in parallel with this language.

alexanderpann (dismissed):
Looks good. One last tip: Use the intention `Specify Rule References in Test Annotation` in the test to reference the error statement. This way, you test for the specific error, and can do find usages on the error statement to find the test.

alexanderpann (approved):

PR #714, opened 2023-10-09
Improve unit error messages

HeikoBecker:
This is a follow-up to #709, rewording the error messages a bit.


alexanderpann (changes requested):
Looks better to me but the hyphen (-) should probably be a colon (:). Did you want to use an en dash (–) instead?

HeikoBecker:
> Looks better to me but the hyphen (-) should probably be a colon (:). Did you want to use an en dash (–) instead?

I would prefer using an en dash instead of the hyphen. MPS will show the message otherwise as `Error: Different types: ...`
and the double colon is annoying in my opinion. Will change immediately

alexanderpann (approved):
We should probably check the existing error messages in a separate ticket for other words that users don't understand. There are a few other places where we use the term supertype. This change seems fine to me.

Issue #691, opened 2023-07-23
Incorrect storing of results in lookup table


alexanderpann:
The lookup method always returns the value of the first column of the selected row instead of the request column.
The [code](http://127.0.0.1:63320/node?ref=r%3Aa13c86fe-c40a-4a06-b1a1-208e5bed665c%28org.iets3.core.expr.lookup.interpreter.plugin%29%2F5860825012169204468) for storing the result is wrong and should be replaced with the following code:

```
foreach row in node.rows { 
  Object rowVal = #(row.val); 
  result[rowVal] = new hashmap<Object, Object> ; 
} 
foreach cell in node.cells { 
  Object res = #(cell.val); 
  Object rowVal = #(cell.row.val); 
  Object colVal = #(cell.col.val); 
  result[rowVal][colVal] = res; 
}
```

PR #715, opened 2023-10-13
Fixes markus20213

markusvoelter:
A couple of smaller bugfixes and "quirk removals" in data tables and test vectors.
No really breaking changes. See discussion in Slack.

### DIFF
--- a/code/languages/org.iets3.opensource/.mps/libraries.xml
+++ b/code/languages/org.iets3.opensource/.mps/libraries.xml
@@ -7,8 +7,8 @@
           <value>
             <Library>
               <option name="name" value="mbeddr.plattform" />
-              <option name="path" value="${iets3.github.opensource.home}/build/dependencies/com.mbeddr.platform" />
-            </Library>
+              <option name='path' value='$PROJECT_DIR$/../../../build/dependencies/com.mbeddr.platform' />            
+   			</Library>
           </value>
         </entry>
       </map>
--- a/code/languages/org.iets3.opensource/.mps/libraries.xml
+++ b/code/languages/org.iets3.opensource/.mps/libraries.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project version='4'>
-  <component name='ProjectLibraryManager'>
-    <option name='libraries'>
+<project version="4">
+  <component name="ProjectLibraryManager">
+    <option name="libraries">
       <map>
-        <entry key='mbeddr.plattform'>
+        <entry key="mbeddr.plattform">
           <value>
             <Library>
-              <option name='name' value='mbeddr.plattform' />
-              <option name='path' value='$PROJECT_DIR$/../../../build/dependencies/com.mbeddr.platform' />
+              <option name="name" value="mbeddr.plattform" />
+              <option name="path" value="${iets3.github.opensource.home}/build/dependencies/com.mbeddr.platform" />
             </Library>
           </value>
         </entry>
--- a/code/languages/org.iets3.opensource/_spreferences/TestExecutionPreferences/models/org.iets3.opensource.__spreferences.TestExecutionPreferences.mps
+++ b/code/languages/org.iets3.opensource/_spreferences/TestExecutionPreferences/models/org.iets3.opensource.__spreferences.TestExecutionPreferences.mps
@@ -15,12 +15,12 @@
       <concept id="4473287864570292399" name="org.iets3.core.expr.testExecution.structure.TestExecutionConfig" flags="ng" index="3ZOQsN">
         <child id="4473287864570320758" name="executionMode" index="3ZOXzE" />
       </concept>
-      <concept id="4473287864570320840" name="org.iets3.core.expr.testExecution.structure.InterpreterExecutionMode" flags="ng" index="3ZOXxk" />
+      <concept id="4473287864570320810" name="org.iets3.core.expr.testExecution.structure.GeneratorExecutionMode" flags="ng" index="3ZOXwQ" />
     </language>
   </registry>
   <node concept="3ZOQsN" id="6pNCASbHXH8">
     <property role="TrG5h" value="KernelFTestExecution" />
-    <node concept="3ZOXxk" id="2R7vP7hzmjd" role="3ZOXzE" />
+    <node concept="3ZOXwQ" id="6wzrxL2Tnzf" role="3ZOXzE" />
   </node>
 </model>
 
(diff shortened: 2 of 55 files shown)


alexanderpann (changes requested):
Looks good, just one small change:

- The short description is missing for `CurrentDateExpr` and it should use the iets3Keyword

code/languages/org.iets3.opensource/.mps/libraries.xml
```diff
@@ -1,13 +1,13 @@
...
+    <option name="libraries">
       <map>
-        <entry key='mbeddr.plattform'>
+        <entry key="mbeddr.plattform">
           <value>
             <Library>
-              <option name='name' value='mbeddr.plattform' />
-              <option name='path' value='$PROJECT_DIR$/../../../build/dependencies/com.mbeddr.platform' />
+              <option name="name" value="mbeddr.plattform" />
+              <option name="path" value="${iets3.github.opensource.home}/build/dependencies/com.mbeddr.platform" />
```
alexanderpann, line 10: I am fine with the change but let's not switch between the project dir variable and the path variable in other PRs anymore.

code/languages/org.iets3.opensource/_spreferences/TestExecutionPreferences/models/org.iets3.opensource.__spreferences.TestExecutionPreferences.mps
```diff
@@ -15,12 +15,12 @@
       <concept id="4473287864570292399" name="org.iets3.core.expr.testExecution.structure.TestExecutionConfig" flags="ng" index="3ZOQsN">
         <child id="4473287864570320758" name="executionMode" index="3ZOXzE" />
       </concept>
-      <concept id="4473287864570320840" name="org.iets3.core.expr.testExecution.structure.InterpreterExecutionMode" flags="ng" index="3ZOXxk" />
+      <concept id="4473287864570320810" name="org.iets3.core.expr.testExecution.structure.GeneratorExecutionMode" flags="ng" index="3ZOXwQ" />
```
alexanderpann, line 18: We always had the execution mode set to the interpreter but again this change seems better for testing our languages anyway :+1:

alexanderpann (changes requested):
Thanks for the efforts, especially with writing a generator for the data table language:

*   PriorisationHelper: Please select a color for the Darcula theme with the constructor JBColor(regular,dark). We have another PR for Darcula support that we haven't merged yet.
*   Do you mind including the proposed fix in #691 for the lookupBy method as well as checking if it is compatible with your generator? The way the table lookup works right now is definitely wrong.
*   Adding the `genjava.advanced` devkit to the `datatable@tests` model, I see that the `EmptyValue` is not supported by the generator as well as the `equals expression` and the `not equals expression` in the test case tests.
*   e.get(col) in `reduce_DataTable` is not null safe and throws a NullPointerException if wrong values are supplied as arguments. Example for TestsForDataTableGenerator:
    *   assert Table.lookUpBy\<val1>(30)!.val2 equals 4

code/languages/org.iets3.opensource/.mps/libraries.xml
markusvoelter: I have changed it back.

code/languages/org.iets3.opensource/_spreferences/TestExecutionPreferences/models/org.iets3.opensource.__spreferences.TestExecutionPreferences.mps
markusvoelter: Ok, then no change here. I will keep the generator.

markusvoelter:
Regarding the more detailed comment:

- PriorisationHelper: I have changed it to a JBColor. I use the same color for both cases (see image at the bottom). However the color for the regular menu entries doesn't work. But I suspect that is the PR for Darcula that you were talking about.
- EmptyValue: I have added the generator for EmptyValue. Why were the tests green even though this was missing? I don't understand the equals/not equals expression thing though. Shouldn't this work generically if/when the expression is reduced to null? Also, same Q: why did this work until now?
- reduce_DataTable is null-safe.

Let's see if the build runs.

image

alexanderpann:
You have to restart MPS when you change the theme because it isn't applied everywhere automatically. The color doesn't look nice when using the Darcula theme:

Screenshot 2024-01-09 at 13 09 54


I just noticed that in `datatable@tests` the generated code for keys like `keyB` is also generated wrong when cells are null or empty values are used despite the constructor wanting two integer arguments in the example:
`public static final TableWithEmpties_Table_Entry keyB = new TableWithEmpties_Table_Entry(null);`
Exception:
```
constructor TableWithEmpties_Table_Entry in class test.in.expr.os.datatable.Generated_DataTable_Test.TableWithEmpties_Table_Entry cannot be applied to given types;
  required: java.lang.Number,java.lang.Number
  found: <nulltype>
  reason: actual and formal argument lists differ in length (/Applications/mps/platforms/2021.3/iets3.opensource_5/code/languages/org.iets3.opensource/tests/test.in.expr.os/test_gen/test/in/expr/os/datatable/Generated_DataTable_Test.java:181)
```

The reason why we didn't notice all those issues is that the devkit `genjava.advanced` is missing for `datatable@tests`, so the tests where only executed with the interpreter and not the generator on the CI. The issue with the equals expressions is that the type of the left and right side are not supported. This has nothing to do with the null values. Those reductions rules are not generic and are always specific for a type e.g: http://127.0.0.1:63320/node?ref=r%3A3490917a-a35f-43d7-9a1f-1311fda2da92%28main%40generator%29%2F8286534136181769422

markusvoelter:
> You have to restart MPS when you change the theme because it isn't applied everywhere automatically. 

Ok. Fixed. It's dark blue-ish now. 

> I just noticed that in datatable@tests the generated code for keys like 
> keyB is also generated wrong when cells are null or empty values are used 
> despite the constructor wanting two integer arguments in the example:

Fixed.

> The reason why we didn't notice all those issues is that the devkit genjava.advanced 
> is missing for datatable@tests, so the tests where only executed with the 
> interpreter and not the generator on the CI. 

Interesting that this doesn't fail somehow then. Fixed,

> The issue with the equals expressions is that the type of the left and 
> right side are not supported. 

Fixed.

This last test here
   WithDefault.lookUpBy<val1>(30)!.val2
cannot work. The lookup by 30 doesn't find anything, so lookup returns null
as it should. But we're forcing the null check away, so this has to result
in an NPE. Deleting the test.

alexanderpann (approved):
LGTM.

PR #716, opened 2023-10-15
TemporalValue/SliceValue is missing equals() and hashCode()

AlexeiQ:
For comparison of `TemporalValue`s and `SliceValue`s the `equals()` and `.hashCode()` have to be overwritten:
- Implemented `TemporalValue.hashCode()`, so that equal `TemporalValue`s have also the same hashcode and e.g. `list<TemporalValue>.distinct` is now working correctly
- Implemented `SliceValue.equals()` and `SliceValue.hashCode()`, so that `SliceValue`s can be checked for equality
- Added tests
- Removed obsolete vars in test

Compatibility:
- `SliceValue` is not compared by object address anymore. HashCode-calculation is not based on the objects address anymore.
- `TemporalValue`s hashCode-calculation is not based on the objects address anymore.


HeikoBecker (changes requested):
Thanks for the fix. I have 1 comment:

 In `SliceValue`: Why do we need to use `Object.equals` instead of `this.myTime.equals...`?
 Looking online, the `Object.equals` method will anyway relay to the `myTime.equals` method.
 
 
 Other than that, the changes look good to me.

AlexeiQ:
I'm not sure, that my new tests are running already on CI. I'll have to verify this.
Additionally, I need to add a feature-flag to smooth over the breaking change (equality for `SliceValue`, and hashes for `TemporalValue` will be different as before)

HeikoBecker:
> I'm not sure, that my new tests are running already on CI. I'll have to verify this. Additionally, I need to add a feature-flag to smooth over the breaking change (equality for `SliceValue`, and hashes for `TemporalValue` will be different as before)

I would argue that this is not necessary since it is an additive change which does not break existing code (as far as I can see). Maybe @alexanderpann has a better insight on this.

AlexeiQ:
> In `SliceValue`: Why do we need to use `Object.equals` instead of `this.myTime.equals...`? Looking online, the `Object.equals` method will anyway relay to the `myTime.equals` method.

This is correct. I used it, because `Objects.equals()` can handle the case, that `this.myTime` is null as well. We need it for the `this.value`, so I used here as well to be consistent, but also because it's possible to have a `null` there, even though it doesn't make sense in the domain.

alexanderpann:
> I would argue that this is not necessary since it is an additive change which does not break existing code (as far as I can see).

Agree and it only affects the temporal language which doesn't have many users. @HeikoBecker I've retriggered the review because the questions seems to be answered.

alexanderpann (changes requested):
You have to add the solution with the tests to `org.iets3.opensource.tests`, I didn't see the tests on the build server.

I believe that for `TemporalValue` you should include for both methods the object `defaultValue` which is set by the reduce method.

alexanderpann:
We had some issues with equality in temporal values (see test case) and the two equals helper classes, so I've merged them. @AlexeiQ can you please check if the changes are okay and then I'll approve.

AlexeiQ:
@alexanderpann I've checked your changes and most of it looks good 👍 

1. I've added a commit to fix a simple mistake while replacing the helper-calls
2. You've added `this.defaultValue` to the `TemporalValue.hashCode()`-method. 
If this applied to `hashCode()`, then it should also be applied to `equals()`, shouldn't it? Otherwise there could be a difference comparing two objects by `equals()` or by `hashCode()`. Have you changed your mind, since you also proposed that both method should be adapted?
> I believe that for TemporalValue you should include for both methods the object defaultValue which is set by the reduce method.

3. I think replacing the last `equals()-call` in `TemporalValue()` to `!EqualsHelper.equals(s.value(), otherSlice.value())` will cause a problem because if a `null` is compared against a `null` the helper will return `false`, though it should return a `true` in that case. This is achieved by calling `!Objects.equals(s.value(), otherSlice.value())` (see a7a89e1c)

alexanderpann (approved):

HeikoBecker (approved):
Thanks for the changes!

Issue #717, opened 2023-10-16
Nested short lambda expressions have wrong types inside the expression for it 


mgronover:
The following nonsense example code:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/4b01d98e-76d4-4d86-acbb-a804c650a194)

Results to the following error when generation was performed:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/da0f397a-bcd0-481d-838e-b3cba753404d)

The according part of the generated code:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/f799d907-6f46-4ac5-b9c0-e5019cce557e)

shows that the type of `it` is not correctly resolved in any case.

PR #718, opened 2023-10-16
Bugfix/nested short lambda expression

mgronover:
Issue: #717 

Simply corrected the generated type for `it` in short lambda expressions.


alexanderpann (approved):
The change looks fine to me. 
Regarding the typesystem errors: The fix is fine but why not just make the method return void? 

Maybe next time it would be nice if you could insert the generated code as text in the original ticket instead of an image, so that the reviewer can make a diff between the old and the new generated code.

mgronover:
I will take your advice for the next time.
And i made the methods return void now ;-) .

alexanderpann (approved):

PR #720, opened 2023-11-16
Stringvalidation: fix Runtime#findAllNumbers index calculation

alexanderpann:
The new test cases didn't work in the past.


mgronover:
I suppose it would be good to have another test:

`assert validate "xx19yy" [each number must be at position index 3] .ok equals true`

If I take a look at [OccurenceMatch](http://127.0.0.1:63320/node?ref=r%3A92c59e9f-983f-4c1f-bcfc-4dc3733c05db%28org.iets3.core.expr.stringvalidation.runtime.runtime%29%2F5001505504946008260) i suppose the field `position` describes the starting position of the string that matches (this is not exactly clear). 
In your case this will become the index of the last character of the match.

mgronover (approved):

PR #725, opened 2023-11-20
expr.collections: generalize sort method

alexanderpann:
The `sort` method of collections now supports more types not only numbers: all primitive types, the option type, all datetime types, the temporal type, and the record type.
 
- Records: The sorting order can be added through the intention `Add a Comparison Order`, otherwise, the records are sorted based on the declaration order of the members.
  
- Option: Sorting removes all `none` values since the underlying data structure of collections doesn't support null values.

Sorting can be enabled for other types by overwriting the method `Type#canBeSorted` and implementing the `Comparable` interface in the interpreter and generator for used Java classes. 

**Implementation detail**

The double cast [here](http://127.0.0.1:63320/node?ref=r%3A3490917a-a35f-43d7-9a1f-1311fda2da92%28main%40generator%29%2F2481710459440424430) is necessary because we use the Java class `Number` for KernelF numbers but this class doesn't implement the interface `Comparable`, only it subclasses ([explanation here](https://stackoverflow.com/questions/480632/why-doesnt-java-lang-number-implement-comparable)). That's why we have to do a double cast to erase this type information first.

Another interesting type that is currently not covered are tuples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). 
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## November 2023
+
+### Added
+
+- The `sort` method of collections now supports more types: all primitive types, the option type, all datetime types, the temporal type and the record type
+   - Records: The sorting order can be added through the intention `Add a Comparison Order`, otherwise, the records are sorted based on the declaration order of the members
+   - Option: Sorting removes all `none` values since the underlying data structure of collections doesn't support null values.
+
 ## September 2023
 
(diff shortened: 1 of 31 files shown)


mgronover (changes requested):
- For lists I simply would expect that list A is behind list B if list B is shorter (i.e. contains less elements). I wouldn't look into the lists and do some evaluation ! (see `OH.compare()`)
 
- For ranges of time and date I would expect, that ordering is done by the range itself (i.e. a timespan of 15s is more than one of 23s and of cause a day is less 1.5 days). No evaluation of begin and end within the comparison should be done i think. (`ArbitraryDateRangeValue`, `DiscreteDateRangeValue` and `TimeDeltaValue`)

mgronover (approved):
Still time/date ranges are compared by begin and end of the range but this might be okay. Reason: We could introduce a TimeSpan that could hold the duration as an numerical value.

So, let's go.

PR #726, opened 2023-11-21
String validation: Fix wrong indices in Runtime#findAllNumbers and XExecutor successor Check

alexanderpann:
This is a follow-up of https://github.com/IETS3/iets3.opensource/pull/720. The actual error was in a different class. While the predecessor code also doesn't look right to me, the tests are passing but I've at least added some more tests to make sure that the code is working.


mgronover (approved):
Good work!

Issue #728, opened 2023-11-24
Unit definition should detect cycles


HeikoBecker:
Currently, if a user defines a unit `unit myUnit := myUnit `, MPS crashes when using the unit in a KernelF expression. 

This should be fixed by adding a cyclicity check and raising an error.

PR #729, opened 2023-11-27
Add a cycle check to the unit definition + use correct conversion specifier when converting units

alexanderpann:
This PR fixes https://github.com/IETS3/iets3.opensource/issues/728. In addition, the automatically set conversion specifier of the `convert` expression and the `convertTo`  target is now used instead of the first applicable conversion specifier. This way, the converting even works with those cyclic definitions. The editor also shouldn't freeze anymore when using cyclic definitions. Example:
                      
```
quantity quantityA                                          
quantity quantityB                                          
                                                                               
unit a := b for quantityA
unit b := a for quantityB
unit c := c
                                                            
conversion a -> b {                                         
val as <no type> -> val * 1000                              
}                                                           
conversion b -> a {                                         
val as <no type> -> val / 1000                              
}                                                           

test case testConversionInterpreterCyclic [incomplete] {    
  assert convert[1 a -> b] equals 1000
  assert convert[1000 b -> a] equals 1
                                                            
  assert 5 a.convertTo(b)equals 5000
  assert 1000 b.convertTo(a)equals 1
}           
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to this project are documented in this file.
 
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). 
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
+
+## November 2023
+
+### Fixes
+
+- Unit definitions are now checked for cycles.
+- When units are converted, the specified conversion specifier is used, instead of the first one.
+
 ## October 2023
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixes
 
 - Unit definitions are now checked for cycles.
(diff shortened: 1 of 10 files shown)


HeikoBecker (changes requested):
Thank you for the work. The code itself looks fine.

But since the feature change is breaking, I would suggest the following:

- [x] recover the old behavior for the conversion specifiers
-  [ ] add an extension point that allows to configure which behavior is used (e.g. via a boolean flag)
- [ ] potentially show a warning that the old behavior is used 

This should make the code more robust in that existing developments (even though they relied on potentially buggy code) will not break.

alexanderpann:
I've enabled the new behavior for the tests which means that I had to add an extension to the test solutions. For KernelF tests, we don't have a good way to change extension points.

alexanderpann:
Not a breaking anymore since the old behavior was restored and you can now switch to the new behavior through the extension point

HeikoBecker (dismissed):
Changes look good to me now. The old behavior is preserved and the new behavior is easily accessible.

HeikoBecker (approved):
Thanks for the fix. With the new extension point there should not be a change in user-facing functionality.

Issue #732, opened 2023-11-28
Introduce Extension Point to customize Action showTrace


mgronover:
Because there is no mechanism to override an action in MPS we need to introduce an extension point to customize the trace explorer.

PR #733, opened 2023-11-28
Feature/732 introduce extension point for showTrace

mgronover:
Addresses issue: #732 

Because there is no way to override an action, we use an extension point to customize it.


alexanderpann (approved):
LGTM

PR #734, opened 2023-11-30
Adjust SimpleSortOp generator to return the correct type

alexanderpann:
The returned type was not correct. We didn't see it because we didn't assign the result to variables in our tests.


mgronover (approved):
LGTM

Issue #737, opened 2023-12-11
Trace explorer: avoid usage of deprecated entities.


mgronover:
Avoid deprecated things using the trace explorer.

PR #740, opened 2023-12-12
trace explorer: avoid usage of deprecated entities

mgronover:
Addresses issue: #737

Little changes to avoid deprecated CTORs and option usages.


alexanderpann (approved):
LGTM.

PR #742, opened 2023-12-13
Fix IndexExpr logic in generator and interpreter + support it in the all and any operation

alexanderpann:
- The `index` expression in collections operations now works correctly in nested expressions in the interpreter and generator.
- The `all` and `any` operation of collections now also support the `index` expression. The concepts `AllWithIndexOp` and `AnyWithIndexOp` are therefore deprecated.


mgronover (approved):
LGTM.

mgronover (approved):

PR #743, opened 2023-12-14
Don't call the interpreter with computation trace enabled with it is not necessary

alexanderpann:
The class `IETS3ExprEvalHelper` was deprecated and a new class `IETS3ExprEvaluator` was introduced that can also influence the creation of the computation trace. The interpreter doesn't automatically create a computation trace anymore that is discarded anyway.
The performance improvements are unfortunately only minimal and are only noticeable for larger computations e.g.
`10.upto(500).map(|it: number{0} => it * 2|).where(|it >= 50|).size` now sometimes takes 125 ms instead of 130 ms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - The `all` and `any` operation of collections now also support the `index` expression. The concepts `AllWithIndexOp` and `AnyWithIndexOp` are therefore deprecated.
 
+### Changed
+
+- The class `IETS3ExprEvalHelper` was deprecated and a new class `IETS3ExprEvaluator` was introduced that can also influence the creation of the computation trace.
+
 ## November 2023
 
 ### Added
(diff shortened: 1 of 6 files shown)


mgronover (changes requested):
- It seems to me that we could remove `getInternalHelper() `from `IETS3ExprEvaluator` and offer `getLastLog()` directly instead.
- check if `IETS3ExprEvalHelper.getInterpreter()` can be deleted made private  or directly use InterpreterEvaluationHelper.getInterpreter()

alexanderpann:
> *   check if `IETS3ExprEvalHelper.getInterpreter()` can be deleted made private or directly use InterpreterEvaluationHelper.getInterpreter()

I've deprecated the full class `IETS3ExprEvalHelper` and its methods call the new implementations in `IETS3ExprEvaluator`. Why should I introduce a breaking change or use the other method directly?

arimer:
Unfortunatelly this change also deactivated the creation of coverage data and therefore introduced this bug #1366. Meaning: Coverage calculation from assessments was deactivated.

PR #745, opened 2023-12-20
Implement named tuples

alexanderpann:
The main motivation was to have multi-decision tables return a named tuple so that you can reference the result columns per name instead of per index. The named tuples are compatible with each other and normal tuples if they structurally match.

Known/found issues:
- When a result column in a dectab is renamed the references to this field break.
- The dot expression implementation of tuple types in the [generator](http://127.0.0.1:63320/node?ref=r%3A4243557f-1c7a-4d6b-953a-807576e4bee7%28org.iets3.core.expr.genjava.base%40generator%29%2F3863896640935437510) looks very broken to me. I've disabled two general operations on tuples as a result.

Screenshot 2023-12-20 at 12 07 19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Changed
 
 - The class `IETS3ExprEvalHelper` was deprecated and a new class `IETS3ExprEvaluator` was introduced that can also influence the creation of the computation trace.
+- Multi-decision tables now return an internal stored named tuple instead of a tuple. The new return type is compatible with the old type.
 
 ## November 2023
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,13 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - The `index` expression in collections operations now works correctly in nested expressions in the interpreter and generator.
-- Generation of nested short lambda expression now use the correct type for "it" as a variable.
+- Generation of nested short lambda expression now uses the correct type for "it" as a variable.
 - Collections: The index expression now works with collection types in the generator.
 
 ### Added
 
 - The `all` and `any` operation of collections now also support the `index` expression. The concepts `AllWithIndexOp` and `AnyWithIndexOp` are therefore deprecated.
-- Named tuples were added. The can be declared at the top level with the keywords `named tuple`. Fields can be accessed witht the `.` expression.
+- Named tuples were added. They can be declared at the top level with the keywords `named tuple`. Fields can be accessed witht the `.` expression.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Added
 
 - The `all` and `any` operation of collections now also support the `index` expression. The concepts `AllWithIndexOp` and `AnyWithIndexOp` are therefore deprecated.
+- Named tuples were added. The can be declared at the top level with the keywords `named tuple`. Fields can be accessed witht the `.` expression.
 
 ### Changed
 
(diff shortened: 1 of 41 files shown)


mgronover (changes requested):
#### CHANGELOG.MD
Fix typo: The --> They

#### SwitchToplevelExpressions
Line 75: Remove: BigInteger i = new BigInteger("0")

#### all @tests.mps: tuples, dectabs & m1
These look to me as they use the reflective editor for _named tuple declaration_.
I think they should use the regular editor instead or both look nearly the same.

alexanderpann:
> These look to me as they use the reflective editor for named tuple declaration.

I am not sure what you mean. This is how it looks for tuples@tests:

Screenshot 2024-01-11 at 14 09 55

markusvoelter:
What is the difference between a named tuple and a record? The declarations seem identical except for the keyword.

arimer:
@markusvoelter I am just started looking into it, but yes the MM looks nearly identical.

arimer (changes requested):
I think in general the idea is good to simplify the table access. It would be an advantage to have the possibility to access the results of a table via the colomn name. However I am not conviced that we need yet an other type declaration in the form of named tuples. 

In general before merging, we should clarify and solve the following findings

### Findings

#### Metamodel 

1. What is the use of [EmptyTupleMember](http://127.0.0.1:63320/node?ref=r%3Adfbbc430-47fe-4054-9d32-72c481150c72%28org.iets3.core.expr.toplevel.constraints%29%2F5749748470417075380) ? It can't be created by the user and is not used anywhere in the code
2. Could you please explain the idea behind [IMultiDecTab.resultTupleDeclaration](http://127.0.0.1:63320/node?ref=r%3A6bb59b1e-6116-48ad-b11d-2641d4f6b6a1%28org.iets3.core.expr.util.structure%29%2F1609062041026819473)? I am really having a hard time getting the motivation for this construct. 
3. Having [model listeners](http://127.0.0.1:63320/node?ref=r%3Affa27aa1-4bbf-456c-82f8-8174ae6a9559%28org.iets3.core.expr.util.listeners%29%2F1609062041027043359) that should "set" and "update" that construct indicates  code smell. I think we should discuss what you exactly want to achieve.
4. If we go down that rabbit hole, why not implement an equivalent concept for tuples as it was done in [InlineRecordType](http://127.0.0.1:63320/node?ref=r%3Ada65683e-ff6f-430d-ab68-32a77df72c93%28org.iets3.core.expr.toplevel.structure%29%2F8293738266741050575)?
5. There proposed changes are not backwards compatible with already existing  `TupleAccessExpr` instances, that access table results via index.


#### Minor proposals

1. I think [NamedTupleValue]() should also reference/indicate the name of the tuple you are instantiating. Otherwise it will get quite messy when you have different tuples, where the members are called equally
2. you have [broken reference](http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F7241462708334131240)s in your test. Looks like those nodes were just copied over from [multidectab](http://127.0.0.1:63320/node?ref=r%3Aae52e1b7-6479-4187-9e09-836b57124d46%28test.in.expr.os.utils%40tests%29%2F457187122965369180). I think it makes sense to avoid addition errors in node tests, which are not part of the "error under test"

#### Usage

1. I am having a hard time creating a multi decision table with a ResultColDef. Using the provided intention [DeclareReturnType](http://127.0.0.1:63320/node?ref=r%3Aecf4180a-6e34-47e0-90d1-b073b95f6f24%28org.iets3.core.expr.util.intentions%29%2F8697767715749753211) results in a stacktrace. Also creating such a table manually does not work. 

This [exception.txt](https://github.com/IETS3/iets3.opensource/files/14094543/exception.txt) is thrown during creating.

arimer:
Closing in favor of #769. Most of the ideas presented here were realized in non braking and backwords compatible way. Named based access can now co-exist with the old index based one. 

In case we want to introduced named tuples as a first level concept, we can have a second look on this inplementation here.

PR #751, opened 2024-01-22
Improve DateLiteral constraints for months and days

alexanderpann:
The problem is `DateLiteral#toDate` which throws a `DateTimeException` because of possible invalid parsed years or days (default value: 0). We can't fix the issue but rather prevent the user from entering invalid values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - The PCollections library was updated to version 4.0.1.
 
+### Fixed
+
+- The DateLiteral constraints for months and days were improved.
+
 ## December 2023
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


arimer:
Will your change lead to errors on once "valid" values or is the change backwards compatible?

alexanderpann:
In the generator and behaviors the same values will be returned. It only prevents the user from entering an invalid month or day. There was an error before from a checking rule when the date is not valid. I am wondering if I miss a part in the code where we could prevent the exception all together. The DateLiteral#sameAs method throws but with the new code the user shouldn't be able to enter such a date. Existing instances should be fixed manually but they should make the model checker fail anyway.

arimer:
You are probably refering to this [checking rule](http://127.0.0.1:63320/node?ref=r%3A8fb8c021-a1ab-41a1-a284-a1a372e0faad%28org.iets3.core.expr.datetime.typesystem%29%2F2425357398685009203). I am just wondering, in which case did the checking rule not cover the invalid input of the user? Do you have an example?

alexanderpann:
I don't that's why I am also confused by this request. Maybe the exception happens when a new literal is created and we don't have sensible defaults?

arimer:
I think having proper exception handing either in [toDate()](http://127.0.0.1:63320/node?ref=r%3Aca744960-a59a-476e-a723-f852da3f606c%28org.iets3.core.expr.datetime.behavior%29%2F2425357398686407867) or on the caller side would be a better solution.

arimer:
Closing in favor of #752

PR #755, opened 2024-01-25
Activate automatic merging of PRs into the next maintenance branch

alexanderpann:


sergej-koscejev (approved):

.github/workflows/PR-into-next-version.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  push:
+    branches:
+      - 'maintenance/mps*'
+
+jobs:
+  create-pr:
+    uses: specificlanguages/cascading-merge/.github/workflows/workflow.yml@v1
+    with:
+      exclude_branch_prefix: maintenance/mps3
```
arimer, line 12: @alexanderpann what is this exclude good for? Do we really have branches of `mps3`?

alexanderpann: Of course: https://github.com/IETS3/iets3.opensource/tree/maintenance/mps34

alexanderpann: It is debatable what we should do with those ancient branches. For historical reasons, they might be useful if you want to go back to a specific MPS version. I've done this, for example, in mbeddr.core because I wanted to checkout out the code before the MPS-Extension split. Not sure if we really need them here. What do you think?

Issue #761, opened 2024-01-25
`InfHelper.equals` not working as expected


alamqadem:
The method returns the wrong result for the following cases: 0.0 and 0 or -0 and 0.
Example with the Console:
```java
> {
    #print InfHelper.equals("0", "0");
    #print "\n";
    #print InfHelper.equals("0", "0.0");
    #print "\n";
    #print InfHelper.equals("0", "-0");
    #print "\n";
  }
  true
  false
  false
```

It shouldn't do string comparison, but numerical one like `greatOrEquals`.

arimer:
Should be closed by https://github.com/IETS3/iets3.opensource/pull/762. 
@alamqadem could you please verify and re-open in case the bug is still there?

alamqadem:
> Should be fixed by #762. @alamqadem could you please verify and re-open in case the bug is still there?

I just pulled `maintenance/mps20222` and it's not working.
![image](https://github.com/IETS3/iets3.opensource/assets/10254994/df4adb31-5352-49be-b96a-ab8b4d7d27eb)

arimer:
@alamqadem it was merged on 20213, so I guess some cascading merges are missing.

arimer:
@alamqadem which iets3 versions are you using atm. where that bug should be fixed?

alamqadem:
> @alamqadem which iets3 versions are you using atm. where that bug should be fixed?

We just need it for MPS2022.2

arimer:
A fix is now merged to `maintenance/mps20222`. Feel free to try it out and to close the issue.

alamqadem:
> A fix is now merged to `maintenance/mps20222`. Feel free to try it out and to close the issue.

It works now, thanks 🙂

PR #762, opened 2024-01-29
IntHelper.equals now correctly deals with the combination of 0.0,-0.0, 0 and -0

alexanderpann:
This PR fixes https://github.com/IETS3/iets3.opensource/issues/761.


alexanderpann:
I've reverted the accidental change in the intentions model. MPS shows me that the files are identical.

markusvoelter:
You wrote tests of course, but shouldn't the equals test use the == 0 test instead of 
   new BigDecimal(l).compareTo(new BigDecimal(r)) >= 0

alexanderpann:
Yes, that was a copy-paste error. I fixed it. Scary that despite so many tests, I still missed such an easy case.

markusvoelter:
Yes, indeed.

markusvoelter (approved):

PR #764, opened 2024-01-30
Fixed the StringLengthTarget

markusvoelter:
There was a strange condition in the StringLengthTarget Java generator that prevented correct generation inside a block or if. I have removed this condition. Nobody seemed to remember why it was there in the first place.


alexanderpann (approved):
LGTM.

PR #767, opened 2024-01-31
Migrate to MPS 2023.2

alexanderpann:
- There was a change in the declaration of stub models. See: https://youtrack.jetbrains.com/issue/MPS-36814/Update-documentation-about-solution-settings-for-stubs
- The tests now run as JUnit5 tests. See: https://www.baeldung.com/junit-5-migration
- Runtime modules also have to be declared correctly or you will get ClassNotFoundExceptions when executing tests out-of-process.
- To avoid [MPS-25087](https://youtrack.jetbrains.com/issue//Concept-icons-do-not-work-when-building-standalone-IDE) which is the issues where icons are not showing up in actions which results in [MPS-36726](https://youtrack.jetbrains.com/issue/MPS-36726/NPE-at-java.base-java.util.Objects.requireNonNullObjects.java209) and makes even main menu entries crash, the fix in https://github.com/mbeddr/mps-qa/pull/24 must be applied to all projects that this repository depends on.


alexanderpann:
In a different project where I am testing, I get compilation errors with tests, so this PR can't be merged yet but it can be reviewed.

sergej-koscejev (approved):

Issue #750, opened 2024-01-22
Enums should be sortable


mgronover:
Enumeratian should have an optional order. This can be the position of the literal within the declaration, the literal itself (lexical order) or the value assigned to the literal.

arimer:
Closed by PR #770

PR #770, opened 2024-02-05
Enumerations can have an order

mgronover:
This PR addresses issue #750 

It is fully compatible with previous versions (no breaking changes).
No migration is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/). 
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## February 2024
+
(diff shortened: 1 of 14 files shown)


arimer:
I fixed 1,2,4,6 and and 7. Hope that is ok for you :)

arimer:
@mgronover is this PR ready for a second round of review?

mgronover:
Not yet, there are some test cases to do.

mgronover:
Thanks for fixing 1,2,4,6 and and 7! 

add 3)
We need it to call compareTo on the object of the value associated with the enum literal.

add 5)
I saw it in the code: you also fixed item 5 of your list above. Thanks!

I added some test for enum values that implement canBeSorted() retruning "true". Additionally I introduced an assertion to stay safe in the future. Ensuring the interface Comparable is implemented for these classes.

arimer (approved):
The implemenation looks finally good :) 

Please a proper PR description and sketch really short what was done on a very high level.

PR #773, opened 2024-02-08
Sortable enums

mgronover:
this PR addresses: #770


arimer (approved):
Already reviewed in #750

Issue #779, opened 2024-02-21
Lazy initialization for enum values during interpretation


mgronover:
Instead of enumeration values being initialized directly when the interpreter is evaluating the according literal, it is needful, to do this later i. e. when the value is requested.

arimer:
Fixed via #781

PR #780, opened 2024-02-21
lazy initialisation for values of an enum #779

mgronover:
This PR addresses issue: #779 

Customer requires to create instances of an EnumLiteral class without the evaluation of an associated value of the literal or the enum.

This changes only have a technical aspect.


arimer:
@mgronover I guess the target branch should be maintenance/mps20213

PR #781, opened 2024-02-21
Bugfix/779 lazy initialization for enum values

mgronover:
Addresses issue: #779

Customer requires to create instances of an EnumLiteral class without the evaluation of an associated value of the literal or the enum.

The [getDefaultLiteral()](http://127.0.0.1:63320/node?ref=r%3Acfd59c48-ecc8-4b0c-8ae8-6d876c46ebbb%28org.iets3.core.expr.toplevel.behavior%29%2F1075037996903058823) of an EnumDeclaration can be overriden now (added "virtual"):

image

The reason is that it might depend on the order of the enumeration what the first element is.

This changes only have a technical aspect.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/behavior.mps
@@ -6670,6 +6670,7 @@
     </node>
     <node concept="13i0hz" id="VFjlN5t4Q7" role="13h7CS">
       <property role="TrG5h" value="getDefaultLiteral" />
+      <property role="13i0it" value="true" />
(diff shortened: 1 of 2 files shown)


arimer (commented):
Why did you remove the additional assert
grafik

mgronover:
Because within the `getInstance()` methods, the constructor is called with `null`  for the `literalValue` field.
That means, the literalValue will never be a Comparable.
When someone compares two EnumLiteral another assertion is active and will fail if EnumSortByValue is used as order and literalValue contains something that is not a Comparable.

arimer (approved):
Changes look ok for now. 

It is a good start to make [getDefaultLiteral()](http://127.0.0.1:63320/node?ref=r%3Acfd59c48-ecc8-4b0c-8ae8-6d876c46ebbb%28org.iets3.core.expr.toplevel.behavior%29%2F1075037996903058823) virtual to allow users to override and customize that implementation. 

However I think the framework should deal itself with all possible orderings and their influence of the `getDefaultLiteral()` implementation. 

- [x] Could you please create a ticket for this?

Issue #774, opened 2024-02-08
Compilation error: Incompatible types exception thrown from generated TupleAccessExpr


arimer:
In case `TupleAccessExpr` is used to access a return type of a multi decision table, the generated java code does not calculate a proper type. Instead `Object` is generated which leads to a Type Exception when it is assigned to a more concrete type.

See screenshot for more information

<img width="2098" alt="grafik" src="https://github.com/IETS3/iets3.opensource/assets/16648247/83e4808b-c09a-4db6-bf44-ac8b2c4ea2a3">

arimer:
closed via #787

PR #787, opened 2024-02-26
core.expr.genjava.base: added type cast for TupleAccessExpr

arimer:
This PR  updates the reduction rule for TupleAccessExpr and adds an additional type cast in case a typesystem type is present. 
Those change prevents issue #774 from happening and avoid a compilation error.


alexanderpann (approved):
LGTM.

mgronover (approved):
Fine

Issue #786, opened 2024-02-22
Using list types in conditional expressions (if, alt) results into compiler failures not fetched by tests


mgronover:
I wrote some code as an example, that currently is not able to be successful generated to Java:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/4207baa9-2dbc-4457-aaff-11e724a5e4ac)

Rebuilding it results in warning issues:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/1544ade1-f763-4bde-bfd7-7b4e0cb6957f)

and the generated source is not compilable at all:
![image](https://github.com/IETS3/iets3.opensource/assets/38247916/6e409339-6e3e-49c4-a7ed-b21e8bff855c)

But now it is getting even worse:

When I switch to ` KernelF Tests: Generator` and `KernelF Tests: Generator (workspace)` and trigger the test to be executed.....

"WOW", I get all test green:
![image](https://github.com/IETS3/iets3.opensource/assets/38247916/95b5f6b6-6811-4d7a-8ce2-c9f22df8b4ca)

That's not what the way it should be!

If you wanna try out:
https://github.com/IETS3/iets3.opensource/tree/bugfix/MeetTypeCreationOnFunctionGeneration

mgronover:
I suppose we require a suitable test environment to check if some domain spec. code can be successful translated (generated **and** compiled) into something executable (without warnings from the typesystem, the generator or compiler) .

mgronover:
Here they are:
![image](https://github.com/IETS3/iets3.opensource/assets/38247916/3357870d-491a-4651-95ba-3fd7838cb0df)

arimer:
@mgronover can we close this one due to the merge in #790 ?

mgronover:
Yes, we can.

PR #790, opened 2024-02-27
Added typing information within generated sources to create TreePVectors.

mgronover:
This ticket addresses the warning and errors mentioned in #786

The generation of the according factory method or TreePVector is now done with the type parameter to avoid MeetType derivation during the generation process.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.base/generator/template/org.iets3.core.expr.genjava.base@generator.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.base/generator/template/org.iets3.core.expr.genjava.base@generator.mps
@@ -6,6 +6,7 @@
     <use id="fd392034-7849-419d-9071-12563d152375" name="jetbrains.mps.baseLanguage.closures" version="-1" />
     <use id="f2abf2a6-552e-4906-ab58-930f9a79aff4" name="org.iets3.core.expr.genjava.contracts" version="-1" />
     <use id="b401a680-8325-4110-8fd3-84331ff25bef" name="jetbrains.mps.lang.generator" version="4" />
+    <use id="d7706f63-9be2-479c-a3da-ae92af1e64d5" name="jetbrains.mps.lang.generator.generationContext" version="2" />
     <devkit ref="a2eb3a43-fcc2-4200-80dc-c60110c4862d(jetbrains.mps.devkit.templates)" />
   </languages>
   <imports>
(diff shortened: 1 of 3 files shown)


arimer (commented):
In your code you are accessing `listLiteral.typeConstraint.typeConstraint` [here](http://127.0.0.1:63320/node?ref=r%3A4243557f-1c7a-4d6b-953a-807576e4bee7%28org.iets3.core.expr.genjava.base%40generator%29%2F8100874504470626464) to calculate the concrete type of a `TreePVector<concreteType>`. In your particlar case that generator will only by applied for `ListLiterals` that do not have any elements. 

Can you maybe explain to my what the [ListLiteral.typeconstraint ](http://127.0.0.1:63320/node?ref=r%3A61b1de80-490d-4fee-8e95-b956503290e9%28org.iets3.core.expr.collections.structure%29%2F8694548031077041593) is supposed to be used for? I have a hard type understanding how your change have impected the test results of the tests you've added. To be more concrete:  Inside the [test suite functionType](http://127.0.0.1:63320/node?ref=r%3Af9a70877-66f3-464b-86cb-67a9d543be8c%28test.in.expr.os.functions%40tests%29%2F728017840040128456) there is not a single presence of `ListLiteral.typeconstraint`, which means your generator change would not have any impact on the test suite execution.

mgronover:
I typified the call to `TreePVector.<...>from()` just to to make that type more explicit and ensure that the type inference from the elementtype is not required. So the type is already known without inspecting the type of the elements.
This is the reason of typifying the method call and elimination the MeetType problem at various places and I just decided to do this for the most calls in an equal way. 

> Can you maybe explain to my what the [ListLiteral.typeconstraint ](http://127.0.0.1:63320/node?ref=r%3A61b1de80-490d-4fee-8e95-b956503290e9%28org.iets3.core.expr.collections.structure%29%2F8694548031077041593)is supposed to be used for?

The [ListLiteral.typeconstraint ](http://127.0.0.1:63320/node?ref=r%3A61b1de80-490d-4fee-8e95-b956503290e9%28org.iets3.core.expr.collections.structure%29%2F8694548031077041593)is used to ensure that a given list will only contain elements that have a type fulfilling that type constraint.

> I have a hard type understanding how your change have impected the test results of the tests you've added. To be more concrete: Inside the [test suite functionType](http://127.0.0.1:63320/node?ref=r%3Af9a70877-66f3-464b-86cb-67a9d543be8c%28test.in.expr.os.functions%40tests%29%2F728017840040128456) there is not a single presence of ListLiteral.typeconstraint, which means your generator change would not have any impact on the test suite execution.

There is no impact at all here. The ListLiteral is tested by numerous other test cases.

PR #793, opened 2024-02-27
Separate REPL and tracing from the core language

alexanderpann:
- A new devkit `org.iets3.core.expr.base.devkit` was added that excludes the REPL and tracing language.
- `org.iets3.core.expr.core.devkit` and `org.iets3.core.expr.advanced.devkit` were extracted into a separate plugin `org.iets3.core.expr.devkits`. Why? If you exclude the REPL or tracing language, the core devkit would show an error that it can't load these languages. Instead, you should use `org.iets3.core.expr.base.devkit` then.
- A new devkit `org.iets3.core.expr.interpreter.devkit` was created for all expression interpreters.
- A new devkit `org.iets3.core.expr.repl.devkit` was created for the REPL language.


HeikoBecker:
Overall changes look good. Two questions:

* Why is thre now a `tracing.plugin` used in the core devkit?
* Where are the interpreters gone in the core devkit? I see that they were removed in favor of the repl plugin but that one does not include the interpreters as far as I could see.

HeikoBecker (approved):
Thanks for the clarifications @alexanderpann. Changes look good to me. But I would like to at least hear the opinion of someone else (e.g. @arimer or @AlexeiQ) before this is merged.

arimer:
@alexanderpann What was the reasong for not doing that change on an earlier version like `maintenance/mps20213`?

alexanderpann:
@arimer It was a customer (feature) request and not a bugfix and therefore is done in the branch that they are using.

arimer:
>Should we count refactorings/restructurings as bug fixes?

I think we should always, regardless of the kind of change, have in mind that we still support multiple maintenance branches that are still actively (!) under maintenance. Therefor we should handle changes that have impact on older versions with high caution and try to avoid situations where overhead (conflicts etc.) is created due to such a change.

alexanderpann:
This would mean that we must always make breaking changes/changes with overhead in the oldest supported version to avoid that. The problem is that we sometimes have feature requests in maintenance branches that are also breaking changes that conflict with changes such as this one. Only bug fixes should be done on those branches in theory but we can't do that.

PR #807, opened 2024-02-29
build.gradle: added build fail on test failure

arimer:
At the moment our TC configuration allows the execution of the `publish` task despite of failing tests. This PR adds additional functionality to fail our build in case we have failing tests. 

We want to still execute all tests and collect their results as part of the junit result xml file. However in addition we want the build to fail in case tests are failing.  Therefore the `buildAndRunTests()` task was updated to parse the result xml file and throw an exception in case of test failures.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 import de.itemis.mps.gradle.*
-
+import groovy.xml.XmlSlurper
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import java.time.format.FormatStyle
@@ -212,6 +212,16 @@ task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
             report(format: 'frames', todir: "$buildDir/junitreport")
         }
         ant.echo("JUnit report placed into $buildDir/junitreport/index.html")
+
+        // evaluate junit result and fail on error
+        def junitResult = new XmlSlurper().parse(file('TESTS-TestSuites.xml'))
+        def failures = junitResult.'**'.findAll { it.name() == 'failure' }
+        def errors = junitResult.'**'.findAll { it.name() == 'errors' }
+
+        if (failures || errors) {
+            def amount = failures.size() + errors.size()
+            throw new GradleException(amount + " JUnit tests failed. Check the test report for details.")
+        }
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -212,8 +212,13 @@ task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
             report(format: 'frames', todir: "$buildDir/junitreport")
         }
         ant.echo("JUnit report placed into $buildDir/junitreport/index.html")
+    }
+}
 
-        // evaluate junit result and fail on error
+task failOnTestError() {
+    description 'evaluate junit result and fail on error'
+    doLast {
+        
         def junitResult = new XmlSlurper().parse(file('TESTS-TestSuites.xml'))
         def failures = junitResult.'**'.findAll { it.name() == 'failure' }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -212,6 +212,16 @@ task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
             report(format: 'frames', todir: "$buildDir/junitreport")
         }
         ant.echo("JUnit report placed into $buildDir/junitreport/index.html")
+
+        // evaluate junit result and fail on error
```
arimer, line 216: Tbh. I am not sure if we make this check only work in case we run on CI and not locally. Feedback is highly welcome here.

alexanderpann: Why should be there a difference when you run it locally? I am always in favor of the local and CI builds behaving the same way.

sergej-koscejev:
How about setting “halt on failure” to true in the MPS build scripts instead (on the ‘run tests from modules’)?

alexanderpann:
> How about setting “halt on failure” to true in the MPS build scripts instead (on the ‘run tests from modules’)?

With this setting, the build fails immediately when the first test fails. We want to run all tests and see what tests are failing, not only the first result.

sergej-koscejev:
> With this setting, the build fails immediately when the first test fails. We want to run all tests and see what tests are failing, not only the first result.

My experiments show that the build fails after the `<junit>` task but all tests in the test suite are run.

arimer:
> https://mps.builds.itemis.cloud/buildConfiguration/Iets3_BuildOrgIets3core/356044?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildTestsSection=true&expandBuildProblemsSection=true

Our TC behaviour is slightly different. All tests are executed until one failing test is detected. Afterwards the test execution of the remaining test is not done and those tests are does not end up in the junit xml file evaluated by TC.

Here an example of a build with `halt on failure` enabled.

https://mps.builds.itemis.cloud/buildConfiguration/Iets3_BuildOrgIets3core/356044?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildTestsSection=true&expandBuildProblemsSection=true

We have more then 5k tests, but only 3.2k tests were found.

build.gradle
```diff
@@ -215,6 +215,22 @@ task buildAndRunTests(type: TestLanguages, dependsOn: buildLanguages) {
...
+        def errors = junitResult.'**'.findAll { it.name() == 'errors' }
+
+        if (failures || errors) {
+            def amount = failures.size() + errors.size()
+            throw new GradleException(amount + " JUnit tests failed. Check the test report for details.")
+        }
+    }
+
+}
+buildAndRunTests.configure { finalizedBy failOnTestError }
```
arimer, line 233: @sergej-koscejev as suggested I've added a finalizedBy dep. from our test task to a task failing the build. 

Pls. have a second look

alexanderpann: Are you sure about the way to calculate the amount?  
[https://stackoverflow.com/a/30060286](https://stackoverflow.com/a/30060286) Testsuites contains the fields errors and failures and there is a separate error and failure element.

alexanderpann (changes requested):

build.gradle
arimer: Nice catch. I'd say catching for the xml elments names: `failure` and `error` (instead of the plural with `s`) should do the trick. What do you to think?

arimer:
@alexanderpann do you want  to have a second look?

alexanderpann (approved):
I'll approve but can you run it locally and make a few tests fail to check if the numbers line up?

arimer:
@alexanderpann of course.

PR #828, opened 2024-03-07
New Unit language: org.iets3.core.expr.typetags.physunits

alexanderpann:
Work in progress. There are more details about this language and some documentation in a separate document that we can't share yet because it contains sensitive information.

For the reviewers: please only look at the changes files, not the commits.

TODO:

- [x] make "with dimension" text configurable through the extension point
- [x] make "exponent comparator configurable through the extension point
- [x] write inline documentation using the mbeddr.doc language


HeikoBecker (commented):
Thank you for all the work you put into this @alexanderpann ! This looks great overall.

I still need to review the `physunits.behavior`. 

Here are some minor comments on the current code that I would like to resolve before merging:

- [x] Why does `org.iets3.core.base.mpl` get new dependencies if there is no other diff on the language? Can we revert the change?
- [x] What does this method in `EndCell` do:
image
- [x] TaggedTypeBehavior: Why can the `isSameAs` function now rely on the ordering of typetags? I did not see a change in the ordering of the tags anywhere else
- [x] `TestInterpreterForUnitConversionCyclic`: There is an empty line added there which is not necessary, can you revert it?
image
- [x] TestInterpreterForUnits: Can we remove this equalApprox function?
image
- [x] math Test: Why does this value change?
image
- [x] Can we remove the library `QuantityTests` in the playground now since it is superseeded by the tests?
- [x] This code change in the original units library seems unneccessary:
image
- [ ] In the documentation, 4 pictures are used to show the error cases, can we use the original models instead of screenshots? Otherwise I fear they will get outdated quickly
- [ ] the math interpreter now uses a `big-math` library. Is this safe to include in our project? Is this part of JBR?

HeikoBecker:
Continuation of review:

- [ ] GlobalUnitPrefixManager: I do not understand the logic in this function:
image
Why is the manager for the source used if both units are unscaled?
- [ ] What is the purpose of `UnitExpression.getReducedExpression`? Should it not also perform some additional reduction steps, e.g. here: 
image
- [ ] DimensionDivision: In the `getDenominator` method please return the actual denominator: 
image
- [ ] Fraction class: I think negate needs to return `(-this.numerator, this.denominator)`:
image


If these comments are addressed, we can merge the PR in my opinion.

HeikoBecker:
One more thing: Can you add the language to the changelog?

alexanderpann:
> Why does `org.iets3.core.base.mpl` get new dependencies if there is no other diff on the language? Can we revert the change?

This migration always shows up for me, so the issue must be on the maintenance/mps2022 branch already.

> What does this method in `EndCell` do:

I copied it from [EditorCell\_Empty](http://127.0.0.1:63320/node?ref=r%3A2af017c2-293f-4ebb-99f3-81e353b3d6e6%28jetbrains.mps.editor.runtime%29%2F8313721352726362705): It enables executing the right-side transformation action as well as setting the text for the hint cell (the text that you enter to the right of the end cell).

> math Test: Why does this value change?

We didn't work with infinite precision when interpreting the power expression. Yes, it is a a breaking change but let's fix it anyway.

HeikoBecker (approved):
Thank you for the fixes for my comments and the great work with this language. 

LGTM!

arimer:
Maybe we could have squashed those 228 commits locally 😓 before the merge

alexanderpann:
Into 1 commit? We can still do this but why lose all the information from the single commits?

arimer:
Judging form the description:
> please only look at the changes files, not the commits.

I assumed that squashing those commits (maybe not into 1, I agree) might have been an option

alexanderpann:
For documentation purposes in the Git history, those single commits are still helpful. For reviewing not really because I completely reworked parts of the new language multiple times and some intermediate fixes are not present in the final version of the PR.

Issue #823, opened 2024-03-05
Adapt iets3.opensource to use snapshot repo


psuzzi:
Adapt the iets3.opensource gradle script to use the snapshot repository. 

- all except maintenance and cust. branches goes to SNAPSHOT.
- for an example see: [MPS-extensions/build.gradle L:83](https://github.com/JetBrains/MPS-extensions/blob/0f22380f75b8f53d08026a575c5b188c964b7a66/build.gradle#L83)

psuzzi:
I received a report that the SNAPSHOT branch is incorrect in a specific case. See image. It needs to be fixed.

![image](https://github.com/IETS3/iets3.opensource/assets/352831/bf1ffb51-04a2-4ce4-b8a6-f2413604e2c3)

PR #857, opened 2024-03-19
fix customer branch matching for release

psuzzi:
solves #823 fixes customer branch

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,7 +82,7 @@ if (project.hasProperty('iets3OpenSourceVersion')) {
         currentBranch = GitBasedVersioning.gitBranch
 
         def buildNumber = System.env.BUILD_NUMBER.toInteger()
-        if (currentBranch.startsWith("maintenance/") || currentBranch.startsWith("datev/")) {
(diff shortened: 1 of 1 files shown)


arimer (approved):
I like the more specific check better. Thx.

arimer (changes requested):
see coments

build.gradle
```diff
@@ -82,7 +82,7 @@ if (project.hasProperty('iets3OpenSourceVersion')) {
         currentBranch = GitBasedVersioning.gitBranch
 
         def buildNumber = System.env.BUILD_NUMBER.toInteger()
-        if (currentBranch.startsWith("maintenance/") || currentBranch.startsWith("datev/")) {
+        if (currentBranch.startsWith("maintenance/") || currentBranch.startsWith("datev-loon-staging-") || currentBranch.startsWith("datev-steuer-staging-") ) {
```
arimer, line 85: I think this check needs to be a `currentBranch.startsWith("maintenance-")` as well (insetald of a `/`)

Otherwise maintenance branches are published as snapshots. See: 
grafik

psuzzi: Force-pushed a single change including `maintenance-`

arimer (approved):

PR #865, opened 2024-03-20
Fix type calculation of field setter

alexanderpann:
Example node where it didn't work: http://127.0.0.1:63320/node?ref=r%3Aa2594bdc-f128-474d-863a-401664a7ab8e%28test.in.expr.os.todo%40tests%29%2F5070313213711196305
There was a warning that the type never gets concrete.


HeikoBecker (approved):
Thank you for the fix. 

LGTM.

PR #869, opened 2024-03-21
Fix failing word tests

alexanderpann:
With the latest change in https://github.com/JetBrains/MPS-extensions/pull/805, words in the documentation are not empty anymore when they are created but also contain a character (space).


psuzzi (approved):
LGTM

Issue #848, opened 2024-03-14
Typecalculation of TupelType and EmptyType in conditional expressions changed between 2022.2 and 2022.3


svott7:
Hey,

why has the type calculation for conditional expressions changed?

Particularly, we observed a different behaviour between 2022.2 and 2022.3 when one conditional path is of type empty and the other one of type tuple.

### `org.iets3-opensource-zip-datev-steuer-staging-20222.2022.2.6213.b313950`
No error
![image](https://github.com/IETS3/iets3.opensource/assets/20836169/bdb9c5f5-455e-46da-bcba-0dacd47d8810)

### `org.iets3-opensource-zip-2022.3.6568.f3902e4`
Error
![image](https://github.com/IETS3/iets3.opensource/assets/20836169/09297f9c-d5d8-4893-b21d-0c2fb48b10b2)

svott7:
(`[empty, empty]` would fix the typesystem error but we don't like it as there was a shorter syntax before)

alexanderpann:
You get the same result in 2022.2 as in the second screenshot. There must be some commit on your branch that changes that behavior. The only relevant change that I can see was in TypeHelper https://github.com/IETS3/iets3.opensource/pull/678 but this was done in 2021.1.

svott7:
hmpf. I will check that. Thanks for your quick feedback.

svott7:
I guess it's related to the changes implemented in #678 . The pull request is not part of the old datev staging branches.

@mgronover Is the calculated type a bug or a featrue regarding our tuple vs. empty use case?

mgronover:
Because line  7 in the picture below is supported and tuples should behave the same way. I think it is a bug.

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/757fe12f-04c3-4551-aa84-9439d691a921)

The only error that should appear is the one in line 15.

svott7:
@mgronover thanks for the quick fix :)

PR #872, opened 2024-03-21
Bugfix/type system improvement

mgronover:
Issues: #505 and #848

EmptyType is deleted from the list of types of a JoinType, if there is at least one other type (in the another case the least common supertype is simply the EmptyType and not a JoinType at all).
The list of types of a JoinType is now sorted lexical. So there is no difference in the order of given of types within a  JoinType.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.base/org.iets3.core.base.mpl
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.base/org.iets3.core.base.mpl
@@ -20,7 +20,7 @@
     <dependency reexport="false">f0fff802-6d26-4d2e-b89d-391357265626(de.slisson.mps.hacks.editor)</dependency>
   </dependencies>
   <languageVersions>
-    <language slang="l:63e0e566-5131-447e-90e3-12ea330e1a00:com.mbeddr.mpsutil.blutil" version="1" />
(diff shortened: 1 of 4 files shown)


HeikoBecker (changes requested):
Somehow the version of com.mbeddr.mpsutil.blutil is downgraded for org.iets3.core.base. Can you please try reverting this change?

Other then that, the changes look good to me.

HeikoBecker (approved):
Thanks for the fix!

LGTM

arimer:
Just for my understanding, why do we need to [sort](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F8126947098204166783) the types list?

arimer:
@mgronover do you mind closing the issues in case your chages fix the bugs?

HeikoBecker:
> Just for my understanding, why do we need to [sort](http://127.0.0.1:63320/node?ref=r%3A5bf19129-2710-45a6-906e-9ee2d0977853%28org.iets3.core.expr.simpleTypes.plugin%29%2F8126947098204166783) the types list?

See the last comment by Klaus on https://github.com/IETS3/iets3.opensource/issues/505.

The idea was to allow join types to not be bound to the structure of the alt expression, i.e., both 
```
alt [
  x ==1 => 1
  x ==2 => true
  ]
 ```
 
 ```
alt [
 x == 2 => true
  x == 1 => 1
  ]
  ```
  
are typesystem-equivalent

PR #877, opened 2024-03-26
expr.simpleTypes: support infinite values in NumberType#containsZero

alexanderpann:


arimer:
The implementation itself looks fine. However I think your test is really hard to read and to maintain. In case of of the asserts fails, you need to look manually for the failing one, which is really cumbersome :/ 

What was the reason for using the light quotation instead of embedding the number type directly in the test?

arimer:
I've improved slightly the assert statements. We now have an error message which helps indicating which assert failed.

PR #681, opened 2023-07-05
Cleanup + model checker + linters (2022 branch)

alexanderpann:
- general cleanup 
-  the model check is now run as part of the build (no model check errors at all)
- support for some linters from [mps-qa](https://github.com/mbeddr/mps-qa/blob/master/code/languages/org.mpsqa.lint/README.md)
- small fixes (especially type system)

PR #887, opened 2024-04-03
Migrate or remove deprecated code

alexanderpann:
- The target branch is 2022.2 because I did the original cleanup in https://github.com/IETS3/iets3.opensource/pull/681 also on this branch.
- The deprecation warnings went down from about 353 to 145. For most of the remaining warnings, we would have to remove the deprecated concepts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## April 2024
+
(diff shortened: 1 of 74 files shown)


arimer:
Just out of curiosity, how did you migrated the deprecated code in 481cacbb9a3f6594a695bbc0392857b101495fb0? 
Was it via a migration script that was executed or via a manually written console script?

alexanderpann:
I've migrated everything manually with a console script.

alexanderpann:
1 and 2 : All of those are necessary standard implementations of `VisibleElementProvider#visibleContentsOfType` which is already deprecated. No need to add any other information here.

3\. Fixed.

4\. The old method invocation without prefix was deprecated and the prefix was set to null before.

arimer (approved):
All findings addresse! 

let's go 🚀

Issue #888, opened 2024-04-03
Using methods on list types having list types as a result causes compilation errors (javac).


mgronover:
The compilation of some of the following test function fails (root node: [functionType](http://127.0.0.1:63320/node?ref=r%3Af9a70877-66f3-464b-86cb-67a9d543be8c%28test.in.expr.os.functions%40tests%29%2F728017840040128456) ) :
![image](https://github.com/IETS3/iets3.opensource/assets/38247916/89ab7a34-7802-4627-90ac-f3a213b27a5d)

Here are the error messages
![image](https://github.com/IETS3/iets3.opensource/assets/38247916/7258369b-5450-427d-bbd1-98f222cc1c9f)

arimer:
Fixed by https://github.com/IETS3/iets3.opensource/pull/889

PR #889, opened 2024-04-04
Fixes compilation errors when using various methods on lists

mgronover:
addresses issue #888

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.base/generator/template/org.iets3.core.expr.genjava.base@generator.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.genjava.base/generator/template/org.iets3.core.expr.genjava.base@generator.mps
@@ -4438,151 +4438,117 @@
               <node concept="2OqwBi" id="4moR4VMUsCv" role="3cqZAk">
                 <node concept="1bVj0M" id="4moR4VMUsCw" role="2Oq$k0">
                   <node concept="3clFbS" id="4moR4VMUsCx" role="1bW5cS">
-                    <node concept="3J1_TO" id="4moR4VMUsCy" role="3cqZAp">
-                      <node concept="3clFbS" id="4moR4VMUsCz" role="1zxBo7">
-                        <node concept="3cpWs6" id="4moR4VMUsC$" role="3cqZAp">
-                          <node concept="1eOMI4" id="4moR4VMUsC_" role="3cqZAk">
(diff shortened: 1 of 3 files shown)


arimer:
Updated with latest changes from [maintenance/mps20213](https://github.com/IETS3/iets3.opensource/tree/maintenance/mps20213)

arimer (changes requested):
In general the changes look fine. Just some minor quenstions and proposals

1. why are you so sure about the type being a CollectionType [here](http://127.0.0.1:63320/node?ref=r%3A4243557f-1c7a-4d6b-953a-807576e4bee7%28org.iets3.core.expr.genjava.base%40generator%29%2F4707892895154038338)and [here](http://127.0.0.1:63320/node?ref=r%3A4243557f-1c7a-4d6b-953a-807576e4bee7%28org.iets3.core.expr.genjava.base%40generator%29%2F4707892895166599208)?
2. are we sure that every [collectionType.baseType that can be sorted](http://127.0.0.1:63320/node?ref=r%3A4fb4439d-f0bb-4e10-af94-6474b0ac1392%28org.iets3.core.expr.collections.constraints%29%2F3980268926917885790)is also implementing the interface [Comparable](http://127.0.0.1:63320/node?ref=6354ebe7-c22a-4a0f-ac54-50b52ab9b065%2Fjava%3Ajava.lang%28JDK%2F%29%2F~Comparable)? Because in your change here you expect exactly this
3. instead of using `node.expr.type:CollectionType.baseType` multiple times in the consequence of mapping rule, you could just define a `$VAR$ baseType` macro with the same impl. and use this var macro instead

mgronover:
Thanks again for the good review!

1. Both occurencies comes into account if the target of the node  is a [TailOp](http://127.0.0.1:63320/node?ref=r%3A61b1de80-490d-4fee-8e95-b956503290e9%28org.iets3.core.expr.collections.structure%29%2F393796118209799325) which extends ICollectionOp:
image
...
That ensures the `expr` of a `DotExpression` can only be a `collection`:
...
image

2. There is a filter used on the stream(), ensuring that OH (the ObjectHelper) is able to compare the objects of the collection:
image
The OH class does this by using a static Predicate combining Comparable and Comparator by OR operation.

3. I'll change that - thanks.

arimer:
upated branch again

arimer:
Remark on `Optional.orElse(null)` introduced in bbc021c0b4487a404b9d50496d170a334559d743. 

So I guess the idea is, instead of catching the exception and return null, we just  return null in case our filtering does not return any element, right? Meaning: At the end we have the same runtime behaviour

mgronover:
Exactly.

PR #891, opened 2024-04-04
Remove unnecessary super types of the tuple type

alexanderpann:
Due to changes in the MPS type system or our type calculation, the workaround in https://github.com/IETS3/iets3.opensource/pull/565 is not necessary anymore (see still working [tests](http://127.0.0.1:63320/node?ref=r%3A75f6a9e9-cf21-495f-b1aa-7fd992d0d7d0%28test.ts.expr.os.m1%40tests%29%2F1881564090920410105)). The workaround needs to be removed because for tuples with many elements the calculation is so slow that MPS freezes (time complexity: _n!_)


psuzzi (dismissed):
With check on size looks good to me.

arimer:
Why cant we make use of overriden implemenation of [allSuperTypes()](http://127.0.0.1:63320/node?ref=r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%29%2F5585772046592317725) in there? Might be easier to understand and easier to access all runtime supertypes.

PR #900, opened 2024-04-05
If there is only a single CC menu entry, select it for concepts that act as empty placeholders

alexanderpann:
These old steps to create a new member:
- create a record
- then you create an empty line
- then you need to create a record member if you don't know that you can just start typing (1)
    - Screenshot 2024-04-22 at 10 32 19
    - then you can set its name and data type
    
There is no information about (1), so people use the code completion menu to select the only applicable entry: a new member. With this PR, the only CC menu entry is automatically selected. The empty line is also inserted automatically when you create a new declaration.

The request was initially only for record declarations but I also added this feature to other concepts that represent empty placeholders like empty lines or other empty contents. It can still be relevant if those other concepts are filtered and only one subconcept is allowed at the end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - The unnecessary super types of the tuple type were removed and are not necessary anymore.
 
+### Fixed
+
+- Empty-like nodes now automatically select the first code completion entry if there is only one entry.
+
 ## March 2024
 
 ### Added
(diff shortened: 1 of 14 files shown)


arimer:
How about we add a small ticket that sums up this feature or add some more descriptions to this PR? Could be a nice documentation after this PR is merged.

alexanderpann:
I improved the description a bit.

psuzzi (approved):
LGTM

The [RecordDeclaration_Editor](http://127.0.0.1:63320/node?ref=r%3A1d6bd88a-7393-4b32-b0e6-2d8b3094776e%28org.iets3.core.expr.toplevel.editor%29%2F8811147530084018442) now allows to create a record and avoid the unnecessary user selection from the substitution menu for sub-items. In short, you can insert only a [RecordMember](http://127.0.0.1:63320/node?ref=r%3Ada65683e-ff6f-430d-ab68-32a77df72c93%28org.iets3.core.expr.toplevel.structure%29%2F8811147530084018361), which is automatically selected. 

image


The new behavior is implemented in the [SingleEntrySelector](http://127.0.0.1:63320/node?ref=r%3A8405f486-53b5-4fe6-af3e-7f68358bd631%28org.iets3.core.expr.base.editor%29%2F1016688335521801948)

This improvement is now used also for comments on Components, Instance, and Substructure.

PR #903, opened 2024-04-08
Implement exclusive number ranges

alexanderpann:
Example: number(1,3) and number[0,3)
I use the mathematical symbols for[ open and closed intervals](https://en.wikipedia.org/wiki/Interval_(mathematics)#Definitions_and_terminology).


psuzzi (approved):
LGTM

This commit adds exclusive number ranges. As the ranges are now inclusive by default, they can be transformed to exclusive by intention. Comparison logic is good and tests are extensive enough.

Issue #905, opened 2024-04-08
build: Misleading prompt during local build


arimer:
When building maintenance/mps20222 locally with a configured JDK 17 the following missleading information pops up.

```
> Configure project :
Local build detected, version will be 2022.2-SNAPSHOT
MPS requires Java 11 but current JVM uses 17, starting MPS will most probably fail!
```

This info is on wrong in multiple ways
- JDK 17 should be fine for building projects depend on >= MPS 2022.2
- iets3.os build is managing and downloading a compatible JBR version which is used during the build. So Only a java version for running the build.gradle file would be required

alexanderpann:
I have no clue where this message is supposed to be located and I don't see it locally or on the CI.

arimer:
How to reproduce:
1. Checkout maintenance/mps20222 brach
2. Use a local JDK installation in a version > 11
3. Run the `./gradlew`

Example of a local build
<img width="672" alt="grafik" src="https://github.com/IETS3/iets3.opensource/assets/16648247/6b83179f-f1c9-4703-b3eb-e7c530aad966">

arimer:
> I don't see it locally or on the CI.
Most probably our TC agens are still using JDK 11 as well as your local setup.

alexanderpann:
Yes, it comes from the mps-gradle-plugin: [https://github.com/mbeddr/mps-gradle-plugin/blob/a54f29f24d3155ecacd21d64de2b7a0f07ce660e/src/main/kotlin/de/itemis/mps/gradle/Common.kt#L69](https://github.com/mbeddr/mps-gradle-plugin/blob/a54f29f24d3155ecacd21d64de2b7a0f07ce660e/src/main/kotlin/de/itemis/mps/gradle/Common.kt#L69)

arimer:
fixed with #906

PR #906, opened 2024-04-09
Update the mps-gradle-plugin

alexanderpann:
- should fix https://github.com/IETS3/iets3.opensource/issues/905 through https://github.com/mbeddr/mps-gradle-plugin/commit/156775ab580c3d9269f052b460950a0f4d3e42c6

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,7 @@ buildscript {
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.7.288.4ea765f'
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.23.1.412.a54f29f'
     }
 }
 


build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
     }
     dependencies {
-        classpath 'de.itemis.mps:mps-gradle-plugin:1.7.288.4ea765f'
+        classpath 'de.itemis.mps:mps-gradle-plugin:1.23.1.412.a54f29f'
```
arimer, line 13: Is it still fine to use this version with MPS 2021.3 and higher?

arimer: Seem to work just fine.

arimer (approved):

build.gradle
alexanderpann: The 1.x branch is our active branch. Except for some specific features not available in older MPS versions, it can be used for all new MPS versions.

Issue #886, opened 2024-04-03
Vector Test doesn't handle enum


tomb50:
Vector tests cannot handle when the type of the function contains an enum

![image](https://github.com/IETS3/iets3.opensource/assets/2701339/f83ac601-088a-4e5c-8ef1-6bbf9816375b)

Tested on latest maintenance/mps20223

arimer:
Fixed by #907

PR #907, opened 2024-04-09
implement enum equality comparison

alexanderpann:
- The _equals_ method was missing for _EnumLiteral_.
- fixes https://github.com/IETS3/iets3.opensource/issues/886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - The unnecessary super types of the tuple type were removed and are not necessary anymore.
 
+### Fixed
+
+- Enum literals now can be compared for equality in the generator.
+
 ## March 2024
 
 ### Added
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tests/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tests/models/behavior.mps
@@ -9567,9 +9567,9 @@
     <property role="3GE5qa" value="vector" />
     <ref role="13h7C2" to="av4b:1bwJEEfQxC8" resolve="TestSubjectAdapter" />
     <node concept="13i0hz" id="1bwJEEeSLhl" role="13h7CS">
-      <property role="13i0iv" value="true" />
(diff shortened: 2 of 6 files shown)


arimer (commented):
In genal it looks fine. 

Some minor questions:

1. why do you delegate to the [super.equals()](http://127.0.0.1:63320/node?ref=r%3A93cd1fe8-b296-405c-a6e6-040c940ccfa1%28org.iets3.core.expr.toplevel.plugin%29%2F6127542811960197981) instead of returning "false"?
2. what is this [default concrete concept](http://127.0.0.1:63320/node?ref=r%3A17e537dd-82d4-485e-a86b-bac87852274c%28org.iets3.core.expr.tests.constraints%29%2F4196960000451511183) doing? Will it create a concerete concept, where an abstract concept would be created elsewise?
3. Why are you delegating to [EqualsHelper.equals()](http://127.0.0.1:63320/node?ref=r%3A7961970e-5737-42e2-b144-9bef3ad8d077%28org.iets3.core.expr.tests.behavior%29%2F6127542811960532028) here although we have no special case handling of EnumLiteral inside the EqualsHelper?
4. Maybe you know why In geneneral, EqualsHelper is performing additional checks (like for RecordValue) instead of moving that logic to the `equals()` method of RecordValue?

alexanderpann:
1.  I don't know if it makes a difference but I only wanted to overwrite the comparison logic for enum literals and nothing else.
2.  Yes.
3.  Yes, that is correct but we always use this class for comparison. I changed it to prevent future bugs.
4.  The changes in EqualsHelper were merged from a class called `ComparisonHelper` but I agree that the logic could be moved to the class RecordValue itself and checked by it is not the same implementation.

psuzzi (approved):
The PR adds the comparison logic for enum literals, and add tests. 

Based on prior review comments and answers, the PR looks good to me. 

As noted in the review, I agree it would make sense returning false in [EnumLiteral#equals()](http://127.0.0.1:63320/node?ref=r%3A93cd1fe8-b296-405c-a6e6-040c940ccfa1%28org.iets3.core.expr.toplevel.plugin%29%2F6127542811960197972) instead of using equals, but this does not prevent the approval..

psuzzi (approved):
LGTM

PR #912, opened 2024-04-12
DataTableLookup - Java generation column index is always zero

tomb50:
Fixed column index to reflect the correct position of the column, previously it was always zero due to indexing the wrong node.

Change screenshot
Screenshot from 2024-04-12 14-55-19


arimer:
@tomb50 thanks a lot for your contribution and the bugfix 👍🏼 

To make it available in older maintenance versions of iets3, I am porting your  to maintenance/mps20213. It will end up in your proposed target branch with the next cascading merge.

PR #918, opened 2024-04-16
Use BUILD_COUNTER instead of BUILD_NUMBER

sergej-koscejev:
BUILD_NUMBER is redefined by this build file on CI so we cannot run more than one Gradle step. BUILD_COUNTER is defined on TeamCity to be %build.counter% and is not touched by the build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,7 +81,7 @@ if (project.hasProperty('iets3OpenSourceVersion')) {
     if (ciBuild) {
         currentBranch = GitBasedVersioning.gitBranch
 
-        def buildNumber = System.env.BUILD_NUMBER.toInteger()
+        def buildNumber = System.env.BUILD_COUNTER.toInteger()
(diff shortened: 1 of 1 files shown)


psuzzi (approved):

build.gradle
```diff
@@ -81,7 +81,7 @@ if (project.hasProperty('iets3OpenSourceVersion')) {
     if (ciBuild) {
         currentBranch = GitBasedVersioning.gitBranch
 
-        def buildNumber = System.env.BUILD_NUMBER.toInteger()
+        def buildNumber = System.env.BUILD_COUNTER.toInteger()
```
arimer, line 84: Just out of curiosity, is the value of the old `BUILD_NUMBER` identical with the value of `BUILD_COUNTER` after that change?

sergej-koscejev: Yes, the old build number was initially set to %build.counter%.

Issue #932, opened 2024-04-23
Objects of RecordValue are incomplete in string representation


mgronover:
The method toString() does not return the keys.
Nor is it adaptable for customer (e.g. doing some kind of sort, filtering or string representation at all)

arimer:
Closed by #932

PR #933, opened 2024-04-23
RecordValue: toString can be modified an has default implementation r…

mgronover:
Addresses Issue #932 

Compatibility issues:
The string representation of RecordValues is also containing the keys now.


arimer (changes requested):
Thanks for your changes! I appriciate the backwards compatible change of the constructor of RecordValue 👍🏼 Changes look good, just some minor finding.

1. Was there a specific reason why you changed the toString() representation (besides adding the mapped.key information) by wrapping it into curly braces? I am not against it, just curious.
2. We should update the changelog before merging
3. I think the modules.xml change was not on puprose, do you mind reverting it?

arimer (approved):
Looks fine.

PR #939, opened 2024-04-29
expr.genjava.data:DataTableLookup: fixed java generation column index

arimer:
porting changes  from #912


alexanderpann (approved):
The change looks fine but the generator and interpreter implementation look quite different to me in general. Why is DataTable#defaultLookupColumn only supported by the interpreter?

alexanderpann (approved):
The change looks fine but the generator and interpreter implementation look quite different to me in general. Why is DataTable#defaultLookupColumn only supported by the interpreter?

arimer:
> Why is DataTable#defaultLookupColumn only supported by the interpreter?

Unfortunatelly I don't know why :/

PR #953, opened 2024-05-16
Remove the incorrect punctuation-right from NumberRangeSpec

alexanderpann:
The space between the range and, for example, the name of variables was missing + the side transformations for the precision didn't work.


psuzzi (approved):
Editor looks correct, and we can add the precision again.

Issue #956, opened 2024-05-21
Generator: Compilation error when an alternative or if expression uses list types


mgronover:
These four functions produces compilation errors due to invalid Java code (tested in MPS2021.3 and also in MPS2022.2):

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/9a671152-097b-4088-b98c-a91698489743)

The generated code of the second one is:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/54025d2e-acbb-4577-9e2f-9c05a07e591c)

And the error message that occurs when compiling:

![image](https://github.com/IETS3/iets3.opensource/assets/38247916/b9d4dab9-baa7-4163-bf69-e2bc86a29b32)

arimer:
Fixed with #957

PR #957, opened 2024-05-21
Generator: Added tests producing invalid java code (#956).

mgronover:
Issue: #956


arimer (changes requested):
Generator changes look fine for me. However, we should do one thing before mergen:

- [Here](http://127.0.0.1:63320/node?ref=r%3Af9a70877-66f3-464b-86cb-67a9d543be8c%28test.in.expr.os.functions%40tests%29%2F728017840040128456) you create a bunch of functions like: test_listFlatteningByIf, test_listChoiceByAlt, test_listChoiceByIf but only test of them inside the `checkAllFunctionsCanBeExecuted() ` check. Could you please add also the other ones to test?

arimer (approved):

PR #960, opened 2024-05-23
Bugfix/956 merge maintenance mps20222

### LINKED ISSUE #956: Generator: Compilation error when an alternative or if expression uses list types

mgronover:
Fixes #956 for customer.


mgronover:
Redo required. Looks fishy.

PR #970, opened 2024-06-07
Refactor/revert nix support

arimer:
Rerverting #652


kbirken (approved):
Looks good, reverts for all original commits in reverse order.

PR #975, opened 2024-06-11
Remove push-y from EndCell

alexanderpann:
We have to look carefully if it is indeed not needed or otherwise other editors will break. What happened is that the last entry of a collection was not selectable with the mouse because the end cell was visible in the editor and forced layout changes because of the` push-y` attribute.


HeikoBecker:
I have taken a look at IETS3.opensource editors for tables (`IMultiDecTab` uses `EndCell`), string validation (editor of `ValidateStringExprs`), record builder expressions (editor of `BuilderExpression`), and tagged expressions (editor of `TaggedExpression`). 

For these editors, I could not see any issues with the visualizations.

HeikoBecker (approved):

PR #980, opened 2024-06-17
Add editor hint support for trace 

### LINKED ISSUE #974: Add missing editor hint for conditinal editors of trance
Having dedicated editor hints which are used in conditional editors trageting tracing might improve overall runtime performance.

arimer:
Fixes #974

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tracing/models/editor.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.tracing/models/editor.mps
@@ -57,6 +57,13 @@
       <concept id="1080736578640" name="jetbrains.mps.lang.editor.structure.BaseEditorComponent" flags="ig" index="2wURMF">
         <child id="1080736633877" name="cellModel" index="2wV5jI" />
       </concept>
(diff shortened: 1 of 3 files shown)


alexanderpann (changes requested):
Did you mean `!isTraceHintAvailable()` in the is applicable block? Otherwise, you set the hint only when it is already set.

arimer:
Might, be. Let me check.

arimer:
Now you made me think if we do even want to check in the editor hint in the precondition of the action. I don't think we want to deactive the action if the hint is already set. 🤔

mgronover:
I found the following points:

1. The menu entries "show trace" and "stop trace" causes an influence on user settings in editor hints.

2. I would suggest that we do not need a second point to activate/deactivate the trace explorer "show trace"/"stop trace". The Toolbar of the TraceExplorer can be used to hide the decorated code:
image

I totally agree with the editor hint for the Tracing Editor. The reason is explained here: https://github.com/IETS3/iets3.opensource/pull/1005

So I would prefer using the approach of https://github.com/IETS3/iets3.opensource/pull/1005.

arimer:
drop in favor of #1005

PR #990, opened 2024-06-21
Force a empty default completion menu (including the base concepts menu) for IDotTarget

alexanderpann:
- follow up of #984 

This change is necessary so that the menu also works for derived interfaces/concepts. It didn't work for the following case: `concept C` implements `I`, `interface I implements IDotTarget`.


psuzzi (approved):
instead of using the `superconcepts menu`, we now `include default menu for BaseConcept for current node`. This shifts to directly including the default menu for the BaseConcept associated to the current node

PR #994, opened 2024-06-26
IDotTarget: Fix duplicate right side transformation entries

alexanderpann:
The parent side transformation entries were included for some reason which duplicates some entries. I really don't understand why this was necessary in the first place (and only for the right side). I feel like this was a workaround for an old (fixed) issue. For the new unit language, for example, we do it on purpose but here it doesn't make sense.


arimer (approved):

arimer:
Go for it.

PR #996, opened 2024-07-01
Implement an ncopies operation for tuples to ease initialization

alexanderpann:
The use case is long initializers with the same values such as initializing a tuple with zeros.                                                                                                      

Example: `assert [1, 2].nTimes(3) equals [1, 2, 1, 2, 1, 2]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,15 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## July 2024
+
+### Added
+
+- Tuple types now support a new method `nTimes` to ease initialization of long tuples.
+
 ## June 2024
 
-### Fixes
+### Fixed
 
 - An issue was fixed where the substitute menu entries of `IDotTarget` were duplicated because two default menus were created.
 - The same issue was fixed for the right transformation menu of `IDotTarget`.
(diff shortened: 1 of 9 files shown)


arimer (changes requested):
In general I like your idea, however there is still a limitation we should address before merging:

- at the moment the  **NCopiesOp** contains an [expression](http://127.0.0.1:63320/node?ref=r%3A66e07cb4-a4b0-4bf3-a36d-5e9ed1ff1bd3%28org.iets3.core.expr.base.structure%29%2F2403760773179493048) which allows users to define  arbitrary expressions. However it looks like only number literals are supported at the time being. 
- Creating a `[1].nTimes(3 + 1)` results in a warning and throws an exception during evaluation 
grafik
- It throws an `EvaluatorMissingException`
- in case we only support number literals we should we should add some constraints to only allow that kind of expression.

Here the stacktrace: 
[com.mbeddr.mpsutil.interpreter.rt.Evaluate.txt](https://github.com/user-attachments/files/16069447/com.mbeddr.mpsutil.interpreter.rt.Evaluate.txt)

alexanderpann:
I've changed the implementation to use an integer property but this issue bothers me and I saw it before. In your case, we try to invoke the interpreter from an inference rule but MPS can't calculate the type for 3+1 which is necessary to check the interpreter constraints. As a result, the interpreter rule for the plus expression is not applied. It works [here](http://127.0.0.1:63320/node?ref=r%3A6081d9c2-0e60-4ebc-a451-4434c25f85c8%28org.iets3.core.expr.repl.typesystem%29%2F5949014564335455787&project=org.iets3.opensource) for the Cell concept because it is a checking rule though. It also doesn't work for replacement rules.
We really need to find a workaround in the future to make that code work.

arimer:
I am just wondering if it would help to have a secod `when concrete` block that forces MPS to wait for the type calculation of the expression 🤔

alexanderpann:
Thanks, that works indeed.

arimer (approved):
Looks better 👍🏼

PR #997, opened 2024-07-02
Fix update inequality check

### LINKED ISSUE #701: Mutable: Update currency check in box update target uses wrong equality check
The following expression will fail although the old value (12) is equal to the expected value (12).

```
fun f/RM(b: box<number>) { 
  val x = b.val 
  b.update(12) 
  b.update from (12) to (x + 1) 
}
```

[Here](http://127.0.0.1:63320/node?ref=r%3Afbe318af-2afe-4858-8d66-a206b7f47bc8%28org.iets3.core.expr.mutable.interpreter.plugin%29%2F3487973603071619826) should be an not `equals` check instead of `!=`.

psuzzi:
Fixes #701 

Prevent the following expression from failing 

```
fun f/RM(b: box<number>) { 
  val x = b.val 
  b.update(12) 
  b.update from (12) to (x + 1) 
}
```


arimer (approved):
:seemsgood

PR #1005, opened 2024-07-10
 Enable override of Tracing editor

mgronover:
This Ticket addresses: https://github.com/JetBrains/MPS-extensions/issues/855

A ConditionalEditor by default has the hints `conditionalEditor` and `conditionalEditor_doNotUseThisHint`. These hints are enabled by default when the plugin becomes initialized.
If someone wants to overload a conditionalEditor it is not enough to just give it a higher priority. The priority just defines the order of the applied editors (a conditionalEditor uses [NextEditor]) see http://127.0.0.1:63320/node?ref=r%3A8d20232d-87e2-425b-b4d7-a9790e401b85%28de.slisson.mps.conditionalEditor.structure%29%2F2877762237607058140)" (s. screenshot on the mps-extension ticket above). So these can be cascaded somehow.

To enable overloading:
Generally an implementation of a conditional editor should also get an new defined editor hint if this editor should be able to be overloadable. The editor implementation that overloads that one must disable that and enable its own editor hint. 

For the Tracing editor this is done here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - Tuple types now support a new method `nTimes` to ease initialization of long tuples.
 
+### Fixed
+
+- Tracing editor can be exchanged by customers using editor hints.
+
 ## June 2024
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


mgronover:
See also the issue #974 and the according PR: #980

arimer (changes requested):
Thanks for the invetigation and the changes. 👍🏼 
- The implementation looks fine and I like the idea in setting the editor on startup of the plugin. 
- I cleaned up some deps and updated the build

However we should definitively document that change in the changelog before merging. Could you please update it?

arimer:
@mgronover we should also think about about updating our [NextEditor](http://127.0.0.1:63320/node?ref=r%3A8d20232d-87e2-425b-b4d7-a9790e401b85%28de.slisson.mps.conditionalEditor.structure%29%2F2877762237607058140) to allow removing not needed editor hints. Could you please create a ticket in mps-extensions and reference your PR there?

arimer (approved):
Let's go 🚀

PR #1012, opened 2024-07-16
cleanup dependencies of the unit language

alexanderpann:
There were some dependencies included like the tracing language which is not needed. This language is even excluded in one of our projects.


arimer (approved):
Dep. cleanup looks fine.

Issue #993, opened 2024-06-24
Make AsyncSolverTaskCreator more robust 


dbinkele:
Changes are restricted to 'org.iets3.analysis.base'. 
Basically this ticket aims to enable #1296 in iets3.core.

1. #1296 changes the way nodes can be treated. Before the FeatureModelConfiguration FMC was part of the model the 'Generator' was considering. After only a reference to FMC is passed to the Generator and then the necessary nodes are copied into the model under consideration. This implies that in the above language several lines of code which access the model need to be modified. It is not  always true that a node has now a model.
2. 'AsyncSolverTaskExecutor' should be more aligned with MPS and use the MPS-Threadpool for async Solver runs

dbinkele:
done #1018

PR #1018, opened 2024-07-25
Feature/make async solver task creator more robust #993

dbinkele:
solves #993

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/plugin.mps
@@ -498,6 +498,8 @@
       </concept>
       <concept id="7453996997717780434" name="jetbrains.mps.lang.smodel.structure.Node_GetSConceptOperation" flags="nn" index="2yIwOk" />
       <concept id="1173122760281" name="jetbrains.mps.lang.smodel.structure.Node_GetAncestorsOperation" flags="nn" index="z$bX8" />
+      <concept id="8866923313515890008" name="jetbrains.mps.lang.smodel.structure.AsNodeOperation" flags="nn" index="FGMqu" />
+      <concept id="1143226024141" name="jetbrains.mps.lang.smodel.structure.SModelType" flags="in" index="H_c77" />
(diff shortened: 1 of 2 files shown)


psuzzi (changes requested):
This PR increases the robustness of the AsyncSolverTask by making the execution more robust against empty models, removing unneded write access, removing semaphores in favor of properly implemented `CompletableFutures` . Also the `CancelableSupplier<T>` is now a `Callable<T>` , plus other minor improvements.

Overall the quality of the change is good, and everything makes sense to me. However, there is a short list of code cleanup fixes that I propose :
- [x]  The field [repository](http://127.0.0.1:63320/node?ref=r%3Ad354209e-0bea-497f-b905-d66f72900fa8%28org.iets3.analysis.base.plugin%29%2F3025971394814046910) in [SolverTaskRunner](http://127.0.0.1:63320/node?ref=r%3Ad354209e-0bea-497f-b905-d66f72900fa8%28org.iets3.analysis.base.plugin%29%2F3025971394814046905) is not used anymore, so we should remove it
- [ ]  If possible, remove the SRepository parameter in constructor [SolverTaskRunner(node st, SRepository r, node<> messageNode)](http://127.0.0.1:63320/node?ref=r%3Ad354209e-0bea-497f-b905-d66f72900fa8%28org.iets3.analysis.base.plugin%29%2F3025971394814046915) . I observed the only usage is at [this specific node](http://127.0.0.1:63320/node?ref=r%3Ad354209e-0bea-497f-b905-d66f72900fa8%28org.iets3.analysis.base.plugin%29%2F4802452575802883134) in SolverTaskCreation#createSolverTask. Since SolverTaskRunner is public, we might need to search for usages in client code

dbinkele:
@psuzzi I implemented the proposed changes. Even the breaking one. Theoretically a customer might use those classes. But   there are only 2 currently and they do not reach that deep into the framework. Only 'AsyncSolverTaskExecutor' should be publicly visible from the folders 'AsyncISolvable' and 'ISolvableMonitor'. Thus, anything else was made package-private. I am very very sure that these are not used, but also this changes prevent them from being use. We must be stricter here. Thanks for pointing this out.

psuzzi (approved):
The PR looks good to me. 

Having package-private internals is a very good idea, as this avoids improper usage, and simplifies our work when changing internals in the future.

PR #1020, opened 2024-07-30
expr.base: rewrite NCopies rule to avoid freezes for large numbers

alexanderpann:
For example _[[1, 2], [3, 4]].nTimes(20)_ completely froze MPS. I don't know if the order of the when concrete block was the issue or inefficient list operations.


arimer (dismissed):
Looks fine 👍🏼 
However I am really suprised that the light quotation was so much "slower" then the new implementation.

arimer (approved):
Thanks for the minor improvement 😘

Issue #778, opened 2024-02-14
Remove write access in async running if Solverfacade


dbinkele:
https://youtrack.jetbrains.com/issue/MPS-36395

Currently we generate a SolverTask by calling a generator facade from a checking rule.
It was necessary to use a 'write-access'. This is not reasonable as the  generator facade does only write on transient models.
The ticket https://youtrack.jetbrains.com/issue/MPS-36395 fixes this bug, such that we can remove the write access in 

http://127.0.0.1:63320/node?ref=r%3Ad354209e-0bea-497f-b905-d66f72900fa8%28org.iets3.analysis.base.plugin%29%2F4386560887386140284

arimer:
Could you please add the corresponding MPS version via label, where this bug appears?

alexanderpann:
According to the comment, https://github.com/IETS3/iets3.opensource/pull/1023 also fixed this issue.

Issue #1022, opened 2024-08-01
Make sure that ISolvable is run in low-prio thread


dbinkele:
Due to model access the creation of a solvertask is done async in a thread-pool by MPS.
But for actually running the ISOlvable which takes quite some time, we need a seperate Thread-Pool with low prio in order to have still a snappy UI

dbinkele:
#1092 done

PR #1023, opened 2024-08-02
Run SolverTask on low-prio thread (#1002)

dbinkele:
implements #1022


dbinkele:
@kbirken it seems with this PR also this one is fixed: #778

kbirken (approved):
Tested running solver check on IVAA in variability, against https://github.com/IETS3/iets3.core/pull/1336 -> works.

Cannot test the positive impact of these changes on performance right now. But we can do this in customer project after the integration chain.

kbirken (commented):
Just found a problem with iets3.core and checking Feature Models, has to be solved first.

kbirken (changes requested):
Problems occurring in checking of feature models in IETS3.core have to be solved first.

arimer:
@dbinkele , @kbirken why is this PR targeting the datev staging branch?

kbirken:
> @dbinkele , @kbirken why is this PR targeting the datev staging branch?

It probably targeted `maintenance/mps20222` before, and was retargeted to some arbitrary branch when the previous one was deleted by accident. Just guessing….

arimer:
I changed the target branch to the restored one: maintenance/mps20222

PR #1027, opened 2024-08-08
physunits: improve the code completion menu

alexanderpann:
The main goal was to make it possible to enter all units with the keyboard. The class _UnitNameReplacementHelper_ can also be used for generator purposes.


mgronover (approved):
LGTM.

Issue #1032, opened 2024-08-13
TraceExplorer should also show evaluations of Lambdas


mgronover:
Traced values of nodes that reside within lambda expressions should also be shown in Editors.

mgronover:
Though the main part is done by PR #1034, there is a further minor bug:
The focus is not set if the first selected TraceExplorer entry is below a lambda expression.
This could easily fixed, if getMappedNode() was public.
The TraceTab in mbeddr should also use that method.

mgronover:
The points above are solved with PR #1040.

PR #1034, opened 2024-08-13
Bugfix/1032 trace explorer show evaluations of lambda nodes

mgronover:
addresses issue: #1032 

Values of nodes in lambda expression are used decorate the corresponding editor nodes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## August 2024
+
+### Fixed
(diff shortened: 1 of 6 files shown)


alexanderpann (approved):
LGTM, one small remark:

` InterpreterUserObjectDecorator#getMappedNode` already assumes a `LambdaExpression`, so the method should probably only be used if the node contains an lambda expression or be renamed.

mgronover:
> LGTM, one small remark:
> 
> ` InterpreterUserObjectDecorator#getMappedNode` already assumes a `LambdaExpression`, so the method should probably only be used if the node contains an lambda expression or be renamed.

It should be callable with any recorded node and checks if that recorded node is a descendent of a lambda expression (i. e. has an anchester that is a lmabda expression). If that is true the lambda expression may contain a user object, a map associating nodes contained in a model to nodes evaluated and contained in that trace.
So it gets the node that is contained in the model for any node that is recorded (these might be the same if no lambda expression is used).

PR #1037, opened 2024-08-14
expr.base: correctly pass error types without checking every expression

alexanderpann:


HeikoBecker (approved):
LGTM. Thanks for the improvement!

PR #1040, opened 2024-08-19
TraceExplorer: Fix setting focus when using node mapper (#1032)

mgronover:
addresses issue: #1032


alexanderpann (approved):
Is the "+" in node.ancestor<concept = LambdaExpression, +>  on purpose? It didn't exist in the old version.

mgronover:
Yes, because the map contains the LambdaExpression itself:

**Within the Interpreter both lambda expressions are evaluated by creating LambdaValues:**
image

**The LambdaValue holds the mapping (created by copying or already existent in case of ShortLambda see below):**
image

**ShortLambda-Case:**
image

arimer:
Do we need that change in the maintenance/mps20222 branch as well?

arimer:
Well I think we still need it on maintenance/mps20222. Will  check if it is already there with #1051

PR #1048, opened 2024-08-28
physunits: make the unitless unit name customizable (followup)

alexanderpann:
The access to the repository was not safe for the generator, so we have to hard-code the name.


HeikoBecker (approved):
Thanks for the quick fix!

PR #1057, opened 2024-09-12
Use the typesystem devkit in typesystem models

sergej-koscejev:
Without the devkit corresponding checkpoints are not produced, making it impossible to override typesystem rules of the language.


sergej-koscejev:
Running the following console query shows that there are no more models left which do not use the devkit:
```
#models.ofAspect<typesystem>.where({~it => !new  ModelImports(it).getUsedDevKits().contains(module-reference/jetbrains.mps.devkit.aspect.typesystem/); })
```

alexanderpann (approved):
LGTM. Surprisingly, with the global query scope there are also no instances in our other platforms.

PR #1063, opened 2024-09-19
physunits: make it possible to replace ÷ automatically with /

alexanderpann:
Some users really don't like seeing, for example, 10 m²÷s instead of 10 m²/s. The other special characters seem to be fine. For generation purposes, they can be replaced by calling `UnitNameReplacementHelper#replaceUnitName`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 # September 2024
 
-### Fixes
+### Fixed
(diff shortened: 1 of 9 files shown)


arimer:
What is the reason for not providing that fix on 2022.2?

alexanderpann:
After merging the PR, we will backport the latest unit language PRs to 2022.2

arimer (approved):
I like the idea of making this change opt-in ❤️

PR #1069, opened 2024-09-25
UnitConversionUtil#createUnitMultiplication: use sequence instead of list

sergej-koscejev:
Using a list with a strict bound makes it impossible to pass a list of subconcepts to the method. This error is new in MPS 2023.2.


alexanderpann (approved):
LGTM, code is equal to the old one.

PR #1070, opened 2024-09-26
core.simpleTypes: add a migration for precision cant be derived from range

alexanderpann:


HeikoBecker (approved):
Thanks for adding the migration. LGTM.

PR #1074, opened 2024-09-26
core.expr: implement hex numbers

alexanderpann:
This is a more editor-heavy reimplementation of https://github.com/IETS3/iets3.opensource/pull/630 that is hopefully is not a breaking change anymore. There is only a small issue that when you create a hex value in the max field in the range specs the cursor jumps to the min field afterwards. I couldn't fix this issue but it is only a small inconvenience. See previous PR for a screenshot. The approach can be adapted for any string properties that are used for storing numbers.

Changes:
- Number literals, collection size specifications and number range specifications now support entering numbers in hexadecimal format. This feature can be deactived through PrimitiveTypeMapper#allowHexadecimalNumbers.
- Collection size and number range specifications now initialize with an infinite range by default. Setting the min and max field therefore is equal to setting them to infinity.


HeikoBecker (changes requested):
Thanks for the change. Overall, I really like the approach of using the attributes to provide different views on the underlying value of a number literal. I am leaving some comments here that I found during a first review.

**General**
- How will this feature work with migrations? If a migration changes the underlying value of a number literal do we now have to manually remember to update the attribute as well? Or will this work out-of-the-box?

**Functionality**
- overall approach works, I can enter hex values in ranges/number literals 
- Issue: Using Hex-values is not propagated into the typesystem, see e.g.
image
- Issue/Question: Can we make auto-completion also work when someone enters only the string `0x`?
image
- Question: What direction do we want to enforce by default? For me the values `0x0F` and `0xF0` lead to unintuitive results (`0x0F` is mapped to 15, where I would expect 240 and `0xF0` is mapped to 240, where I would expect 15). I am not sure whether we should here follow e.g. C-semantics or the base-10 notation direction (i.e. whether "LSB" is at front or back)

**Technical**
- Is there a possibility to automatically remove the hex-value? The current approach of having to reinitialize/ clean-up the model with the `removeHexValue` function seems very errorprone.
- Can we per default switch this feature off (`PTF.allowHexadecimalNumbers` returns `false`) to preserve the current behavior for now?

alexanderpann:
> How will this feature work with migrations?

When you set node.value you lose the hex value. I introduced a new method to also set the hex value. Example: `node.updateHexValueOrProperty(property/NumberLiteral : value/, "100")`

> Using Hex-values is not propagated 

Fixed.

> Can we make auto-completion also work when someone enters only the string

I implemented it so that 0x0 is automatically inserted but that also feels awkward. Should we generate a new hex value with invalid value "" instead? If the user enters 0x the editor cell is red and he has to enter a value.

> What direction do we want to enforce by default?

According to the internet, the evaluation from right to left as implemented here seems to be the standard. You can try it with different tools e.g. enter "0x0f in decimal" on Google.

> Is there a possibility to automatically remove the hex-value?

The easiest way to remove it, is to write `node.value = node.value`

> Can we per default switch this feature off

Done.

HeikoBecker (commented):
Thanks for the updates. The functionality now looks more like what I would have expected.

However, the substitute menu for `ISupportHexValues` does not yet work for me:
Peek 2024-10-09 14-47
The editing experience currently is a bit suboptimal. I can turn a literal into a hex literal by prepending the `0x` but not the other way round. Is there something we can do about this?

No other comments from my end. Thanks for making this feature also opt-in per default.

alexanderpann:
@HeikoBecker Thanks for noticing. I fixed both issues.

arimer:
@alexanderpann I think your merge went sideways. It looks like parts of the fixes introduced in commit a93b766a were removed/overriden. Could you please double check?

arimer (changes requested):
See last comment.

arimer (approved):
Fixed latest findings.
Now the feature is opt-in by default 👍🏼

PR #1075, opened 2024-09-26
add a replacement rule for homogeneous tuples and lists

alexanderpann:
Tuple types where all elements have the same type can now be used interchangeable with list types. Here is a concrete example where `nTimes` returns a tuple but the result is assigned to a list:

Screenshot 2024-09-26 at 17 54 57

On the generator level this level should also be fine as both tuples and lists use the class `PVector`. I just had to use `PVector` instead of `PVector<Object>` in some places.


mgronover (changes requested):
I do not understand why there is a requirement for the for loop (containing the foreach). For me it looks like everything is already added to the list by the marked line.

(If I comment out the loop the test cases pass.)

image

mgronover (approved):
As I understood now you have to infer the whole typesystem structure.

So everything looks fine.

Issue #1072, opened 2024-09-26
Why does Typedef has no RuntimeType?


arimer:
Figure out if it is intended or by accident

arimer:
fixed with #1077

PR #1077, opened 2024-10-01
Improve error messages and fix error reporting of cases where the common super type can't be calculated

alexanderpann:
- The goal was to make error messages more friendly, fix the spelling and make them more consistent.
- The common super type calculation itself is correct, but we still assumed that it returns null in the error case instead of a `JoinType` which is in many cases an indication that the calculation failed.
- For some cases, the interface `ITyped` is now implemented which makes the usability better because it has a cardinality of 0..1 + a checking rule. This is the main reason for the many changed files because I created automatic migrations for that. As a followup, we could also take a look at the interface `IOptionallyTyped` which could be implemented by many concepts.
- The PR also fixes https://github.com/IETS3/iets3.opensource/issues/1072 if I understand the ticket correctly.


arimer (changes requested):
Thanks for the effort of that big-ish pr :) 
Thankt a lot for the refactorings and simplifications. Findings below

### Current findings

#### Type Alignment
- What was the motivation of given a typedef a TS type? I also saw that it was missing, but my understanding was that, a TypeDef is an AbstractType (like a Java Class) and therefor does not need to have a type. When we go down that road we would also need maybe aligh with other `AbstractTypeDeclaration`
	- RecordDeclaration is not a type of it self, it is a `RecordType`
	- Does a TypeDef than also need to have a runtime type off `TypedefType`?


#### Error messages
- According to the caller, the [actual type could be null](http://127.0.0.1:63320/node?ref=r%3A80cf2246-750c-4158-9056-a619ebcf894c%28org.iets3.core.expr.base.typesystem%29%2F1152594682400803463)should your error message cover that case?


#### other stuff
- I think the [Assignable.getInitExpression()](http://127.0.0.1:63320/node?ref=r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%29%2F3723661587907930262)needs to be nullable. Because in most of the cases like for [Variable](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F8434481698271949638) and [Parameter](http://127.0.0.1:63320/node?ref=r%3Ab3786745-c763-4a49-a754-f84e15236f18%28org.iets3.components.core.structure%29%2F229512757699544992) those init expressions are optional. This might result in NPE in the [AssignableHelper](http://127.0.0.1:63320/node?ref=r%3A80cf2246-750c-4158-9056-a619ebcf894c%28org.iets3.core.expr.base.typesystem%29%2F3723661587906275020)

#### Tests
- what was the problem with `TestSuiteUsingCommented`? You deleted it

#### Typesystem and Supertypes
- during the refactoring of TyingHelper.calcCommonTypeCore() you changed the returned types. Initially `RuntimeErrorType` was returned in case the TS type is not `Type`. Now the behaviour is different, why?
grafik

- In typeof_ImmutableListLiteral, why cant we change that `PTF.computeSupertype(types, false, typechecker.getSubtypingManager())` to your newly introduced API of the TypingHelper [here](http://127.0.0.1:63320/node?ref=r%3A1fd78142-d7d8-42c9-9cbb-0609b1bc5311%28org.iets3.core.expr.collections.typesystem%29%2F99955696380893481)?

alexanderpann:
> What was the motivation of given a typedef a TS type?

It was already inconsistent, enum and record declarations already have a type. I just added it for convenience because the concept doesn't implement `IHasBaseType` and you need to query .originalType for the original type.

> Does a TypeDef than also need to have a runtime type off TypedefType

Do you mean an equivalent in the Java generator? We replace all instances with the original type.

> Error messages

Nice find, the cast is unnecessary. Example:
`val a : number = isSome(1) `

> I think the Assignable.getInitExpression()needs to be nullable

I tested it. The expression that you highlighted is null safe, the type operation as well and the type system seems to also ignore null.

> what was the problem with TestSuiteUsingCommented?

Every time we run migrations in the project, this invalid test suite with node test case statements pops up. Let's remove it before it happens again.

> Typesystem and Supertypes
The common type is already calculated in TypingHelper#computeRegularSupertype, we can just return this type.

Fixed the code in typeof_ImmutableListLiteral.

arimer (approved):
Go for it 🚀

PR #1078, opened 2024-10-03
expr.toplevel: Populate record literals with default values

alexanderpann:
Changes:

- The interpreter of the `success` expression was fixed.
- Custom Java exceptions have now a `equals` and `hashCode` implementation so that they can be compared in tests.
-  Let more types implement `ITypeSupportsDefaultValue`
- Record literals are automatically initialized based on their type.
- Record literals now show the referenced members of the record declaration.

Screenshot 2024-10-03 at 09 58 55

The syntax is `<expression>:<member name>` for technical reasons. An implementation with query lists was not possible, so I had to misuse the separator for it. A conditional editor would be too expensive as you would have to define it for the concept `Expression`.


psuzzi (approved):
The changes look good to me. 
There is no tests to verify equals and hashcode for MFIException, but that would not be really interesting, I guess.

PR #1080, opened 2024-10-08
physunits: make the kilogram check faster

alexanderpann:
Iterating over the full standard library is not a good idea.


arimer (approved):
In general I like the idea to use a dedicated node pointer 👍🏼 .

However, this also brings a lot of recstrictions on our `SIDerivedUnits` root node and the model. We must have a really close look on any changes in `SIDerivedUnits`, because deletion/move/cut-paste of those nodes might result in broken TS rules now.

alexanderpann:
As far as I remember, this approach is necessary to avoid a cyclic dependency between the language and the standard library. I know but we could use the [pseudoreferences](https://specificlanguages.com/posts/2022-02/02-pseudoreferences) from the migration language if we have more such use cases.

PR #1083, opened 2024-10-09
Introduce interpreter, trace & test infrastructure interfaces

enikao:
When using the KernelF interpreter in customer projects, we often run into issues where too much infrastructure is hard-coded, preventing us from adjusting it to our needs. Specifically, I'd like to introduce some kind of weak references at several places, so processing big models doesn't run out of memory because of too many interpreter trace objects in memory.

This PR includes:
* Changing all relevant behavior methods to `virtual`
* Introducing interfaces for interpreter and tracing-related classes, and use them throughout.

So far, the changes should be binary compatible to previous versions.
This would break once we changed the return types from e.g. `ValueAndTrace` to `IValueAndTrace`. The migration should be straight-forward: 1/ swapping the two types; 2/ changing field access of `ValueAndTrace.value` to `IValueAndTrace.getValue()`

We also identified a potential bug in `DefaultCoverageAnalyzer.isInInterestingContextEx()` (see comment).


mgronover:
@enikao is master really the correct target branch?

enikao:
> @enikao is master really the correct target branch?

I am targeting MPS 2023.2, and I think that's `master` branch? Or should I target it at an older branch and migrate forward?

mgronover:
The changes look good to me.
The bug you found is a real bug! Please fix it (exchange [_n_](http://127.0.0.1:63320/node?ref=f%3Adiff_diff_model_1%23r%3A83e946de-2a7f-4a4c-b3c9-4f671aa7f2db%28org.iets3.core.expr.base.behavior%40diff_model_1%29%2F8971071329112738046) by _it_  ). 

@AlexeiQ , @arimer for me it would be good to have these changes also in earlier version, what do you think?

enikao:
I just pushed lots of tests, and a fix for the bug.

enikao:
After merging in `master`, I have a dead reference in http://127.0.0.1:63320/node?ref=r%3A9741ad23-5b7e-4040-b070-51e24fd6bfee%28org.iets3.core.expr.repl.editor%29%2F6371013116349124293

image

arimer:
@enikao try a `./gradlew --refresh-dependencies` it might be the case the the platform has changed but you are still using an older cached version of it.

enikao:
> @enikao try a `./gradlew --refresh-dependencies` it might be the case the the platform has changed but you are still using an older cached version of it.

That helped, thanks!

arimer:
@enikao I triggered your build manually on TC, cause your branch name was not matching the build trigger config. Branches should start with: 
grafik

enikao:
Is there anything I can help with for review or merging?

enikao:
Superseded by #1098

PR #1091, opened 2024-10-22
physunits: check compatibility of units with same quantities

alexanderpann:
- For units to be compatible, they need to have the same quantity and an (implicit) conversion must exist between them.
- Quantity types are also not allowed in tagged types anymore. That doesn't make sense at all. I think we should also discuss the feature that you can use quantities as tagged types such as 1 force + 2 force in a followup ticket. It looks ridiculous to me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 - Error messages and checks were improved.
 - More error messages are now shown when the supertype can't be calculated.
 - The end cells of table rows of all tables where improved and now all support delete and insert actions.
+- Quantities are not allowed in tagged types anymore.
+- Units with same quantities are now checked if they can be (implicitely) converted between each other.
 
 ### Added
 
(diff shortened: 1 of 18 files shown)


HeikoBecker (changes requested):
Thank you for the changes. 

Some technical findings:

- Why is the error handling removed for `AbstractMinMaxExpression`?
image
- Why do we need to use the rhs of the binary operation in the compatibility check? 
image
- Do we need a migration for the part removing quantities from tagged types? I could see this cause problems in cases where this feature might have been used already. Is this actually migratable at all?

alexanderpann:
> Why is the error handling removed for AbstractMinMaxExpression?

I reverted the change, but the runtime error would come from `TypingHelper#calcCommonTypeCore` anyway and would be assigned in the next line.

> Why do we need to use the rhs of the binary operation in the compatibility check?

We only have binary operations on tags and I need to figure out if the left side can be converted to the right side.

> Do we need a migration?

There is no migration possible and the use case doesn't make sense. If there is really someone using them this way we can think about enabling them again.

HeikoBecker (approved):
Thank you for this improvement. LGTM.

PR #1092, opened 2024-10-22
Feature/make sure that i solvable is run in low prio thread

dbinkele:
solves #1022


kbirken:
Q regarding `ExecOnPooledThreadInWriteAccess`: It seems the method `AsyncSolverTaskExecutor.wrapInWriteAccess` does something similar as `ExecOnPooledThreadInWriteAccess`. Could one of these maybe be removed?

dbinkele:
> Q regarding `ExecOnPooledThreadInWriteAccess`: It seems the method `AsyncSolverTaskExecutor.wrapInWriteAccess` does something similar as `ExecOnPooledThreadInWriteAccess`. Could one of these maybe be removed?

Has been adressed, check out once more

kbirken (approved):
For me it looks good now. I did some small refactorings, esp. removed class `SolverTaskCreation` as it was used only locally and the `AbstractSolverTask` was stored in `SolverTaskWrapper` anyway. Pls. check if this is okay before merging.

PR #1093, opened 2024-10-23
Run the interpreter tests when no generator tests are available

alexanderpann:
It is so annoying to have red (not failing) tests because you are trying to execute generator tests but have not added the corresponding devkit to the model. Let's execute the interpreter tests instead for those cases. You can still see based on the name of the test which tests you are executing (generator tests have the word "generate" in their name)


alexanderpann:
I closed the PR because I misspelled the "bugfix" prefix.

PR #1096, opened 2024-10-24
New String type with constraints

alexanderpann:
This PR enables adding a regular expression to string types using the built-in regexp language + some extensions in [MPS-Extensions](https://github.com/JetBrains/MPS-extensions/pull/994) to enable most of the features of the [Java regex language]([docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/regex/Pattern.html)).

You can add the constraint manually when you use the string type or add the constraint dynamically when calculation a type. When one string type has a constant text constraint it can be matched against the constraint of another string type (see implementation of string literal type).

One implementation note: a removed the support for parenthesis in those regexes and added groups instead. In practice, parentheses are non-capturing groups which is super confusing when you try to refer to them by index such as \1 or \2 and it doesn't work. You can also create an external regexp as a root node and refer to it from your constraint. This is the first time we are mixing a base language-related language into KernelF.

Some examples from the tests:
Screenshot 2024-10-25 at 01 20 53

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.adt/models/typesystem.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.adt/models/typesystem.mps
@@ -1688,18 +1688,18 @@
     <property role="TrG5h" value="typeof_LocDotTarget" />
     <property role="3GE5qa" value="dot" />
     <node concept="3clFbS" id="7aipPVpFzey" role="18ibNy">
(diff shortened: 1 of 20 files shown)


HeikoBecker (approved):
Thank you for adding this cool new feature. LGTM.

PR #1098, opened 2024-10-30
Introduce interpreter, trace & test infrastructure interfaces

enikao:
When using the KernelF interpreter in customer projects, we often run into issues where too much infrastructure is hard-coded, preventing us from adjusting it to our needs. Specifically, I'd like to introduce some kind of weak references at several places, so processing big models doesn't run out of memory because of too many interpreter trace objects in memory.

This PR includes:
* Changing all relevant behavior methods to `virtual`
* Introducing interfaces for interpreter and tracing-related classes, and use them throughout.

So far, the changes should be binary compatible to previous versions.
This would break once we changed the return types from e.g. `ValueAndTrace` to `IValueAndTrace`. The migration should be straight-forward: 1/ swapping the two types; 2/ changing field access of `ValueAndTrace.value` to `IValueAndTrace.getValue()`

Supersedes #1083 (bad branch name)


enikao:
With this PR I chose the careful approach to split off breaking changes. Given the time it takes to process it, would it be better to include them? We have serious memory performance issues at a client, and the requested changes are a prerequisite to tackle them.

mgronover:
@enikao what do you mean with:
> With this PR I chose the careful approach to split off breaking changes.

Currently I do not see any breaking changes.  There is no compulsion to exchange ValueAndTrace by the interface IValueAndTrace right now.

mgronover:
@alexanderpann it would be good to have a second review.

enikao:
> Currently I do not see any breaking changes. There is no compulsion to exchange ValueAndTrace by the interface IValueAndTrace right now.

Yes, that's what I mean. I currently didn't change the return type of e.g. `IETS3ExprEvalHelper.evaluateWithTrace()` from `ValueAndTrace` to `IValueAndTrace` to stay compatible with existing usages.

This means that currently any alternative implementation of `IValueAndTrace` must effectively subclass `ValueAndTrace`.
So the proper solution would be to only use `IValueAndTrace` on all APIs.
However, that breaks users of these APIs, as they have to replace e.g. `ValueAndTrace.value` (direct field access) with  `IValueAndTrace.getValue()`.

alexanderpann (approved):
LGTM. Thanks also for providing new tests!

PR #1100, opened 2024-10-31
physunits: improve implicit conversions

alexanderpann:
- Type calculation for implicit conversions in the physunit language was improved.
- Implicit conversions can now be deactivated conditionally in the physunit language. You can for example, only activate them for binary operations, so that normal assignments like val length: number<m> = 1 km still leads to errors
- The `noConvert` expressions in the physunit language doesn't strip the unit anymore. Use the `stripUnit` expression for that. 

I also have to fix the type calculation for div expressions since they contained unnecessary trailing zeros which leads to ugly number ranges with those implicit conversions. The changes for the type calculation were also done for the `convertTo` expression.

Implementation details: The reason why we need to widen the precision in the tagged expression is that the interpreter needs to work with BigDecimal and not BigInteger in our calculations and with this change we enforce floating point calculation. The previous real type was not necessary, changing the precision is enough. I also try to use the precision and range from the original expression if possible. The conversions itself can overwrite this, so we do this only for implicit generated conversions rule of prefixes.
 
Examples:

Screenshot 2024-10-31 at 08 27 26

Screenshot 2024-10-31 at 08 19 16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 ### Fixed
 
+- Type calculation for implicit conversions in the physunit language was improved.
+- Number types have trailing zeros stripped in their ranges after type calculations for divisions.
 - The compatibility check of quantities of the physical unit language was improved.
 - The interpreter of the `success` expression was fixed.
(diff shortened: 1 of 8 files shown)


arimer (approved):
Changes look from the technical point of view all fine 👍🏼

PR #1103, opened 2024-11-05
datetime.runtime: Fixed compareTo in ArbitraryDateRangeValue

arimer:
The [endComparison](http://127.0.0.1:63320/node?ref=r%3Aa9ac3767-b241-4aa4-a973-d04bb5ce184c%28org.iets3.core.expr.datetime.runtime%29%2F7492452870508394462) was using wrong value. 

In addition I've also bumped up the MPS version to 2022.2.**4**

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -57,7 +57,7 @@ def major = "2022"
 def minor = "2"
 
 // Dependency versions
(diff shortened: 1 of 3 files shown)


mgronover (approved):
Well done

PR #913, opened 2024-04-12
correction and newly added SI unit libs

yonghui-git:
**Minor corrections in the existing unit libs (SIUnits & SIUnitsDerivedAndScaled) and a new lib (SIUnitsDerivedAndScaledWithSpecialNames)**

The newly added unit library "" is based on the units from the International System of Units on wikipedia https://en.wikipedia.org/wiki/International_System_of_Units.

Hopefully, all units mentioned by this wikipedia page are covered by these three libraries, i.e., **SIUnits**, **SIUnitsDerivedAndScaled** and **SIUnitsDerivedAndScaledWithSpecialNames**.

PR #1107, opened 2024-11-05
Updated SI units

arimer:
Porting changes from https://github.com/IETS3/iets3.opensource/pull/913


alexanderpann:
All of this exists in the new unit languages and even more.

Issue #1108, opened 2024-11-06
TestCase evaluation by an Interpreter shouldn't require the generation and build of the model containing that TestCase.


mgronover:
Currently we generate and build the model to evaluate the containing TestCases. 
This is done in the way the JUnit-Test-Framwork expects it and is also working for the CI (AFAIK).

For DSL users implementing new TestCases this approach slows them down, because it requires to build the TestCases.

A new Run Configuration, having an Executor that calls the Interpreter to evaluate the TestCase, get the result and update the MPS-JUnit-ToolWindow just like it was done before, would solve this situation.

So currently we are focusing on the IDE usage here.

What is to be expected:
 - [ ] the run configuration should check and warn if the interpreter was not build
 - [ ] the interpreter tests should always be executed in the same process
 - [ ] the configuration panel only suitable options (no more than required!)

PR #1109, opened 2024-11-07
Create a separate run configuration for JUnit interpreter tests

mgronover:
addresses issue: #1108

## Why do we need this

This PR adds a new "run configuration" to evaluate tests by the Interpreter without the requirement to generate and build the tests. So we assume a speed up for all users when developing tests in iets3 hereby. 

image

## Development issues and explanations:
The corner stone for the development was the already existing run configuration for unit test, that required a generation and compilation of the code. Due to the fact, that this code is not very generic, many parts of it has to be reimplemented. This implies that in future versions, we may be have to modify some parts.

**Of cause this run configuration only works for nodes, that can be evaluated by the interpreter!**

The old JUnit run configuration is able to execute a superset of that tests (because it will generate and compile the models and so be able to also execute test for the typesystem, node tests, editor tests, generator tests....).

The test result by this run configuration are of cause the same as with the old JUnit run configuration. Except, that we can omit the jvm stack traces as long as no exception is thrown. In the old JUnit run configuration every expected value of a test that does not match with its actual value will throw an assertion exception and therefore cause a stack trace, what is not very helpful in this case.

**Concerning the CI integration we should better do this on another ticket!**


arimer:
Does it make sense to create a short video/gif/screen series to document the usage of that new feature?

mgronover:
Unfortunately I had to duplicate and modify some code and I would appreciate if several of you would do the review.

(I now understand why JetBrains is thinking about a new JUnit Framework adaption, as far as i understood @sergej-koscejev correctly)

alexanderpann:
Just a remark but in newer MPS versions and in our current version there were multiple changes to the JUnit Framework done.

sergej-koscejev:
Yes, I was referring to those changes (that have already been done).

mgronover:
Well it doesn't bother me if I have to fix the things for the cascading PRs.

arimer:
**UI Part (1)**
- I think we should package and ship the "UI" integration part as a new plugin. Therefor we should add it to the [org.iets3.opensource](http://127.0.0.1:63320/node?ref=r%3Ac3d6ae0c-8b10-477f-a3e9-5dc8700ceb13%28org.iets3.opensource.build.build%29%2F6354991446090808661) build script and use the right versions etc. 
- This allows consumers of iets3 to actually make use of this UI features

*Command Line Part (2)*
- It looks like we will habe a custom solution for the command line integration anyway, meaning: we will have a custom MPS build script with the solution containing the  "Main" class
- this new build script will also probably need to configure the test solutions holding the models/root nodes suited for interpreter test execution


How about we split (1) and (2) into dedicated PRs?

mgronover:
Yes but the names currently do not express what the content is about. This should also be fixed.

arimer:
Missing pieces

#### Make the Execute Code from Solution smarter
- [ ] load models from sources (not from jars) without the need to build the solution first
- [ ] extend therefore the [BuildSolutionRunnerAspect](http://127.0.0.1:63320/node?ref=r%3A0f2b4a26-93a1-4327-97ef-ca91b7a4cf5e%28jetbrains.mps.build.mps.runner.structure%29%2F4173297143638950526) and its generator to allow first point
- [ ] Configure [org.iets3.opensource.interpreter.testExecutor](http://127.0.0.1:63320/node?ref=r%3Ac3d6ae0c-8b10-477f-a3e9-5dc8700ceb13%28org.iets3.opensource.build.build%29%2F9022445546689294517) with all models that contain intepreter tests

### Rethink our test soutions and test strategy (optional)
- [ ] SplitUp up test solutions for interpreter/generator tests only 
- [ ] re-configure the build with the new test feature

arimer:
Is this PR ready for a final review? (IDE part only)

mgronover:
Nope, we should filter the test cases more strictly for interpreter evaluation.

arimer:
>Nope, we should filter the test cases more strictly for interpreter evaluation.

You can check for roots of test suites I guess? I saw some in `in.tests `and `ts.tests` and some sandboxes. So just moving stuff from one to the other or vice versa would be enough I guess?

arimer (changes requested):
The changes look good! 👍🏼 
However I still have some findings/remarks: 

**Minor cleanups from my side**

- I cleaned up dependencies and rebuild the build models
- renamed your run configuration solution
- updated the idea plugin name accordingly
- deleted the solution that does not belong to the project
- simplified code by using `jetbrains.mps.lang.smodel` API

**Technical Findings/Questions**

1. [junitParams.getTests(mpsProject)](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F6062668769034010908)delegates to your custom [collect()](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F1467430125247248562)interface method. However, I could not find any concrete impl. of that method. Which code is actually executed when collect is called?
2. in [WrapperFilter.filterByConcept()](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F4910986300292768337) you explicitly check for `isInstanceOfConcept`, which would also include nodes, where the concept is a sub concept of the given concept. Is this on purpose? 
3. I don't think this [comment](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F4023783813281576099) is true any longer, can we remove it?

4. You are still canceling the test execution process in case the `Compile in MPS` flag is un-checked. [Here](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F5740494279801064516) and [here](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F5740494279801242975) . Why? I still think it is a valid use to assume that our kernelF tests can live in a model/solution that is not generated and compiled into java. Is this a restriction by the junit/runner framework? However it seems this assumption is deeply part of our kernelF language: grafik

**Functional Findings**
- after executing all interpreter tests successfully via the dynamically created configuration, the icon still indicates "skipped" tests instead of a green checkmark: grafik Can we do anything about this?

mgronover:
ad 1)
It is still the same as in JUnit run configuration [here](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.JunitInterpreterTest.plugin%29%2F6062668769034855736). This assumes the [compileInMPS](http://127.0.0.1:63320/node?ref=r%3A63a75970-913d-4a7b-99e2-8ca72ff6f509%28jetbrains.mps.baseLanguage.unitTest.execution.client%29%2F8617997997504895386) flag to be set, but currently that should not be an issue as long as we require the old JUnit run configuration to be executed.

ad 2)
Yes, it is. We wanna have TestSuite and TestCase which both implement ITestItemContainer, that we are using as "filter" in the current implementation. Generally we could create an interface just for tagging concepts implementing something that can be evaluated by the interpreter and play a role of a test.

ad 3)
I totally agree.

ad 4)
see ad 1)

mgronover:
> Functional Findings
> 
> after executing all interpreter tests successfully via the dynamically created configuration, the icon still indicates "skipped" tests instead of a green checkmark: grafik
> Can we do anything about this?

This is now fixed.

arimer (approved):
Thanks for the remaining fixes ❤️ 
Works like a charm
grafik

PR #1111, opened 2024-11-12
core.expr.base: show hex attribute better in diff view

alexanderpann:
Screenshot 2024-11-12 at 12 06 23


alexanderpann:
There are two different editor components in the VCS language and there could be more of them in the future. That's why I only check for the package name. Very unlikely that the package name changes but this way we also avoid a dependency on the VCS plugin.

mgronover (dismissed):
LGTM.

mgronover (approved):
Fine!

PR #1112, opened 2024-11-12
Updating HasValue_OP to not propogate optional type to dot expression

tomb50:
Updating HasValue_OP to not propogate optional type to dot expression

When HasValue_OP was being used on a expression tyoe that was an optional, the resulting type of the dot expression also became an optional -this can be seen in the below screenshot
Screenshot from 2024-11-12 12-27-33

This made it difficult to work with since you had provide x.hashValue :? false etc everywhere

By checking the DotExpression Type system aspect i could see reference to the dot expression method neverMakeOption, which default to false but for this case HasValue presumably should be true.

Screenshot from 2024-11-12 13-43-37

I added sometests within the existing test suite

Screenshot from 2024-11-12 13-42-08


mgronover (approved):
LGTM.

PR #1119, opened 2024-11-13
removed math.interpreter lib folder as model folder

enikao:
For solution `org.iets3.core.expr.math.interpreter`, both the `models` and the `lib` folder were registered as model roots.
However, `lib` doesn't contain any models, which lead to model loading warnings on every MPS startup.


alexanderpann (approved):
Thanks for the fix!

PR #1121, opened 2024-11-14
Make it possible to customize where hex values are enabled

alexanderpann:
With PrimitiveTypeMapper#filterHexadecimalSupportingNodes you can now enable/disable them for specific nodes.


arimer:
Are those changes really only relevant for master?

alexanderpann:
A customer wants to active hex values but needs more customization options. The feature itself was introduced on maintenance/mps20223 in case we also need it there.

arimer (approved):
Thanks for porting it back to 20223 👍🏼

PR #1122, opened 2024-11-14
Feature/improve value inspector mps20222

mgronover:
cherry picked commits of #558 (because target branch of that is not supported anymore).


arimer (approved):
Changes look good 👍🏼 

Thanks for the cleanup and for porting it to the newer version ❤️

PR #1123, opened 2024-11-14
Migrate to 2024.1.1

alexanderpann:
There is only one annoying issue in MPS: [youtrack.jetbrains.com/issue/MPS-36888/java.lang.UnsupportedOperationException-not-implemented](https://youtrack.jetbrains.com/issue/MPS-36888/java.lang.UnsupportedOperationException-not-implemented)

Model checking was never fully enabled. I disabled it so that it doesn't throw errors.


build.gradle
```diff
@@ -280,9 +280,11 @@ modelcheck {
 
 
 afterEvaluate {  project ->
+    /*
```
arimer, line 283: How about we configure checkmodels properly instead of removing it? 
Although tbh. I do not know what we be a suitable candidate for checking...

build.gradle
```diff
@@ -53,14 +53,14 @@ if (project.hasProperty("forceCI")) {
 
 def forceLocal = project.hasProperty("forceLocalDependencies")
 
-def major = "2023"
-def minor = "2"
+def major = "2024"
```
arimer, line 56: In addition to the new MPS version, don't we need to update the JBR as well for the CL build?

build.gradle
arimer: Are there now more/new errors that do not appear on earlier versions ?

alexanderpann: We introduced some errors in earlier versions. No new errors I believe. They are still not that easy to fix.

arimer:
@alexanderpann looks like your mentined YT issue is closed. 👍🏼

arimer:
Why are all groups set to `update in background: true`? 
grafik

alexanderpann:
The updating of groups and actions usually takes place on the EDT/UI thread. If this takes a long time, the UI is frozen. With this MPS version, you even get a fatal error if the update takes too long.

arimer:
In TestAsyncFromSingleThread you have added a new test. How do you ensure that this test does not remove anything that is required by any other test in that model?

```
test cleanup { 
  execute command in EDT with project.getRepository() { 
    model.roots(DummyISolvable).forEach({it => model/.removeRootNode(it); }); 
  } 
}
```

alexanderpann:
It is executed after the other test. The cleanup in this model still is not perfect by the way but better than before. The problem is that in the before and after test methods I can't access the model, so I can't do any proper cleanup. If you use a try-catch-finally block you mess up the asserts as far as I remember.

alexanderpann:
> How do you ensure that this test
The test is always executed after the other test. We can't use the before and after test blocks because they are static methods that, for example, also don't have access to the model.

arimer (dismissed):
Chapeaux 🎩 

Short summary of the changes: 
- bunch of dependency changes + cleanups
- cleanups (deletion, renaming, tests)
- internal API changes (access von repository in tests)
- Actions need to run in BGT

The fix of the deprecated API in the comment is optional from my point of view and can be done "on the go". 

Fine work my man 💪🏼

build.gradle
arimer: ok, lets talk offline what could be a fitting model for model checking

arimer:
Oh, and maybe update the readme to anounce that master is on 2024.1

arimer:
I've created a ticket ot not forget about it: https://github.com/IETS3/iets3.opensource/issues/1131

arimer (approved):

PR #1136, opened 2024-11-20
Reliably remove NixType from ListType

enikao:
This used to work in previous versions of iets3.opensource (at least 2023.2.7319.4e501a1), but not currently.

I added all these tests (and an test language with a nix type).
The goal is to remove `NixType` from _xxx_ in `ListType<xxx>`.

```
Lists with nix                                                                        
------------------------------                                                        
                                                                                      
val oneNix = <check list(nix) has type list<nix type>[1|1]>                           
val twoNix = <check list(nix, nix) has type list<nix type>[2|2]>                      
val stringFirst = <check list("hello", nix) has type list<string>[2|2]>               
val nixFirst = <check list(nix, "hello") has type list<string>[2|2]>                  
val stringFirstThree = <check list("abc", nix, "def") has type list<string>[3|3]>     
val nixFirstThree = <check list(nix, "abc", nix) has type list<string>[3|3]>          
val stringFirstFour = <check list("abc", nix, "def", nix) has type list<string>[4|4]> 
val nixFirstFour = <check list(nix, "abc", nix, "def") has type list<string>[4|4]>    
```

All of them work, except the last one. There we get `list<join<nix type, nix type, string>>[4|4]`.
Any idea why?

### DIFF
--- a/code/languages/org.iets3.opensource/.mps/modules.xml
+++ b/code/languages/org.iets3.opensource/.mps/modules.xml
@@ -81,6 +81,7 @@
       <modulePath path="$PROJECT_DIR$/languages/org.iets3.req.plugin/org.iets3.req.plugin.msd" folder="req" />
       <modulePath path="$PROJECT_DIR$/languages/test.iest3.component.attribute/test.iest3.component.attribute.mpl" folder="comp" />
       <modulePath path="$PROJECT_DIR$/languages/test.org.iets3.analysis.base.solvable/test.org.iets3.analysis.base.solvable.mpl" folder="analysis.test" />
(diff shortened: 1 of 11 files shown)


enikao:
@alexanderpann I think it's related to #1096

alexanderpann:
@enikao My guess would be rather https://github.com/IETS3/iets3.opensource/pull/1077. Seems like the super type calculation doesn't handle this case correctly anymore.

alexanderpann (approved):
LGTM. I think we still have an issue with some string comparisons, for example in messages that is unrelated to this PR: http://127.0.0.1:63320/node?ref=r%3Acc6f834a-e980-4402-b501-57396f6330a8%28messages%40tests%29%2F4026566441522104723
We really need to get model checking working for this repo.

PR #1138, opened 2024-11-21
physunits: Improvements to units of digital information

alexanderpann:
- The physical units B and b were renamed to byte and bit to avoid confusion.
- Breaking change: The units of digital information were split into 3 different libraries: `UnitsOfInformationIEC`, `UnitsOfInformationJEDEC`, `UnitsOfInformationMetric`. They are still considered part of the derived units. If you want to use them, you therefore have to include them in your scope method.
- Physical units now also support metric scaling for only the positive and negative prefixes.
- I also noticed that many derived units can also have metric prefixes. The information for the last and this item are AI checked, so we have to trust the information that we got.
-  Scaling can also be overwritten for units by overwritten `IUnitLangConfig#getOverwrittenScaling` for the extension point `PhysUnitLangConfig`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,12 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Changed
 
 - It is now possible to customize where hex values are enabled. With PrimitiveTypeMapper#filterHexadecimalSupportingNodes you can now enable/disable them for specific nodes.
-- The units B and b were renamed to Byte and bit to avoid confusion.
+- The physical units B and b were renamed to byte and bit to avoid confusion.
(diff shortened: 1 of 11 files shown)


HeikoBecker (changes requested):
Thanks for the changes! 

I have done a functional review first and noticed that there is no error/warning shown if I import both/all libraries of digital information, which leads to confusing UX:
Menü_003

I think there should be a warning shown, in case two units of the same name are in scope on the tag used for the `val` assignment.

alexanderpann:
I made the error message a bit clearer, but some people might want to use multiple versions. You can add a `ChunkDependencyConstraint` to a chunk where you define `ChunkDependencyConstraint#canImport` to disallow multiple such imports. Other than that it should be handled through the scoping. I don't want to add unit-specific checks to the generic chunks in KernelF.

HeikoBecker (approved):
Thanks for the update! This is great.

Second functional review: Check for incompatibility works now.

Technical Review:
No findings.

PR #1143, opened 2024-11-26
Adapt icons to new UI design

alexanderpann:
Apparently there is not much to do. I didn't change the concept icons and also not the icons of the actions since we don't have good replacements. There is also not a single tool icon to change 🤷🏻  If you see anything else, let me know.


arimer (approved):
Looks fine. 👍🏼 

Just keep an eye on those actions and the background update change when you merge them to master. I think they all are running in BGT on master atm.

sergej-koscejev:
FYI @alexanderpann @arimer the change done in https://github.com/IETS3/iets3.opensource/pull/1143/commits/554610fe015eec4c8a6ceb7a96bb592d491489c2 turned out to be an important bugfix for a customer. It would be good to mention such changes in the release notes in the future (and perhaps create a separate PR for them).

PR #1146, opened 2024-11-27
expr.base: convert negative decimal values correctly to hexadecimal values

alexanderpann:
Example: -255 was converted to 0x-ff before.


mgronover (approved):
LGTM

PR #1149, opened 2024-11-28
physunits: allow mixing quantities with different prefixes through IUnitLangConfig#implicitConversionIsEnabledAt

alexanderpann:
This workaround is necessary for cases where projects can't switch to implicit conversions.


wsafonov:
@alexanderpann I have improved the signatures in `UnitConversionUtil` to avoid type errors. But can you please have a look at `QuantityType.createDefaultVarExpr()` - something is fishy. Are we missing the right expression there?

alexanderpann:
@wsafonovI I removed the code. The concept just can't implement `ITypeSupportsDefaultValue` since we can't return an appropriate expression here.

wsafonov (approved):
Looks good now. Thanks for the fixes!

PR #1151, opened 2024-11-28
build: improve local tests

sergej-koscejev:
Minor improvements to local Gradle builds.


alexanderpann (approved):
LGTM. Why does mps.generator.skipUnmodifiedModels needs to be removed?

sergej-koscejev:
> Why does mps.generator.skipUnmodifiedModels needs to be removed?

1. It shouldn't be used for build scripts because the generation result depends not only on the model content but also on the content of module descriptors (in order to report errors if they don't match).
2. If any other generator would use something besides direct model dependencies (like navigating the project to look for some preferences or something), it will also break.
3. If MPS finds no models to generate, it fails the build due to a bug.

Issue #1152, opened 2024-11-29
Provide an additional Environment for AsyncSolverTaskExecutor 


dbinkele:
AsyncSolverTaskExecutor has 2 Phases:
- create the task
- run the task

it might be necessary to provide a certain access level for each phase (read, write) from the calling party.
Currently only the first phase provides such a possibility.
Also the second phase mjst provide one, which is the purpose of this ticket.

alexanderpann:
Already implemented in #1153.

PR #1153, opened 2024-11-29
Provide Env for TaskRunPhase (#1152)

dbinkele:
solves #1152

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
@@ -683,7 +683,7 @@
         <node concept="3clFbF" id="7QODtLvTGsE" role="3cqZAp">
           <node concept="2YIFZM" id="7QODtLvTGNW" role="3clFbG">
             <ref role="1Pybhc" to="hnhi:2f_Mi5mAhjh" resolve="AsyncSolverTaskExecutor" />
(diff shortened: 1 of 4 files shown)


kbirken (changes requested):
Looks good, but: in several places in the productive code and in tests you introduced the following code (with varying inner argument, the one providing the repository):

image

It would be nice to have some static helper method which can be used at all these places.

dbinkele:
> Thanks for refactoring the duplicate code places. But: In the new tests there are also three more of them... example [node URL](http://127.0.0.1:63320/node?ref=r%3A323df21b-bcb0-4525-967d-a41bb2c3fcb8%28test.org.iets3.analysis.base.async%40tests%29%2F682624529721994484)

The mentioned method has a 'Repository' as a parameter.  The methods introduced in the last step has a 'node' as a parameter from which it takes the model and then the repository. In the tests we take the repository of the project. Thus, there is no way of substitute one for another.......?!

kbirken:
> @kbirken The mentioned method has a 'Repository' as a parameter. The methods introduced in the last step has a 'node' as a parameter from which it takes the model and then the repository. In the tests we take the repository of the project. Thus, there is no way of substitute one for another.......?!

During my brief inspection the methods looked quite similar (only the parameter differed), so I thought it were feasible to abstract out a common part. But if not possible, no problem, I'll approve.

kbirken (approved):
LGTM

PR #1162, opened 2024-12-04
Only activate custom views through an extension point

alexanderpann:
All custom views are deactivated by default now. You can enable each of them through their extension point: 
- components explorer (`components` extension point)
- specification explorer (`spec` extension point)
- requirements explorer (`requirements` extension point)
- bookmark explorer (`bookmark` extension point)

Only some were activated before. I decided on the opt-in approach this time because they are rarely desired/used.


HeikoBecker (approved):
Thanks for the update.

Functionality works as expected. By instantiating a new extension point I can enable/disable the views.

No findings in code.

Issue #1115, opened 2024-11-13
Multi dimensional lists: Typesystem of lists does not reflect upper bounds of inner lists


arimer:
**Expected behaviour**
The amount of elements in the inner list should not be limited and no error should be present.

**Current behaviour**
<img width="1545" alt="grafik" src="https://github.com/user-attachments/assets/7240f0c9-f3cb-4c9f-afaa-e736b808dd75">

Issue #1114, opened 2024-11-13
Typesystem of lists does not reflect lower bound


arimer:
Crosscheck if this this a bug or idended behaviour. 

<img width="831" alt="grafik" src="https://github.com/user-attachments/assets/1b131936-c1e5-4e7f-bab4-52fd99e88715">

**Expected behaviour**
 The outer list has a lower bound of 0 and therefore it is not mandatory that the list contains any entries. The error should not be there.

alexanderpann:
The error message is still correct since you need to set the type for empty list literal yourself but the derived list size should be 0.

PR #1170, opened 2024-12-05
Improve (nested) list literal types

alexanderpann:
- fixes https://github.com/IETS3/iets3.opensource/issues/1114
- fixes https://github.com/IETS3/iets3.opensource/issues/1115 only for one level of nesting. I really don't know how to achieve a general solution. You need the types and the original list contents at the same time. At the level of the replacement rules I can't decide if I can merge two list ranges in the case of a list of lists or if I should check if they are compatible e.g. a list literal is assigned to a variable of type list. I also can't adapt `SimpleTypesPrimitiveTypeMapper#computeSupertype` since it can only calculate the type for simple types and not collections.
The target branch is 2023.2 because there was a big refactoring in the common type calculation starting in this branch.

I also enabled the joined type for collections since the type can be calculated already but there is not a single place that supports it. If we don't want to enable it there, I can remove the last commit and we create a separate ticket for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## December 2024
+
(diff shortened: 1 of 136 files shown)


HeikoBecker (approved):
Thanks for fixing these bugs!

Functional review: The described bugs are not reproducible anymore.

Technical review: No findings in code.

PR #1174, opened 2024-12-05
Include a GitHub workflow for vulnerability scanning

alexanderpann:


.github/workflows/vulnerability-scanning.yml
```diff
@@ -0,0 +1,33 @@
...
+          JAVA_HOME: /opt/jdk
+        id: Depcheck
+        with:
+          project: 'iets3.opensource'
+          format: 'HTML'
+          out: 'reports'
+      - name: Upload Test results
+        uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4
+        with:
+           name: Depcheck report
```
arimer, line 32: Lets give it a full name like : "Dependency Check report"

arimer:
At the moment the report is just uploaded as an aritfact. What is the intention behind this? Will there be an automated check of the report or anything like that in the future?

arimer:
1. Are the platform dependencies to mbeddr / mps-extensions / mps-qa excluded by default? A local dependency calculation looks like this, but most of those deps. did not make it to the report?
grafik
**Report:** 
grafik
2. What about the custom JBR that we are using for building? Do we need to list it?
3. Could you maybe link the documentation to the provider of that plugin?

alexanderpann:
> At the moment the report is just uploaded as an aritfact. What is the intention behind this? Will there be an automated check of the report or anything like that in the future?

I first wanted to check if we have any vulnerabilities. We could set `--failOnCVSS 7` (fail on severity level 7 or higher out of 10) so that the build fails when an issue is found. If we fail the build, someone has to fix it immediately, which we might not want.

> Are the platform dependencies excluded by default?

They seem to be excluded, but to be honest I don't understand how it works. For me this behavior is correct since we only want to check the current platform and also keep the checking fast.

> What about the custom JBR that we are using for building?

If we want to check it, the vulnerability check would have to be done in https://github.com/mbeddr/build.publish.jdk. Since we don't build anything here, the generic JDK is fine.

arimer:
@alexanderpann do we still want to merge this into master?

alexanderpann:
Yes. We Also have it enabled an optional check in MPS-Extensions and mbeddr.core.

.github/workflows/vulnerability-scanning.yml
```diff
@@ -0,0 +1,33 @@
...
+        id: Depcheck
+        with:
+          project: 'iets3.opensource'
+          format: 'HTML'
+          out: 'reports'
+      - name: Upload Test results
+        uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4
+        with:
+           name: Dependency Check Report
+           path: ${{github.workspace}}/reports
```
arimer, line 33: Where is this path and the records acutally stored? Is this a publically available directory or something else?

alexanderpann: The reports are only stored temporarily in this "reports" folder. The report itself should be attached in the GitHub actions run.

Issue #943, opened 2024-05-03
Not all concepts with types implement IOptionallyTyped or ITyped


alexanderpann:
Some examples: components.Parameter, components.Variable, adt.AlgebraicConstructorArg, adt.Ancestor, adt.QuotedTermType, EmptyValue

I encountered the issue, for example, with the concept _EnumDeclaration_.  We should also check if IOptionallyTyped could extend ITyped.

PR #1183, opened 2024-12-12
re-enable untyped columns in DataTable

enikao:
re-enable untyped columns in DataTable

Used to work before #943


enikao:
@alexanderpann What do you think, can we merge that?

alexanderpann (approved):
LGTM. You can also add me as a reviewer so that I get a notification the next time

PR #1188, opened 2024-12-18
remove references to removed language com.mbeddr.mpsutil.httpsupport

alexanderpann:
This language doesn't exist anymore.


HeikoBecker (approved):
Thanks for the fix!

PR #1190, opened 2025-01-07
Improve node factories and copy-pasting in general

alexanderpann:
I introduced 2 new interfaces: `IContainAssignableExpression` and `IContainExpressionParam ` and let many concepts implements them. When you now use the code completion menu to replace a node with another one, the expression is copied over to the new node.
Examples:
- sin(10) -> cos(10)
- 3.1415 -> sin(3.1415)

In addition, I implemented a lot of node factories to copy even more information from the old nodes.

For a few top-level expressions, I implemented experimental copy-paste handlers:
- `DotExpression` -> `IToplevelExprContent`
- `Function` -> `Expression`
- `Constant` -> `Expression` 
- `EnumLiteral` -> `Expression`

Example: When you copy a function declaration, it will paste a reference to the function into the expression or a constant reference instead of the constant declaration itself.

As long as https://youtrack.jetbrains.com/issue/MPS-24089/copy-paste-handlers-are-not-called-for-extended-concepts is not implemented, it doesn't make sense to implement them for all concepts as it would lead to a lot of code duplication.

For reviewing: All changes related to the refactoring and introduction of those interfaces don't have to be reviewed since all the migrations were done automatically by MPS. I checked the project for broken references and I couldn't find any. Only the node factories and other commits have to be reviewed.
I **will** squash all the "use X for Y" commits after the review but kept it like that in case we need to remove some commits.

https://github.com/user-attachments/assets/c3c3594f-d7d6-4692-8e98-7b3d60133899

https://github.com/user-attachments/assets/3f9af6d3-cf61-44da-93ca-46c9366ed4bd


HeikoBecker (commented):
Thank you for this change. This is a very helpful feature.

Functional Review:
- copy-pasting a constant as a reference: works
- copy-pasting a function definition as a function call: works
- copy-pasting a enum literal as a reference: works
- replacing a sin with cos: does not work, I cannot open the completion menu
- creating a parenthesis expression and then surrounding it with `sin` does not work either

Overall feedback: For documentation purposes I would suggest to record one or two GIFs or screenshots of how the feature can be used. This will make it easier to demo this to new users.

Minor note: The PR description mentions a `IContainExpression` interface. I cannot find it in the code. I assume it has been replaced with the `IContainExpressionParam` interface.

HeikoBecker (changes requested):
Thank you for this change. This is a very helpful feature.

Functional Review:
- copy-pasting a constant as a reference: works
- copy-pasting a function definition as a function call: works
- copy-pasting a enum literal as a reference: works
- replacing a sin with cos: does not work, I cannot open the completion menu
- creating a parenthesis expression and then surrounding it with `sin` does not work either

Overall feedback: For documentation purposes I would suggest to record one or two GIFs or screenshots of how the feature can be used. This will make it easier to demo this to new users.

Minor note: The PR description mentions a `IContainExpression` interface. I cannot find it in the code. I assume it has been replaced with the `IContainExpressionParam` interface.

alexanderpann:
> replacing a sin with cos: does not work, I cannot open the completion menu

You have to select the full expressions and use the code completion menu. I'd love to fix the code completion menu of all the  alias cells as well, but this change is so big and I want to do it in a separate PR.

HeikoBecker (approved):
With the suggested approach, I could finish the functional review. No further findings.

Also, no findings in the technical review. LGTM.

PR #1192, opened 2025-01-15
Fix for a NullPointerException in one of the new node factories

alexanderpann:


HeikoBecker (approved):
Thank you for the quick fix! LGTM.

PR #1199, opened 2025-01-27
IETS3OS-1197: Provide a marker for variation point

### LINKED ISSUE #1197: Provide new variability-language which contains a marker interface IVariationPoint
In customer applications of IETS3 Variability, it is sometimes required to check if some artifact node is subject to variability (i.e., has a variation point). In order to do the check without a dependency to the IETS3 Variability framework, we would simply require a marker interface in IETS3-opensource, which is the base interface for all variation-point related interface concepts.

The steps that have to be done in iets3.opensource are the following:

1. Introduce a new language `org.iets3.variability.artifacts.base` which contains the marker interface `IVariationPoint`. Note that this language might contain more of these marker interfaces in future.
2. In this new language, provide a utility API to check if some BaseConcept node has a NodeAttribute of concept `IVariationPoint`.
3. We also need an API method `Optional<node<Expression>> IVariationPoint.getPresenceCondition()`, which provides the presence condition if the variation point has one. The dependency to concept `Expression` shouldn't be a problem as this concept is in iets3.opensource anyway.

In future, we can use this new language to provide more structural information about variability without requiring the IETS3-Core dependency.

See more details in [ticket 1469](https://github.com/IETS3/iets3.core/issues/1469) in iets3.core.

mfsoliveira:
This PR creates a language for variability in iets3.os.
ITs first and only feature is a marker interface, so that users can identify which concepts are variation points.
By using this interface features in other languages can test is some concept is a variation point and decide if the feature applies to it, without depending direct on the iets3.core variability framework.

This PR resolves #1197


kbirken (changes requested):
I did a quick review, and came up with these findings:

- After checking the customer project, I saw that we will need another API method soon - thus I extended the #1197 ticket by item 3 (a second API method). Sorry for coming up late with this requirement.
- The empty generator should be deleted from the new language.
- The IETS3 Variability framework tries to cover two use cases: when the variation point is attached as a NodeAttribute, and when the artifact node is directly implementing the variation point interface. The API methods should also handle the latter case (`queriedNode.isInstanceOf(IVariationPoint)`).

kbirken (dismissed):
Looks good now, thanks for adapting and providing the additional method.

code/languages/org.iets3.opensource/.mps/libraries.xml
```diff
@@ -7,15 +7,15 @@
           <value>
             <Library>
               <option name="name" value="mbeddr.plattform" />
-              <option name='path' value='$PROJECT_DIR$/../../../build/dependencies/com.mbeddr.platform' />
```
arimer, line 10: Can we maybe revert this file? Sometimes MPS tend to-regenerate it newly,

arimer:
I am not so familiar with the variablity framework. So my questions might sound suprising. But do you really have SModelAttribitues that implement `IVariationPoint` [direcltly]( nodes of http://127.0.0.1:63320/node?ref=r%3Ab827b941-aa66-4f79-b830-0e73f1e33465%28org.iets3.variability.artifacts.base.behavior%29%2F8774008540172854732)?

arimer (changes requested):
In general the changes look fine to me. Please revert the libraries.xml file and we are good to go 👍🏼 


Just some remark: Please try to avoid replacing part in the build script + re-adding them in a different place like it was done here. This might cause some confusion.
grafik

kbirken:
> I am not so familiar with the variablity framework. So my questions might sound suprising. But do you really have SModelAttribitues that implement `IVariationPoint` [direcltly]( nodes of http://127.0.0.1:63320/node?ref=r%3Ab827b941-aa66-4f79-b830-0e73f1e33465%28org.iets3.variability.artifacts.base.behavior%29%2F8774008540172854732)?

There is an interface concept (currently called `IVariationPoint`, will be renamed to `IVariationPointBase`) in IETS3.core which will implement this marker interface soon. And many concepts in application projects are NodeAttributes and implement this interface concept from IETS3.core. So, not directly, but via intermediate interface  concept(s).

arimer (approved):
Let's go 🚀

Issue #1196, opened 2025-01-23
Interpreter run configuration entry not long enough + not distinguishable from normal run configuration


alexanderpann:
Both entries nearly look the same:

<img width="643" alt="Image" src="https://github.com/user-attachments/assets/7aee1b17-6cc7-4965-b123-9a1014433e59" />

mgronover:
We should also provide a different but similar icon.

mgronover:
Because there seems no way to access the width of the submenu item to enlarge it, i created a ticket for this and the suspicious location of the icon:

https://youtrack.jetbrains.com/issue/MPSI-88/A-new-Run-Configuration-appears-quite-ugly-in-the-menu

alexanderpann:
It is more complicated than that. You probably have to do this on the IntelliJ Platform level or even understand the theme itself. This issue is about automatic ellipsing (...) of text in AnAction items in menus.

mgronover:
Maybe, but it should not be the user to get that correct.
MPS has an extra mechanism to provide run configurations. 
This seems to be incomplete.

PR #1201, opened 2025-01-29
Bugfix/menu entry not distinguishable 1196

mgronover:
addresses issue: #1196

(We should not wait for JetBrains fixing the length of the menu entries. These are distinguishable from the first word and the icon now)


arimer:
@mgronover is this PR ready for review?

mgronover:
yes!

arimer (commented):
To minor findings. We should discuss if they are easy to fix. 

1. Although the intererpreter run is successful, the icon left of the "tests passed" label seems odd. For the regular run configuration a green check mark is used.
grafik
2. To me it looks like that the new ion is slightly bigger then the old one. Any change to give it a similar size?

mgronover:
Ad 1. Fixed
Ad 2. Maybe it is copied from the new UI icons for unit tests with colors exchanged.

arimer:
Instead of creating a new icon that we need to maintain, I would opt for re-using the existing icon and changing it's colors/tones to make it easiert to distinguished. 

**New UI in dark/light mode**
grafik
grafik


**Old UI in dark/light mode**
grafik
grafik


What do you think?

arimer (approved):
To the moon! 🚀

PR #1206, opened 2025-02-04
improve the usability of some editors

alexanderpann:
The goal of this PR is to improve the usability of editors. All aliases cells should support code completion e.g. the word "enum" in an enum declaration and all keywords should have the right color. This must be done for all editors where there is more then one editor cell. For single cells the code completion menu works by default.

In addition, I fixed the issue with string types vs string types with constraints which broken model checking in a few editors.


alexanderpann:
> This editor diff looks strange.

I noticed it as well but it is correct.

> Potential typo
Fixed.

The last issue is unrelated to my changes. I created a ticket for it: https://github.com/IETS3/iets3.opensource/issues/1210

HeikoBecker (approved):
Thanks for addressing my nitpicks and for improving the editors!

Please wait for the review of @arimer before merging this PR.

Issue #1207, opened 2025-02-04
IETS3 depends on mbeddr platform parts that are not in the build script


malueck:
The IETS3 build script has some dependencies on the `com.mbeddr.mpsutil` plugin, which is not a declared dependency in the `org.iets3.opensource` build script. The build script contains only dependencies to the already "modularized" mpsutil plugins.

The dependency issues can be fixed by not using `com.mbeddr.mpsutil`, changing mbeddr to further modularize mpsutils, or (least favorized) add back a dependency to the whole `com.mbeddr.mpsutil` plugin to the build script.

One indirect dependency was `com.mbeddr.mpsutil.multilingual.common`, which was fixed on mbeddr side.

For now I checked only `iets3.core.os`, some remaining such dependencies are:

* through solution model `org.iets3.core.expr.typetags.physunits.documentation` it depends on `com.mbeddr.doc` as "Used Languages", but there is no dependency, in the build script only `iets3.req.os` depends on `com.mbeddr.doc`
* through solution model `org.iets3.core.expr.stringvalidation.runtime` it depends on `iets3.core.assessment` as "Used Languages", but there is no dependency declared, however according to MPS this Used Language is **unused and can just be removed**
* through language aspect `org.iets3.core.expr.tests.structure` it depends on `com.mbeddr.mpsutil.bldoc`, again only as "Used Languages", and `com.mbeddr.mpsutil.bldoc` is actually only packaged as the `group.multilingual` group in the `com.mbeddr.platform` build script, so like the multilingual plugin it is only part of the big `mpsutil` plugin which is not a declared dependency. This again can be easily removed, because there is only one such node instance, a **BLDoc** attached to `AbstractMatcherType` which has an empty text.

PR #1209, opened 2025-02-05
Use Jetbrains httpsupport, and remove lots of unused imports/aspects

malueck:
Removed the URLUtil class (and jetty.io dependency) and use jetbrains.mps.ide.httpsupport instead, which does exactly the same but is part of MPS now.

Also removed lots of unused imports and empty aspects.

In the build script, added as plugin dependencies:
de.slisson.mps.conditionalEditor
de.itemis.mps.celllayout
de.itemis.mps.grammarcells
com.mbeddr.mpsutil.intentions

For this, I am not sure if it is the correct approach to add these like this for all used languages / runtime dependencies. We would that the plugin will disable itself in the MPS Plugin settings and say for example "com.mbeddr.mpsutil.intentions is missing" if this is the case, and at the moment it simply throws an exception on start if com.mbeddr.mpsutil.intentions is absent. Also, we would like that the model checker would find an error in the build script if a dependency is not satisfied. However, I am not sure if it already doesn't do so and it is just the job of the plugin to declare its dependencies somehow differently (adding those as "Dependencies" to the language will show them as unused).
Any thoughts on this? Is this step with adding build dependencies the correct way?


malueck:
This will resolve item 2) of https://github.com/IETS3/iets3.opensource/issues/1207: org.iets3.core.expr.stringvalidation.runtime has dependency on iets3.core.assessment that is not in the build script.

alexanderpann:
Your change with the httpsupport plugin basically reverts https://github.com/IETS3/iets3.opensource/commit/2ebc1181f55b9180b2b159d0df88940a11bfa2e8. To be on the safe side, we should probably add the httpsupport plugin [here](http://127.0.0.1:63320/node?ref=r%3Ac3d6ae0c-8b10-477f-a3e9-5dc8700ceb13%28org.iets3.opensource.build.build%29%2F1893833104133534633) as an additional ideal plugin. Other than that, I am fine with the changes.

malueck:
Oops, you are right, the httpsupport use was indeed removed on purpose. But if this is fine -- I added it as loaded plugin to the tests as you suggested. The tests seem to pass, is there anything else to check that the issue with httpsupport does not appear again?
Looking at the other commit message, I am not sure what "tests executed out of process" means in this context. When I close MPS and run them with gradlew, they pass, but perhaps there is more to that?

I also added com.mbeddr.mpsutil.userstyles as a build dependency, which is a used language through org.iets3.core.base, similar to com.mbeddr.mpsutil.intentions. Like the other one, userstyles was not yet a build dependency (neither direct nor through mbeddr.core.base) but having it missing makes all the IETS3 languages not deploy because they use it. The build solution should now take care of that.

alexanderpann (approved):
"tests executed out of process" referred to the test run configurations where you can make the tests execute in-process or out-of-process. They don't work properly anyway in some cases, so let's just merge this PR to fix the other issues.

PR #1212, opened 2025-02-12
Change some migration descriptions so you can better differentiate them

alexanderpann:
I added the text "of concept X", so that they not all look the same:

Screenshot 2025-02-12 at 10 39 38


mgronover (approved):
Well done!

PR #1214, opened 2025-02-12
Move languages from iets3.core 

mfsoliveira:
Move languages protocol.transport, component.functional, component.hardware, and mapping from iets3.core to iets3.os
The goal is to allow customer projects to have direct access to open source code and remove dependencies to iets3.core

The specific point is: component.functional has dependency to analysis.solvertask, which is exclusively iets3.core.
In order to remove that dependency and be able to move the required languages an extension point was defined. In this way  component.function delegates to the extension point the implementation that needs a concrete implementation of solver task. iets3.core shall implement this  extension point.


kbirken:
Some early review findings: 

- The devkit `org.iets3.functionalcomponenets` has a typo, and I would name it `org.iets3.components.functional.devkit` anyway (most other devkits in iets3-opensource have `.devkit` as name suffix).
- Similarly, the devkit `org.iets3.hardwarecomponents` should be renamed as `org.iets3.components.hardware.devkit`.
- Finally, the devkit `org.iets3.components` should be renamed as `org.iets3.components.devkit`.
- The languages `org.iets3.core.maping` and `org.iets3.protocol.transport` (and their test solutions) are strongly related to the components-languages and should be in same virtual folder `comp` (maybe in a subfolder).

kbirken:
You did moving the languages and adapting the code in one step, so I cannot see the changes you did along the way. At least you replaced the `createSolverTask` implementations of `DataConnectorType` and `AbstractState`, using the STF extension point like this:

image

What else did you (have to) change?

kbirken:
> I renamed those devkits. I thought that maintain the same name is better due to compatibility...but we need to update anyway. Then it is better to have a better name.

On file system level, the directories and files still have the old names. Did you skip this renaming deliberately?

image

kbirken:
> On file system level, the directories and files still have the old names. Did you skip this renaming deliberately?

Well, my fault :-(. The directory and file names are indeed correct.

wsafonov (changes requested):
Why do we check-in iml idea files for the languages? I don't think we need them in the repo.

Also for the test solutions, we used to have a separate tests folder to organize all test solutions there. I would prefer to put all test solutions there. Right now the PR puts some test solutions under `$PROJECT_DIR$\solutions` and some under `$PROJECT_DIR$\tests`

wsafonov (approved):
Looks good now after some clean-up.

PR #1217, opened 2025-02-17
remove references to removed language com.mbeddr.mpsutil.httpsupport

mgronover:
Removes httpsupport references.


alexanderpann (approved):
Thanks for the change!

PR #1219, opened 2025-02-18
Make string types with constraints comparable with strings

alexanderpann:
This is a followup of https://github.com/IETS3/iets3.opensource/pull/1206/commits/4c9d10749eaac618e811c7e09a2066c9d9653597. We also need the same rule for the direction `string type with constraint` -> `string`.


mgronover (approved):
LGTM

PR #1222, opened 2025-02-19
add the missing dependency on com.mbeddr.mpsutil.modellisteners in the org.iets3.core.os idea plugin

alexanderpann:


sergej-koscejev:
Note: the dependency is not detected by MPS ([MPS-38286](https://youtrack.jetbrains.com/issue/MPS-38286) Build language does not detect plugin dependency from used languages) but it is in fact required.

PR #1227, opened 2025-02-26
fix for mixing nix and tuple types

enikao:
before, this created an error:

```
val t1 : [int, int] = if true then [1, 2] else nix 
```

because the type of `if` was _join(nix,tuple)_


alexanderpann (approved):
LGTM, thanks for the contribution!

Issue #1226, opened 2025-02-25
Evaluation of ShortLambda takes more time than execution of the same expression in normal Lambda


mgronover:
Evaluation of ShortLambda takes more time than execution of the same expression in normal Lambda.

PR #1229, opened 2025-03-06
Bugfix/decrease short lamba evaluation time

mgronover:
Addresses:  #1226

This PR decreases the evaluation time for ShortLambdaExpressions (without a breaking change).

The main problem of the old approach was, that caching could not be applied, because the nodes used to evaluate the expression was recreated on every interpretation. 

Also a fix in the trace explorer was done.


mgronover:
> Looks good to me. You just misspelled the word "lambda" a few times in the behavior aspect.

I fixed the misspelling.

PR #1246, opened 2025-03-18
Package SBOM in the language library distribution

wsafonov:
Package current SBOM as part of the language library distribution to provide meta information about the library components and dependencies and enable OSS scanning and analysis.

SBOM is generated using [CycloneDX Gradle Plugin](https://github.com/CycloneDX/cyclonedx-gradle-plugin?tab=readme-ov-file#cyclonedx-gradle-plugin) with default settings (license test is excluded from the generated SBOM). Since this plugin relies on the Gradle dependencies management, as part of this PR dependency JARs stored in the repo have been replaced by the Gradle-managed dependencies:
-  [functionaljava](https://mperry.github.io/fj-web-fidbake/): used only in the KernelF Java generator (previously used version `4.8-SNAPSHOT` has been updated to `4.8.1`)
-  [takari-cpsuite](https://github.com/takari/takari-cpsuite): used for creating TestSuite for all KernelF generated tests

Also Gradle has been updated from `8.0.2` to `8.13` due to the new Gradle API used in `download-jbr` plugin (requires Gradle version >= `8.3`).


alexanderpann (dismissed):
LGTM

arimer (dismissed):

PR #1261, opened 2025-03-26
activate migration checking

mfsoliveira:


build.gradle
```diff
@@ -9,21 +12,28 @@ buildscript {
     repositories {
         maven { url 'https://artifacts.itemis.cloud/repository/maven-mps/' }
     }
-    dependencies {
```
sergej-koscejev, line 12: The whole `buildscript` block can be removed now (including the comment above).

build.gradle
```diff
@@ -9,21 +12,28 @@ buildscript {
...
     id "org.cyclonedx.bom" version "1.8.2"
 }
 
+ext.dependencyRepositories = [
+    'https://artifacts.itemis.cloud/repository/maven-mps/'
+]
+
 repositories {
-    maven {
-        url 'https://artifacts.itemis.cloud/repository/maven-mps/'
```
sergej-koscejev, line 26: What's the point of this change?

build.gradle
```diff
@@ -86,7 +96,7 @@ if (project.hasProperty('iets3OpenSourceVersion')) {
         } else {
             def isSnapshot = !(currentBranch.equals("master") || currentBranch.startsWith("datev-loon-staging-") || currentBranch.startsWith("datev-steuer-staging-"))
             version = GitBasedVersioning.getVersionWithCount(major, minor, buildNumber) + (isSnapshot ? "-SNAPSHOT" : "")
-        }
+        }
```
sergej-koscejev, line 99: Unnecessary
```suggestion
        }
```

build.gradle
mfsoliveira: buidscript block removed

build.gradle
mfsoliveira: I just try to keept the changes from Alexander script. Shall I remove it?
From where come the nexus repository information? It looks like that loop is a workaround for some bug.

sergej-koscejev (approved):

Issue #1264, opened 2025-03-27
The result of an .project() operation produces incomparable items cause a successive toSet operation not working properly


mgronover:
These test will fail:

![Image](https://github.com/user-attachments/assets/e5459fed-f7fc-445b-91c0-1c7f733d8784)

arimer:
Closed with #1266

PR #1266, opened 2025-03-27
toplevel.rt.rt: introduced hashCode() and equals for projection results.

mgronover:
addresses issue: #1264


arimer (approved):
Really nice catch! 💪🏼 

Thanks for the fix

arimer (approved):

PR #1273, opened 2025-03-31
Feature/update datev loon staging 20232

arimer:
Updating with latest maintenance/20232

- kept original DotExpression typing rule


mgronover (approved):
LGTM.

Issue #686, opened 2023-07-20
Numbers in scientific notation without number part shouldn't be allowed


alexanderpann:
The following expression can be entered: `e-5 / 10 equals 1e-6`.
The number only has an exponent (e-1) which is allowed according to the constraints but doesn't work with the type system and interpreter:
```
java.lang.NumberFormatException: No digits found.
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:588)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:467)
	at java.base/java.math.BigDecimal.<init>(BigDecimal.java:896)
	at org.iets3.core.expr.base.plugin.InfHelper.div(InfHelper.java:119)
```

We should probably disallow those numbers.

PR #1276, opened 2025-04-03
expr.simpleTypes: require a number before the "e" in number literals

alexanderpann:
- Number literals in scientific format, e.g. 1e1,0 now require a number before "e". Numbers like "e-5" are not valid anymore because they caused exceptions in other parts of the code base.
- fixes https://github.com/IETS3/iets3.opensource/issues/686


HeikoBecker (approved):
Thanks for the fix. LGTM.

Issue #1283, opened 2025-04-08
In case of interpretation of test cases (no generation!) for modules the prefix "Generated_" has to be removed


mgronover:

arimer:
Fixed with https://github.com/IETS3/iets3.opensource/pull/1284

PR #1284, opened 2025-04-08
fixes module names in case of interpreting test cases

mgronover:
addresses issue: #1283

### DIFF
--- a/code/languages/org.iets3.opensource/solutions/org.iets3.core.junit.interpreter.run.configuration/models/org.iets3.core.junit.interpreter.run.configuration.plugin.mps
+++ b/code/languages/org.iets3.opensource/solutions/org.iets3.core.junit.interpreter.run.configuration/models/org.iets3.core.junit.interpreter.run.configuration.plugin.mps
@@ -353,10 +353,12 @@
         <child id="1170075728144" name="expression" index="1HWFw0" />
         <child id="1170075736412" name="block" index="1HWHxc" />
       </concept>
+      <concept id="5351203823916832286" name="jetbrains.mps.baseLanguage.structure.ResourceVariable" flags="ng" index="3J1hQo" />
       <concept id="5351203823916750322" name="jetbrains.mps.baseLanguage.structure.TryUniversalStatement" flags="nn" index="3J1_TO">
         <child id="8276990574886367510" name="catchClause" index="1zxBo5" />
         <child id="8276990574886367509" name="finallyClause" index="1zxBo6" />
         <child id="8276990574886367508" name="body" index="1zxBo7" />
+        <child id="5351203823916750334" name="resource" index="3J1_TS" />
       </concept>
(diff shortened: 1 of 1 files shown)


arimer (commented):
Some open queations:

1. Why do we need to use a [constructor call ](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F1143330301452148141) to set the ExecutionModePreference. The lcoal variavle you are creating (setting) is never used elsewhere.
2. Don't we need the setting of the ExecutionModePreference.INTERPRETER also when running the testes on model/module/project level?

mgronover:
> 1. Why do we need to use a [constructor call ](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F1143330301452148141) to set the ExecutionModePreference. The lcoal variavle you are creating (setting) is never used elsewhere.

This is a pattern. Instead of resetting the ExecutionModePreference locally each time before we return from the method it is done when the "resource" is released i.e. in the destructor of the released object. Because in Java there is no destructor the "try with resource" is used here.

> 2. Don't we need the setting of the ExecutionModePreference.INTERPRETER also when running the testes on model/module/project level?

Because this is the only case within the producer, where the name of the module is used (only for the menu entry).
(The other case (it was overlooked before) is within the Executor, when the wrapper nodes are created.)

arimer (commented):
I've seen that you have added a done() method to `AutoResetTestExecutionWorkspaceSetting`. However this method is never from the producer. Shoudn't we do this at the end of the try-case on the [settings](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F1143330301452144682) instance?

mgronover:
> I've seen that you have added a done() method to `AutoResetTestExecutionWorkspaceSetting`. However this method is never from the producer. Shoudn't we do this at the end of the try-case on the [settings](http://127.0.0.1:63320/node?ref=r%3A52c49891-65a1-45a5-a86b-4a7aa1e6c12e%28org.iets3.core.junit.interpreter.run.configuration.plugin%29%2F1143330301452144682) instance?

No. The `done` method is just needed to ease usage without a `try-with-resource block`.

We have two use cases to deal with:

1. **With a` try-with-resource block`**: we want to free a previously acquired resource and ensure this without explicitly calling the according methode (here: `close`) when method returns (multiple execution paths/multiple returns).
2. **Without a `try-with-resource block`**: in this case the caller just has a single execution path during the usage of the resource. So acquiring the resource and releasing it can be called just before and after that usage (of cause we could also use the` try-with-resource block` here, but it appears useless).

arimer (approved):
👍🏼

Issue #1307, opened 2025-04-16
IValidNameConcept.getIdentifierName() throws NPE if name is unset


enikao:
If any of `allowUmlaute()`, `allowParagraph()`, `allowNonIdentifierNames()`, `allowEverythingExceptWhitespace()` is `true`, `IValidNameConcept.getIdentifierName()` calls `this.name.replaceAll()`. This leads to a NullPointerException if `name` is `null`.

Version: 2023.2.8104.db4c887

PR #1313, opened 2025-05-06
iets3.core.base: Add a missing null check

alexanderpann:
- fixes https://github.com/IETS3/iets3.opensource/issues/1307


arimer (commented):
Thx for the fix! Great choise of a target branch 💪🏼 

In addition to your change, how about we add a @Nullable() annotation to that method?

arimer (approved):
Thx for the fix 👍🏼

Issue #1318, opened 2025-05-08
org.iets3.safety plugin name collides with the same plugin name in core


mfsoliveira:
The duplication of plugin name in both os and core is causing build problems in customer projects.
Suggestions:
rename org.iets3.safety in os to org.iets3.safety.os

mfsoliveira:
PR # 1319 fixes it

PR #1319, opened 2025-05-08
1318: Rename plugin org.iets3.safety to org.iets3.safety.os

mfsoliveira:
Rename plugin org.iets3.safety because this name collides with a plugin with same name in core (#1318)


arimer:
could you please do a `gradlew migrate remigrate` locally and commit your changes? The CI is still finding some dirty files.

arimer (approved):
Go for it

PR #1320, opened 2025-05-08
expr.simpleTypes: use infinite precision if any of the number types in the supertype calculation has an infinite precision

### LINKED ISSUE #1004: Supertype calculations in alternatives and binary expressions drop `{inf}` precision information when mixing `number{inf}` with integer types (`number{0}` or `int`).
image

The first two examples compute the type as expected. However, in the last two examples we get `int` and `number{0}` as result of the supertype calculation instead of `number{inf}`.

IMO the type calculation should preserve the {inf} precision property. 

image

alexanderpann:
- When calculating the super type of number types, the precision is now correctly set to infinite when one of the types has an infinite precision.
- fixes #1004

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - A NullPointerException was fixed for cases where a node implementing IValidNamedConcept had no name.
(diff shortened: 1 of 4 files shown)


enikao:
Maybe add a more direct test of the changed code, like
image

enikao (dismissed):

enikao (approved):

Issue #1323, opened 2025-05-08
Enable STF to cope with priorities


dbinkele:
STF consists of some logic to determine a a current implementation of a 'SolverTaskFactory' of the extension point 'solverfactories'. 
Add a priority in order to select the best matching in case there are several implementations

dbinkele:
done #1324

PR #1324, opened 2025-05-08
Add priorities & refactor (#1323)

dbinkele:
**[THIS IS TEMPLATE - ADJUST THIS TEXT ACCORDING TO YOUR PULL REQUEST]**

Each pull request should contain a brief summary of the proposed changes  
that can be easily read by the consumers of project releases.

Use the first comment in the pull request and update it as needed during
the work on pull request to keep the description up-to-date.

Some of the items in the description might include:

- compatibility issues with previous versions if this PR introduces any
breaking changes

- hints on the included migrations and things to consider when applying
the migration to the own project

- reference to a related Github issue

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.analysis.base/models/behavior.mps
@@ -680,22 +680,11 @@
       <property role="TrG5h" value="runManuallyAsync" />
       <node concept="3Tm1VV" id="7QODtLvTFn$" role="1B3o_S" />
       <node concept="3clFbS" id="7QODtLvTFnA" role="3clF47">
-        <node concept="3clFbF" id="7QODtLvTGsE" role="3cqZAp">
-          <node concept="2YIFZM" id="7QODtLvTGNW" role="3clFbG">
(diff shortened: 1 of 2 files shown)


kbirken:
Review finding: Shouldn't one of the methods `ISolvable.runManuallyAsync` or `ISolvable.runSolverAsync` be deprecated now?

dbinkele:
@kbirken your issues have been addressed

dbinkele:
@kbirken I came across some 'oddities'. We have to methods called 'createSolverTask'. One in 'SolverTaskFactory', the other in 'ISolvable'. The on in 'ISolvable' consistently delegates to 'STF'. It seems that we do not need this flexibility. 
I think it makes sense to make the one in ISolvable a no-virtual method. The only thing in 'ISolvable'-Implementations is that 'name' is different.  But following the usage path of 'name', it turns out that it is never used. 

My proposal:

1. From 'STF' and 'SolverTaslFactory' remove the 'name' parameter.
2. in 'ISolvable' make the method non-virtual (non-overrideable) and delegate to STF.

kbirken (dismissed):
Looks good now, thanks for improving the generic solvertask infrastructure!

kbirken (approved):
LGTM

Issue #1182, opened 2024-12-12
Do we really want StringTypeWithConstraints everywhere?


enikao:
#1096 introduced `StringTypeWithConstraints`, and it's now effectively the type of any IETS3 string.
This leads to several issues in our project:

* Everywhere we're testing for `StringType` we also need to test for `StringTypeWithConstraints`
* In Typesystem, we cannot test for StringType type equality anymore, need to test for subtype
* I haven't really measured it yet, but it feels like we have both a serious memory overhead (additional contents in temporarily created type nodes) and CPU overhead (e.g. from replacement rule `checkStringConstraint`)

Is this worth the advantages? Maybe we could introduce a switch whether to support constrained types?

alexanderpann:
I guess we should introduce an extension point to be able to go back to the old behavior. As a workaround, you should be able to create a new inference rule for StringLiteral in your language with overrides set to true and set the type back to the old string type.

alexanderpann:
With the latest changes, `StringType` and `StringTypeWithConstraints` should be comparable and don't throw any errors. We need test case though to see if the change actual works.

enikao:
One issue we have at the customer: Users should never see `StringTypeWithConstraints` as available type in autocomplete.

alexanderpann:
I see. If you have your own root concept, you can disable it through the `can be ancestor` constraint.

enikao:
> I see. If you have your own root concept, you can disable it through the `can be ancestor` constraint.

... which we would need to do in several places. I still like the idea of an extension point, or a flag in `PTF`.

PR #1325, opened 2025-05-09
core.expr: Make it possible to deactivate string constraints through the PrimitiveTypeMapper extension point

alexanderpann:
The primitiveTypeMapper extension supports a new method, `PrimitiveTypeMapper#useStringConstraints, ` which can be set to false to disable string types with constraints and go back to regular string types. The default implementation activates the string constraints, so we don't have another breaking change. Fixes https://github.com/IETS3/iets3.opensource/issues/1182

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 - A NullPointerException was fixed for cases where a node implementing IValidNamedConcept had no name.
 - When calculating the supertype of number types, the precision is now correctly set to infinite when one of the types has an infinite precision.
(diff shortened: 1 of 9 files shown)


enikao (dismissed):
We could add a test for it, see http://127.0.0.1:63320/node?ref=r%3A68ad96df-80cb-4919-8408-ee7fd9209228%28test.ts.expr.os.comma.plugin%29%2F3910975805742593968 for examples

enikao (approved):

enikao (approved):
Thanks for providing a test (-:

arimer:
FYI: artifacts.cloud is not reachable atm. That's why the build is failing.

arimer:
Artifacts problem should be solved. @alexanderpann I've re-triggered your build.

PR #1351, opened 2025-06-04
Only recognize `U+00B5 MICRO SIGN` as 'micro' prefix

sergej-koscejev:
Previously, `U+03BC GREEK SMALL LETTER MU` was recognized as well, but this was not consistent and led to problems.

A related change is that `AbstractUnitPrefixManager#isValidPrefix` is now case-sensitive.

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.base/models/behavior.mps
@@ -236,9 +236,16 @@
     </language>
     <language id="fd392034-7849-419d-9071-12563d152375" name="jetbrains.mps.baseLanguage.closures">
       <concept id="2524418899405758586" name="jetbrains.mps.baseLanguage.closures.structure.InferredClosureParameterDeclaration" flags="ig" index="gl6BB" />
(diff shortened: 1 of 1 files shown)


sergej-koscejev:
Can somebody using Windows confirm that AltGr+M produces U+00B5 and not U+03BC?

arimer (commented):
Some open questions:
- In `AbstractUnitPrefixManager` you are no longer use the `equalsIgnoreCase()` for comparing the prefixes with the keys of prefixExps. Why is this no longer needed? 
- Just out of curiosity, what is a `enhancement script`? Is this some kind of migration script?

sergej-koscejev:
> Some open questions:
> 
> * In `AbstractUnitPrefixManager` you are no longer use the `equalsIgnoreCase()` for comparing the prefixes with the keys of prefixExps. Why is this no longer needed?

I think this is actually incorrect because prefixes are case-sensitive in general (although both 'k' and 'K' are used for 'kilo' but e.g. 'n' is a valid prefix for 'nano' and 'N' is not). Also MU case-insensitve-equals MICRO, for some reason, and that's what we don't want.

> * Just out of curiosity, what is a `enhancement script`? Is this some kind of migration script?

Yes, see https://www.jetbrains.com/help/mps/scripts.html

arimer (approved):

Issue #1344, opened 2025-06-02
Add 'clone' method to IResult


dbinkele:
IResult must be able to create an exact clone of itself

dbinkele:
solved #1350

PR #1353, opened 2025-06-06
Feature/add clone method to i result #1344

dbinkele:
solves #1344


kbirken (approved):
Approved, with some remarks:

- There was an unneeded dependency added in structure aspect, removed it by an additional commit.
- The type of argument `errorMsg` in the original method `IResult.make` has been generalized from `string` to `Object`. This is no source compatibility break, but it is now allowed to make an IResult with something else than `string`. As this is also possible by subclassing the interface, it is probably just worth noting and no change request.
- The requirement from the ticket says "must be able to create a clone of itself". In case there are sub-results, the class implementing this interface has to override the clone-method and take care that the sub-results are cloned as well.

PR #1374, opened 2025-06-27
Add mapper to arguments (#1373)

dbinkele:
solves #1373


mfsoliveira (approved):
LGTM - test was added in the core project

Issue #1386, opened 2025-07-11
Ease the Testing of Error Highlightning of ISolvable


dbinkele:
A special Checker 'SolverChecker' is registered in 'TypecheckingSolverTask'-Plugin. To check that it is registered is must be a singleton

dbinkele:
done

PR #1387, opened 2025-07-11
Feature/ease the testing of error highlightning of i solvable #1386

dbinkele:
solves #1386


arimer:
@dbinkele is the target branch correct?

kbirken:
@dbinkele A first remark: Is there any relationship between the `DummyISolvable` stuff with the `H1Dummy` etc concepts, and the new `UITestSolvable` stuff? They are defined in the same language. If they are independent, it would be helpful to put these two independent parts (with all their concepts) into different virtual folders.

kbirken (dismissed):
LGTM, thanks for providing a test for the solver error highlighting.

One question: The `new SolverChecker()` initialization in `TypecheckingSolverTask` is now being done directly for the member, outside of the `init(project)` body. I just do not know if the semantics is same as with plain Java classes.

dbinkele:
> LGTM, thanks for providing a test for the solver error highlighting.
> 
> One question: The `new SolverChecker()` initialization in `TypecheckingSolverTask` is now being done directly for the member, outside of the `init(project)` body. I just do not know if the semantics is same as with plain Java classes.

Semantics is the same

PR #1394, opened 2025-07-15
issue warning for bang operator in case of non-option expression type

wsafonov:
Bang operator (`!`) normally cannot be used for expressions of a non-option type. However, this scenario is still possible if the model has been refactored or if the user unconsciously created an expression with bang operator placed there before the actual expression. Although the interpreter can still handle this scenario, it's probably a code-smell and could cause other unwanted side-effects in user models.

To avoid usage of bang operator for non-option expression types, there is now a warning informing the user of a potential problem.


arimer (changes requested):
Changes look ok in general. However one remark

- could you please update the changelog

PR #1422, opened 2025-08-20
Improved formatting for ListLiterals in multi dimensional lists

eyn9xwy:
This PR fixes an issue with the layout of list literals by using a new line when a threshold of three elements per list is exeeded.
image


alexanderpann:
@arimer Can you check why it is not building?

alexanderpann (approved):

Issue #1429, opened 2025-08-31
Configuration editor should behave properly on INSERT actions


kbirken:
When invoking the INSERT-action anywhere on the last item of the configuration editor, a new configuration should be created as a next-sibling to the current one. Currently this is working only if INSERT is invoked at the feature type icon (`M`, `A`, `O`).

Additional requirements:
- When invoking the INSERT-action at any other item of the configuration editor, the same behavior should be triggered as with the last item (create a new configuration as a next-sibling)
- All actions that could change the structure of the configuration itself in the editor should still be blocked (insert, delete, paste, ...).
- INSERT_BEFORE should executed only in the first line of a configuration; all other invokes of INSERT_BEFORE should be blocked.

PR #1431, opened 2025-09-02
Bugfix/config editor insert action 1429

kbirken:
The configuration editor can now properly handle pressing return/enter and shift-return/enter. The former creates a new line after the configuration (in `VariabilityModelChunk` roots) and can be pressed nearly anywhere in the configuration. The latter creates a new line before the configuration (only active in the header line of a configuration editor).

Due to deprecated configuration-related concepts in the `featuremodel.base` language the action maps for the editor have been located in the `featuremodel.base` language. As the editors of these deprecated concepts will only be needed to show old user models where the corresponding migrations haven't been executed yet, I just removed the definitions of the action maps. These are now reimplemented in the `configuration.base` language (which is the only language where they are actually needed).

Updated the CHANGELOG as well.

This solves #1429.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelo
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
 
+## September 2025
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


mfsoliveira (approved):
LGTM
The editor test is failing because "java.lang.NullPointerException: Cannot invoke "javax.swing.KeyStroke.getKeyCode()" because "this._3_stroke" is null"
It looks like a MPS Bug. We should have a better look at this latter. But for now I tested it manually and it woks fine.

PR #1444, opened 2025-09-18
Bufix/assignment causes behave unexpectedly #1650

dbinkele:
solves #1445


kbirken (changes requested):
The logic of flag `updateChanges` in `ConfigurationSolverFacade.needsUpdateToSolverResult` has been changed. I think the code should be simplified so that the flag is no more needed (and the `updateChanges |= ...` statements.

Additional remark: CHANGELOG should be updated (just a remark that handling of the assignment cause in feature attribute assignments has been fixed).

PR #1446, opened 2025-09-22
improve renderReadable/getPresentation implementations

alexanderpann:
The `renderReadable`/`getPresentation` implementations were improved and are now equivalent in all concepts. I am not sure why we need `renderReadable` in the first place. `getPresentation` should be enough.


arimer:
FYI: Failing due to missing credentials to acces  Sonatype OSS index.

enikao (approved):
proxying for  @AtanasM98

Issue #1460, opened 2025-10-01
Allow FilteringError to store more information


kbirken:
The class `FilteringError` records information about errors during the 150%-to-100% filtering in variability. This class should be extended to store additional information needed e.g. for implementation of IETS3/iets3.core#1654. 

The additional information should include:
- the feature model configuration which is used for the 150%-to-100% filtering
- an (optional) `node<>` depicting the location of the error in the 150% model

PR #1461, opened 2025-10-01
Extend FilteringError class to simplify locating the problem (#1460).

kbirken:
This mini-PR just extends one class (`FilteringError`), which is used during variability filtering. As the actual filtering algorithms are implemented in IETS3.Core, there is no usage of the class in IETS3.OS.

As the change is backward-compatible and not directly user-visible, there is no mention in the CHANGELOG.

This solves #1460.


dbinkele (approved):
LGTM

PR #1478, opened 2025-10-09
New checking rules for detecting inconsistencies in variability configurations

### LINKED ISSUE #1476: Detect and fix inconsistent states in feature model configurations
Due to erroneous resolution of merge conflicts, specific bugs and maybe other root causes there might be feature model configurations which are inconsistent. For the currently known cases, we should provide detection mechanisms (i.e., checking rules), and quickfixes. 

# Case 1: Root feature selection state

Problem description: The selection state of a root feature implicitly is always mandatory. However, this is not checked. It is possible to set it otherwise, thus breaking the solver execution. One possibility with current versions is to open the configuration in a configuration matrix editor and simply change the state (which will be blocked in a future version).

Fix: Add a checking rule which detects this, and an automatic quickfix which sets the selection state to mandatory.

# Case 2: Duplicate FeatureConfiguration for a given target feature

Problem description: This could happen by pressing "Ctrl-D" (duplicate) on a FeatureConfiguration node, or by an error during manual conflict resolution. 

Fix: Add a corresponding checking rule. Be aware that the checking rule should not be active if the duplicate feature is legal because of feature multiplicity. Add an non-automatic quickfix to remove one of the duplicates. The user has to decide which one should be removed, as one of the subtrees will contain productive information.

# Case 3: Duplicate names of FeatureModelConfiguration in a container

Problem description: This could be entered explicitly by the user, or by some erroneous manual merge. The severity is not so big, but the user should be made aware anyway.

Fix: Add a checking rule for duplicates in the same container. We will not check for duplicates across containers, as this might be done deliberately, and the check would probably have performance issues. We do not need a quickfix for this, as the user can simply change the names or remove one of the duplicate configuration.

kbirken:
This PR introduces some new checking rules for variability configurations and extends some existing checking rules. The basic idea is that the hierarchical structure of configurations should be checked more thoroughly, in case that during manual merge conflict resolving some inconsistencies are introduced. The PR also provides some quickfixes for repairing these issues, some automatic, some manual (if the user has to decide how to resolve). 

Note that many of the checking rules are disabled if the inconsistencies stem from a manual change in the feature model. Then, there will be an error "There were changes in the feature model. Run adapt intention..." on the configuration anyway.

Testing: Some tests have been added or extended to test the new checking rules. I had to update a couple of existing tests as well, as the test models already violated some of the new checks.

Compatibility: Although the changes as such are backward compatible, user models might show additional errors after the update. Some of the new errors will be fixed by automatic quickfixes, but for some user input will be needed.

The CHANGELOG has been updated.
 
Closes #1476.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 - Variability: For feature attributes, the inspector in the configuration editor shows the current value of the attribute and what caused this value (manual input by user, default, forced by tool, etc). This assignment cause is read-only now. An intention is added to manually set it to 'manual' state.
 - Variability: In the configuration editor, pressing Return/Enter anywhere will create an empty line after the configuration. Pressing Shift-Return/Enter on the header line of a configuration will create an empty line before the configuration.
 
+### Added
+
+- Data tables, binary and multi-criteria decision tables now support deletion, copying and pasting when multiple cells are selected with the mouse.
 
 ## July 2025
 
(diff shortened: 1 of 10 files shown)


kbirken:
> 1. There is a FM and A Config which are not used in the test

The `ConfWithoutError` should check that features with cardinality (like `F`) might be duplicated and the model check for non_unique_ref is not complaining. Is there a way to specify in a test that there is _no_ error in a part of the model?

dbinkele:
@kbirken an original hack to suppress  some properties

dbinkele (approved):

PR #1496, opened 2025-10-22
Some extensions for the o.i.req.core language

### LINKED ISSUE #1495: Some extensions for org.iets3.req.core language
With ticket IETS3/iets3.core#1663 generic variability support is added for the `org.iets3.req.core` requirements specification language.

The following changes are due to the variability integration:
- The `AbstractRequirement` concept should implement the `ICanHide` interface, thus allowing the variability artifact previewer to hide requirements which are not present in a given variant configuration.
- The indentation string computation for the table views should be refactored (variability also needs to do the same indent).

The following changes are useful even without variability:
- A new relation kind `requires` should be added, which can express that a requirement is based on or more other requirements. 
- A flag should be added to requirement chunk nodes allowing to hide empty child-requirements sections in the table view (shown in inspector).
- The presentation name of editor hint `tableWithDetails` should be changed to "Requirements Table Details", avoiding the duplicate name with hin `table`.

kbirken:
This PR provides some extensions for the `org.iets3.req.core` language. Some of these extensions are needed for variability implementation, some are generically useful. See ticket #1495 for details. 

All user-facing changes are explained in CHANGELOG.md.

Closes #1495.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ The project does _not_ follow Semantic Versioning and the changes are documented
   - Configurations might become inconsistent due to errors during manual conflict merges. A couple of model checks have been added to detect this. Additionally, there are quickfixes to fix such errors.
   - The internal storage of configurations has changed, this requires a language migration. Note that after the execution of the migration, extended configurations must be adapted via intention to their changed base configuration.
 - A VCS merge hint has been added for the `__hash` property (e.g., for variant configurations). This avoids merge conflicts which cannot be resolved manually anyway (in those cases, the hash value has to be recomputed anyway).
+- Requirements modeling (language `org.iets3.req.core`)
+  - It now supports a "requires" relation, which can express that a requirement needs other requirements as a precondition.
+  - The requirements chunk provides a flag "hide empty child requirements sections" (in the inspector). If selected, the flag hides empty child requirements sections in the tabular requirements view.
+  - Some additional internal changes (e.g., implementation of `ICanHide` interface, needed for variability support).
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


arimer (approved):
The implentation looks fine. Maybe some small quetions:

The property`hideEmptyChildReqsSection` is at the moment part of the `RequirementsChunk` concept. 

_Does it semantically make sense to keep it as a property?_

To me it looks like it is only required to manipulate the editors (show/hide cells) and might not belong (semantically) to the chunk itself.  If you just want to manipulate the editor  we could still use userobjects. In case you want to persist that setting (per MPS user and not in the AST) we could use [PropertiesComponent](http://127.0.0.1:63320/node?ref=r%3Ad65c70f5-6c40-4fc7-beaa-ee828f4fe514%28de.itemis.mps.linenumbers.plugin%29%2F7701440471669375886)

However I can't judge about the semantics of that chunk, therefor I am approving.

kbirken:
Thanks for review and approval!

> The property`hideEmptyChildReqsSection` is at the moment part of the `RequirementsChunk` concept.
> 
> _Does it semantically make sense to keep it as a property?_

Good point, it's not 100% clear if it should be a property. However, as the value of the flag is per-chunk and will be linked to the chunk's lifecycle, it likely makes sense to persist it (per chunk). So in an early stage when the chunk is edited a lot, it will be switched off (default `false`), and in later stages (when the chunk is "frozen"), it will be switched on to hide all the clutter (one extra line for each leaf requirement). But as it is not clear yet from real-world projects how the lifecycle should be expressed in the model, I decided to just name the flag according to its impact. We might rename it later.

PR #1497, opened 2025-10-22
fix DecTab bug not respecting the default value

### LINKED ISSUE #1494: Use of `DecTab` default value results in `AlternativesException`
Using the default value of a [`DecTab`](http://127.0.0.1:63320/node?ref=r%3A6bb59b1e-6116-48ad-b11d-2641d4f6b6a1%28org.iets3.core.expr.util.structure%29%2F4214990435115877128) does not work as intended. Instead an `AlternativesException` is thrown every time.

The reason is the wrong location of the default alternative, which is generated outside of the case handling.

Image

Image

nkoester:
This PR fixes #1494 and also adds tests to ensure the correct functionality of the change

The new tests:
image

The new checking rule:
image


arimer:
I've triggered the build manually on TC. Your branch name is not conform to the VCS Trigger settings.

arimer:
It looks like migrations are still pending. Do you mind running a `./gradlew migrate remigrate` locally and commit its results?

arimer:
How about we cherry pick the changes to maintenance/mps20232?

nkoester:
will be re opened for `maintenance/mps20232`

PR #1507, opened 2025-11-12
Bundled deps as provided in pom

sergej-koscejev:
* Add bundled dependencies as 'provided' in pom.xml to help with automated dependency scanning.
* Remove `jarNameOverride`, specify the type of takari-cpsuite artifact explicitly.


arimer:
The build never got picked up by TC due to the branch name.. I triggered it

PR #1534, opened 2025-12-01
Update supported MPS versions information

arimer:
The following versions of the IETS3 platform running out of maintenance and will no longer be actively maintained:
- MPS 2022.3
- MPS 2023.2

Therefor the readme is updated on master.


arimer:
> We should probably also update the table by removing lines 31-33.

Done. Only informaiton about latest supported versions stays

Issue #1521, opened 2025-11-25
Confusing solver warning is shown on TestCases after evaluation


arimer:
The [following warning message ](http://127.0.0.1:63320/node?ref=r%3A452e57fa-fd4c-45a8-b9ba-964614a5a66e%28org.iets3.variability.base.behavior%29%2F8172390074243787550) is shown on testcases that were previously evaluated via the interpreter. It looks like [check_ICanRunCheckManually](http://127.0.0.1:63320/node?ref=r%3Ae5a2b77c-569f-4c13-8679-6ec5a6000fa9%28org.iets3.core.base.typesystem%29%2F3025732926363274907) is creating it. This is confusing for users, because the solver is not relevant for running the interpreter. 

<img width="564" height="209" alt="Image" src="https://github.com/user-attachments/assets/8acfac9d-c7fe-4fec-9daf-e270a09a9787" />

How to reproduce:
- Navigate to a [test suite](http://127.0.0.1:63320/node?ref=r%3A109d7958-4308-4f2a-95cb-ea4731803298%28test.in.expr.os.algebraic%40tests%29%2F5955298286240026862)
- right click and execute `Run/Evalutate/Check All in Root Manually` 
- <img width="579" height="84" alt="Image" src="https://github.com/user-attachments/assets/0dae53b0-0486-47e2-8404-1d94e0d5787a" /> 
- Investigate warning

PR #1537, opened 2025-12-02
Bugfix/confusing solver warning#1521 open 

dbinkele:
solves #1521 

I added a flag to 'ICanRunCheckManually' to switch of the mentioned warning (as the default value is false).
For FeatureModel, FeatureModelConfig and FeatureDecTab it is true.
But for those concepts the checking_rule **(http://127.0.0.1:63320/node?ref=r%3Ae5a2b77c-569f-4c13-8679-6ec5a6000fa9%28org.iets3.core.base.typesystem%29%2F3025732926363202325)**

is not used to propagate the warning. FeatureDecTab is solved 'i typesystem' not manually.
The other two trigger an asynchronous task which does not trigger the rule. 
Thus. there is only one test ensuring that for an ISolvable an warning is not issued.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -369,7 +369,30 @@ fun MavenPom.includeAdditionalInfo() {
     }
 }
 
-fun MavenPom.addBundledDependencies() = bundledDependencies.forEach { addDependency(it, "provided") }
+fun MavenPom.addBundledDependencies() {
+    val seen = mutableSetOf<ResolvedDependency>()
+    val queue = ArrayDeque<ResolvedDependency>()
+
+    // Visit each bundled dependency, including its transitive dependencies if so configured, so that the exact set
+    // of used JARs ends up in the POM as provided dependencies.
+    for (config in bundledDependencies) {
+        queue.addAll(config.resolvedConfiguration.firstLevelModuleDependencies)
+
+        while (!queue.isEmpty()) {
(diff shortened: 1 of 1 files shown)


arimer:
Could you please a run a `migrate remigrate` and push the results afterwards?

kbirken (changes requested):
Here are some findings from my review:

- The method `ICanRunCheckManually.highlightWarning()` should be overridden (with body `true`) for `ICheckableTabularVarPoint` instead of `FeatureDecTab` (this is more general).
- The method `ICanRunCheckManually.highlightWarning()` should also be overridden (with body `true`) for `IVariabilityAwareArtifact`.
- Findings in `Check_ICanRunCheckManuallyUtil`
  - The debug outputs in `run()` should be removed.
  - After moving `run()` into the class the other two static methods can be private. For the calls to these methods I would prefer using intention `ConvertLocalStaticMethodCall`, this is more readable.
  - The class is used only from the typesystem aspect. You could move it there.
  - Optional: After it is moved to the typesystem aspect, you could add a "CheckingMethod" annotation (see below, use the annotation from `jetbrains.mps.lang.typesystem.dependencies`). Then you can call error, warning, info directly without passing the lambdas. (But then the test won't work without changes). Remark: I am not sure if CheckingMethods work for static methods.

image

dbinkele:
@kbirken can you continue?

Issue #688, opened 2023-07-21
Avoid colon casts in type system


alexanderpann:
When the type system expects a certain concept, we should not use the colon cast (:) to cast nodes to a different type but rather use the `as` cast. This cast returns null if the cast is not successful, for example when a user provides a node of a wrong type or the `RuntimeErrorType` is returned. I've fixed a few of such issues in the past but there are so many hidden cases, it's probably better to just ban this cast and replace all instances in the type system.

Example exception:
```
jetbrains.mps.lang.smodel.generator.smodelAdapter.NodeCastException: Can't cast node: 5190981625519587906, concept: org.iets3.core.expr.simpleTypes.structure.NumberType to concept: org.iets3.core.expr.collections.structure.CollectionType
	at jetbrains.mps.lang.smodel.generator.smodelAdapter.SNodeOperations.cast(SNodeOperations.java:559)
	at org.iets3.core.expr.collections.typesystem.typeof_AsListOp_InferenceRule$1.run(typeof_AsListOp_InferenceRule.java:28)
```

HeikoBecker:
I am not sure whether replacing colon casts with `as` is a sufficient solution to these kind of errors.

To me they are more of a symptom than an actual problem:
If I see a `NodeCastException` for a `RuntimeErrorType` in the type system, this means that the typesystem code does not encode enough information about **negative cases**, i.e. the rule in question is not robust against errors in the underlying expressions.

In the long run, replacing a colon cast with `as` might even cause more problems, because now the type system code will also have to be handle the cases where the cast fails.

I see two alternative solutions here:

- Follow the approach of a strongly-typed language and treat `RuntimeErrorType` as an error. Hence if any argument type is `RuntimeErrorType` the overall type should also become `RuntimeErrorType`
- Make `RuntimeErrorType` a sub-type of all KernelF types and silently allow the type system to continue type checking in a "best effort" manner.

Personally, I would favor the first solution, but this is up to discussion.

nkoester:
I also came across this issue in a customer project. It is quite unfortunate that such low-level exceptions end up with language users. They have no idea what any of this means and why it happened. Yes, their model is faulty but shoving a `NodeCastException` into their IDE does not help at all. I would expect the project to have their own error reporting for faulty models in this case.

[...]

PR #1550, opened 2025-12-05
#688 Avoid colon casts in type system

NilsWeidmann:
When the type system expects a certain concept, we should not use the colon cast (:) to cast nodes to a different type but rather use the as cast. This cast returns null if the cast is not successful, for example when a user provides a node of a wrong type or the RuntimeErrorType is returned. I've fixed a few of such issues in the past but there are so many hidden cases, it's probably better to just ban this cast and replace all instances in the type system.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -369,7 +369,30 @@ fun MavenPom.includeAdditionalInfo() {
     }
 }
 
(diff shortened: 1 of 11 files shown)


nkoester:
fixes #688

arimer:
@nkoester what was the motivation for not only converting TS rules but also: quickfixes, helper classes, behaviour methods etc.?

arimer (approved):
let's go 🚀

Issue #1552, opened 2025-12-08
Variability: Improve handling of root constraints with grouping


kbirken:
Root constraints in feature models can be clustered in groups. The feature model editor shows these groups and allows to added new constraints, all root constraints are stored in a linear collection.

- [x] Since new root constraints have to be added quite often, this leads to merge conflicts because new constraints are usually added to the end of the collection and such merge scenario with new elements in the list at the same index as for now cannot be automatically handled by MPS.
In order to reduce the probability of merge conflicts, the insertion of a new constraint should not be done at the end of the whole collection, but should be done after the previous constraint in the same group, or in front of the first constraint of the group if "insert before" has been triggered by the user (Shift-Return). If on the two merged branches constraints have been added for the different group, there should not be a merge conflict at all. 
- [x] The list of root constraints is additionally shown in the inspector. This is not necessary and should be removed.
- [x] In the merge view of MPS, the editor based on query-lists should be replaced by a vertical collection of all root constraints (as stored by persistence). For each constraint, its group should be shown (if any). This should allow doing merges if the group assignment has been changed for one constraint. The group name should be colored according to the color defined for by the group's definition.

PR #1574, opened 2025-12-18
1552 variability improve handling of root constraints with grouping

TorstenGoerg:
- Editor improvements for feature model root constraints.
- Added repositioning of feature model root constraints on constraint group assignment changes.

This solves #1552.


kbirken:
Starting review. Functional testing looks good so far. First finding:

In the merge editor the editor for a constraint should be more compact. Currently it looks like this:

image

A more compact editor would be:

```
group1: feature1 && feature2
```

I.e., the group name (colored), colon, space, constraint

kbirken:
UX finding: Placing the cursor at the end of a constraint in a group and pressing return will create a new constraint line directly below. However, this does not work for the last constraint in the group.

TorstenGoerg:
> But the simplified syntax that you suggest collides with the extensions or default "c:" prefix of the expression constraints. So we would end up with `group1: c: feature1 && feature2` which does not look good. What about the syntax `group1 { c: feature1 && feature2 }` ?

Further possibilities would be `[group1] c: feature1 && feature2` or just `group1 c: feature1 && feature2`.

PR #1581, opened 2025-12-22
add missing dependencies in sbom

mhuster23:
The `languageLibs` and `mps` configuration were missing from the sbom.


arimer:
I think we need to add those featuers for each individual maintenance branch that is still under development. Do you mind porting those changes from 2024.1 -> master?

arimer (changes requested):
Please change the targetbranch to 2024.1 so that the jbom is a well fixed for older maintenance branches.

Issue #1593, opened 2026-01-14
The type of a number with prefixed physical unit should not always have infinite precision


kbirken:
The typesystem of the current unit language `org.iets3.core.expr.typetags.physunits` (see [documentation](https://mbeddr.com/mps-platform-docs/languages/kernelf/tour/solutions/KernelFTour/screenshots/PhysicalUnitsLanguage/PhysicalUnitsLanguage/)) computes the wrong precision for numbers which have a prefixed physical unit. Currently the precision of the computed number type is always hardcoded as infinite precision.

This causes type errors when the type is checked against some other type. Examples:

<img width="1150" height="146" alt="Image" src="https://github.com/user-attachments/assets/e90929ef-d3b8-41d6-87f6-8969fdef0cf3" />

This currently forces the to-be-matched type to have infinite precision as well:

<img width="487" height="66" alt="Image" src="https://github.com/user-attachments/assets/5a0f7cf8-033b-4316-9c70-0314765626b1" />

Expected: The precision of the resulting number type should be as precise as possible (pun intended). Then the to-be-matched type can restrict the expected precision and the normal subtyping rules for number types apply.

PR #1594, opened 2026-01-14
Fix wrong infinite precision for number types with prefixed units

kbirken:
This PR extends  some tests for physunits to reproduce the problem of infinite precision for types of prefixed units. It provides the fix, and updates the CHANGELOG accordingly.

This solves #1593.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


wsafonov:
> Which units did you use here? There are three ones in the standard library.

Hmm, in my case there was only one bit unit available, namely the metric one, which is the correct one since we are expecting a metric-scaled value. 

image

Does it behave differently on your side at the same location?

kbirken:
> Important points are all addressed, looks good now.

Will squash and then merge.

Issue #1601, opened 2026-01-19
Variability: Make viewer tool for skeleton trees more robust


kbirken:
Skeleton trees are the central data structure for many variability algorithms. There is an action "Show skeleton tree" and a corresponding view-tool, which shows a graphical representation of a skeleton tree.

<img width="890" height="250" alt="Image" src="https://github.com/user-attachments/assets/a61e6f6d-700c-4baa-9905-d3dc38fc485a" />

There are a couple of small problems with this view which have to be fixed:

- For feature dectabs, the default expression might be a KernelF `StringLiteral`. Then, the skeleton node generates a label which contains double quotation marks in a string which is limited by double quotation marks. This breaks the graphwiz graph algorithm. 
- Error output from the graphviz execution should be detected and reported to the user.
- If an undetected error during graphviz execution occurs, the view shows the image (svg format) which worked before - this might be a totally unrelated skeleton tree.

PR #1602, opened 2026-01-19
Fix various problems of skeleton tree viewer

kbirken:
Details:
- Labels for tree nodes will not break due to double double quotes now.
- Before generating a new svg file, an old one will be deleted. Thus, if the generation does not work, the old one will not be shown instead.
- Errors from graphviz execution are now detected, handled and reported (in log, but also in messages view).
- A superfluous stderr output on isApplicable has been removed.

CHANGELOG has been updated.

Note that an IllegalStateException is thrown on the first few executions of "Show Skeleton Tree". But this seems to be a problem in MPS (actually, how the action opens the viewer tool, and some kotlinx coroutine magic). No application code is involved, so I could not fix it. Despite of the exception everything works fine.

This is a bit hard to test, as the generation of SkeletonTree graphs is usually quite stable. In order to try the error handling and reporting one could change the code of `SkeletonTreeViewer` like this:

image

Then the file specifying the graph is broken in the first line and the error reporting kicks in.

This solves #1601.


dbinkele:
@kbirken looks good, but what problem are you solving? Is it possible to add a TestCase where the 'old code' failed?

PR #1633, opened 2026-02-09
Update MPS to 2024.1.6

sergej-koscejev:
MPS 2024.1.6 has stricter checks for plugin dependencies. Dependencies were already adjusted in #1624.


kbirken (approved):
Update to 2024.1.6 should be fine.

PR #1634, opened 2026-02-09
variability: remove dependency on the tracing plugin 

sergej-koscejev:
Use direct string editor hint to avoid dragging in the tracing plugin with its potentially slow conditional editors.


sergej-koscejev:
This PR includes #1633 which should be merged first.

PR #1705, opened 2026-03-17
fix JNA path construction for interpreter testExecutor

nkoester:
Fixed a bug in the `execTestsByInterpreter` task which would result in a wrong JNA path:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## March 2026
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


nkoester:
This PR avoids running into [MPS-39518](https://youtrack.jetbrains.com/issue/MPS-39518) by setting `jnaLibraryPath` as part of the `runMPS`.

arimer (commented):
Changes look fine. Some minor questions:

1. How did you figure out that that this changes fixes the JNA issues?
2. Why do we need to remove the property `build.jna.library.path` as well?

nkoester:
1. With the help of @slisson I could find the [corresponding MPS code](https://github.com/JetBrains/MPS/blob/65e95a4da8d14898f4fbebc5cf2f02f9a5521925/core/tool/ant/source_gen/jetbrains/mps/build/ant/MpsLoadTask.java#L262-L265) which sets `jna.boot.library.path` in the `MPSLoadTask` ant task. This overwrites the correct value.
2. The other paths (including the one from the properties) are overwritten by the one set in the `runMPS` and are thus not needed anymore

arimer (approved):

Issue #1713, opened 2026-03-23
Variability: Feature model root editor quirk


kbirken:
**Creating a root for a feature model gives unexpected property constraint violation**
The following error is given when creating a root only:
<img width="357" height="57" alt="Image" src="https://github.com/user-attachments/assets/2383ec7c-f153-48fe-bb96-3deb06554735" />

**To Reproduce**
1. Create a new VariabilityModelChunk
2. Create a new feature model
(3. If you add a feature under the root, the error disappears)

**Expected behavior**
No vague name constraint given, but instead one of the following (I don't know what makes sense because I don't know why that error is given):
 - either: no error given
 - or: an appropriate error given (if that makes sense), e.g.: "no feature models with only one feature allowed"

**Environment (please complete the following information where applicable):**
 - OS: Windows 11
 - MPS Version:
<img width="333" height="155" alt="Image" src="https://github.com/user-attachments/assets/fa7039c7-d7d6-4908-bb20-9c17e5c5ff09" />

Note: This is a duplicate of IETS3/iets3.core#1726. The proble has to be fixed here, so we are transferring the issue.

arimer:
What is the root concept for the feature model?

dbinkele:
> What is the root concept for the feature model?

@arimer 'VariabilityModelChunk'

kbirken:
> > What is the root concept for the feature model?
> 
> [@arimer](https://github.com/arimer) 'VariabilityModelChunk'

No, it is `FeatureModel`.

kbirken:
PR has been merged to mps20241 successfully, closing.

PR #1714, opened 2026-03-23
Fix error for name of newly created feature model

kbirken:
The root feature of a newly created feature model deliberately lacks a name, thus an error is shown. This is correct.

However, the name of the feature model itself is derived from the root feature's name via a get-property constraint. As both the feature model concept and the root feature concept are implementing `IValidNamedConcept`, the non-existent name of a new root feature will cause a second error for the name of the feature model. This second error will not vanish automatically after some valid name is entered, only after F5 is pressed additionally. 

This PR fixes the problem by assigning a specific `NO_NAME` name to feature models without root feature or with root features with an invalid name.

This solves #1713.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - Fixed a bug in the `execTestsByInterpreter` task which would result in a wrong JNA path
+- Variability: Newly created feature models will not show a "Property constraint violation" error anymore.
 
 
 ## February 2026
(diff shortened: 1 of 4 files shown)


kbirken:
Regarding @dbinkele 's offline question about using `StringUtils.isEmpty` vs. `StringUtils.isBlank`: The condition should apply specifically when the feature model is created. As soon as the user enters one or more spaces themselves, it is their responsibility, and the error message may then be displayed at the feature model. So we should use `isEmpty` rather than `isBlank`.

dbinkele (dismissed):

dbinkele (approved):

PR #1726, opened 2026-03-25
Variability: Fix bug with "Adapt to extended configuration"

### LINKED ISSUE #1724: Variability: "Adapt to extended configuration" intention behaves strangely
MPS 2024.1

How to reproduce:
* Create an abstract configuration C with 2 FeatureAttributes and give them values
* Create a non-abstract configuration D that extends C
* Execute the "Adapt to extended configuration" intention on D
* Now you will see that one of the two FeatureAttributes is set to inherited and the other is set to unset
* Execute the intention once more on D
* Now the one attribute is manual and the other is inherited
* Execute the intention once more on D
* Now both attributes are manual

I would expect that upon the first execution of the intention both attributes are inherited.

Also: upon first execution, there is no difference between the behavior of the intention and the behavior of the quickfix (and after the first execution the quickfix is of course not available anymore).

dbinkele:
fixes #1724

### DIFF
--- a/.github/workflows/vulnerability-scanning.yml
+++ b/.github/workflows/vulnerability-scanning.yml
@@ -13,7 +13,7 @@ jobs:
           distribution: temurin
           java-version: 17
       - name: Setup Gradle
-        uses: gradle/actions/setup-gradle@39e147cb9de83bb9910b8ef8bd7fff0ee20fcd6f # v6
+        uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v6
       - name: Call setup
         run: ./gradlew dependencies
       - name: Dependency Check
@@ -27,7 +27,7 @@ jobs:
           format: 'HTML'
           out: 'reports'
       - name: Upload Test results
-        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
            name: Dependency Check Report
            path: ${{github.workspace}}/reports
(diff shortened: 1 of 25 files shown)


kbirken (changes requested):
Here are my findings:

- The problem described can still be reproduced. When running the "Adapt this configuration to the extended configuration" intention on the extending configuration, the attributes are set to manual (although they should be `inherited`). This happens independent of the owning feature in the abstract config being selected or untouched.
- In the variability sandbox, you killed the `HomeNetworkFeatures` examples. Please re-instantiate it. All references in the `Configuratons` chunk are currently broken and the feature models `HomeNetwork` and `Power` have been deleted.
- CHANGELOG should be updated.

Pls. also fix dirty files, squash commits and finally merge from `mps20241`.

dbinkele:
@kbirken Please have a look once more. If all is OK, I can squash the whole thing

kbirken:
> * The problem described can still be reproduced. When running the "Adapt this configuration to the extended configuration" intention on the extending configuration, the attributes are set to manual (although they should be `inherited`).

This is fixed now. However, if after this test the value of an attribute in the abstract config is being changed, then the same attribute in the extending configuration will not get the new inherited value (neither automatically nor by running the "Adapt..." intention). This can be tested with the `AttrPlay` sandbox model.

This issue wasn't mentioned in the ticket, but probably it hasn't been visible before due to the previous problem. But I believe it should also be fixed in this PR.

dbinkele:
@kbirken your finding has been addressed

kbirken:
I found one more logic problem: In the following code, the yellow part should not check the assignment for null, but the _value_ of the assignment. The assignment itself is always notnull.

image

I fixed this and did some refactorings (make code more readable and shorten some names as well).

kbirken (approved):
Looks good now. I fixed some remaining logic problem and now everything should be fine.

PR #1730, opened 2026-03-27
fix: hex support on 2025.1

### LINKED ISSUE #1715: Hexadecimal support broken in 2025.1
The implementation of hexadecimal numeric literals support uses reflection to change the value of a final field (`EditorCell_Property#myModelAccessor`) and this is not allowed in Java 21 which MPS 2025.1 uses.

sergej-koscejev:
Fixes #1715.

Introduce a new language for replacing model accessors in property cells.

Also a few minor build tweaks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 
 - Fixed a bug in the `execTestsByInterpreter` task which would result in a wrong JNA path
-
+- Hexadecimal number support now works for MPS 2025.1 and above.
 
 ## February 2026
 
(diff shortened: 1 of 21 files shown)


arimer:
Looking into it now.

arimer:
Do you think it makes sense to move the new language to the ‘mps-extensions’ ?

arimer (approved):
Changes look good. Thanks for adding the additional editor tests and cleaning up the build 👍🏼

sergej-koscejev:
> Do you think it makes sense to move the new language to the ‘mps-extensions’ ?

I thought about this but decided against it because overriding property accessors is only necessary for the current design of IETS3 hex literals and doesn't feel like an approach that we should encourage other projects to follow.

Issue #1731, opened 2026-04-01
Variability: Switch off automatic run of solver for tabular variation points


kbirken:
Currently the solver check for tabular variation points (`ICheckableTabularVarPoint`) and esp. for `FeatureDecTab` nodes runs as a checking rule. Thus, it is executed automatically when the model is changed in the editor and also when the model check is executed on an MPS model or an MPS solution. 

However, there are several known shortcomings with this. First, problems with automatic execution when the model is changed:
- Sometimes the check is executed only after pressing F5, see IETS3/iets3.core#1613.
- If the root node contains several tabular varpoints, then executing the check on one of them might remove all previous error markers on another.
- When the preview is active, the solver run probably doesn't work (to be checked again, see IETS3/iets3.core#1619).
- The MPS generator is not very robust (although we improved it a lot over the last 1-2 years), so if an error during generation occurs, the users will encounter annoying generator error in the message view. 

Next, problems with executing the check during model check:
- The execution is done asynchronously, and the errors are not collected and shown in the model check result. Thus, executing the checks is useless at the moment.
- Due to the asynchronous execution (where all feature dectabs are processed sequentially, but in a different thread), for models with many (several hundreds) of feature dectabs the execution will last several minutes after the model check seems to have finished. This leads to an unresponsive UI in the MPS IDE. 

As these problems cannot be fixed quickly, we should deactivate the automatic execution of the solver check for feature dectabs and other `ICheckableTabularVarPoint ` concepts. The user has to run the check explicitly then, by using `Run/Check manually` from the context menu.

kbirken:
PR has been merged to mps20241, closing successfully.

PR #1732, opened 2026-04-01
Variability: Switch off automatic solver checking of tabular variation points

kbirken:
The method `solvedInTypeSystem` now returns false. See rationale in the ticket #1731 and in javadoc comment.

CHANGELOG has been updated as well.

This solves #1731.


wsafonov (approved):
Approved, let's ship the hotfix.

Issue #1609, opened 2026-01-26
Variability: Add checking rules for feature attributes with no not-present values


kbirken:
If a feature attribute does not have a not-present value, the concrete value of the attribute is undefined if the owning feature is disselected in a configuration (aka false). Thus the user should be warned if this attribute is used in some expression, as the expression cannot be evaluated if the owning feature is false.

Typical expressions were attributes are referenced are:

- constraint in feature model
- default expression of another attribute in feature model
- presence condition in artifact
- left-hand-side condition in an `ITabularVarPoint` (e.g., a `FeatureDecTab`).

For the first case a check with a warning is already implemented. For the other three cases this should to be provided with this ticket.

dbinkele:
done #1735

PR #1735, opened 2026-04-08
Feature/add checking rules for feature attributes with no not present values #1609

dbinkele:
solves #1609


mfsoliveira (approved):
Warning are added for cases where no default value is present for attributes when a feature containing the attribute is not selected.
it is ok

PR #1737, opened 2026-04-13
Support for enforcing vertical layout of enum declarations

### LINKED ISSUE #1736: Vertical layout for non-valued enum declarations
Currently, enum declarations use a vertical grid layout only when literals are assigned values (i.e., valued enums). For simple enums without assigned values, literals in enum declarations are displayed inline - even when the list spans multiple lines.

This can reduce readability for enum declarations with a large amount of literals. It would be helpful to provide an option to enforce a vertical layout for enum declarations, regardless of whether values are assigned.

wsafonov:
Fixes #1736

### DIFF
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/intentions.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/intentions.mps
@@ -41,6 +41,9 @@
       </concept>
     </language>
     <language id="f3061a53-9226-4cc5-a443-f952ceaf5816" name="jetbrains.mps.baseLanguage">
+      <concept id="7485977462274819189" name="jetbrains.mps.baseLanguage.structure.FormatOperation" flags="ng" index="2cAKMz">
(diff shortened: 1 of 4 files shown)


kbirken (dismissed):
Looks good and works nicely, approved. One proposal though: There is a policy for capitalization in descriptions of intentions:

```
Warning: Naming policies violated: all words except prepositions, articles and particles should be capitalized.
```

Maybe the new intention should adhere to this policy (although the warning is not displayed because of the `.format(...)`).

wsafonov:
> Looks good and works nicely, approved. One proposal though: There is a policy for capitalization in descriptions of intentions:
> 
> ```
> Warning: Naming policies violated: all words except prepositions, articles and particles should be capitalized.
> ```
> 
> Maybe the new intention should adhere to this policy (although the warning is not displayed because of the `.format(...)`).

Thanks for the hint, agree - adjusted the intention name according to the policy.

kbirken (approved):
Looks good now, thanks for improving the intention description.

Issue #1709, opened 2026-03-18
Variability: Restore workaround for MPS-34340


sergej-koscejev:
The problem still appears in MPS 2025.1 and since exceptions thrown from checking rules are now properly reported ([MPS-37044](https://youtrack.jetbrains.com/issue/MPS-37044)), it is blocking the migration of a customer project to 2025.1.

There used to be a workaround for this issue in iets3.core which was removed in https://github.com/IETS3/iets3.core/commit/db21d97c54327ac0b43a150f38647a7a303fbadb, it should be restored.

kbirken:
This problem occurs in the following use case: The application uses the "for-all-variants" language extension to execute specific checking rules for each 100% variant of an 150% artifact model. The artifact model is filtered for each variant configuration in turn and the checking rule is applied. All reported errors and warnings will be specific for a given variant configuration.

Due to the problem in MPS-34340, type system information is lacking during execution of model checks (see details in the ticket). Thus for variability artifact models with multiple roots the filtering to 100% cannot be executed properly. The workaround requested here will check for the specific situation and deactivate the "for-all-variants" execution of checking rules in this case.

kbirken:
PR for mps20241 has been merged successfully, closing.

PR #1738, opened 2026-04-13
Restore workaround for MPS-34340 in variability "for-all-variants" extension

kbirken:
This restores the workaround in the "for-all-variants" runtime needed for MPS-34340. In mps20241, this problem was hidden because exceptions in checking rules were not reported, and so the removal of the workaround didn't cause any visible issues. In mps20251, exception are thoroughly handled, so the workaround is needed again.

Notes:
- The solution `org.iets3.variability.artifacts.typesystem.runtime` which hosts the workaround has been moved from IETS3.Core to IETS3.Opensource in 2025. So the workaround could not be just restored, but had to be re-added by copy&paste.
- There is a test in IETS3.Core (`TwoActualRoots`), which checks that the workaround is working correctly. This test needs stuff from IETS3.Core, so it could not be migrated to IETS3.Opensource together with the `org.iets3.variability.artifacts.typesystem.runtime` solution. Thus, we need a companion PR in IETS3.Core to adapt the test. 

This solves #1709.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## April 2026
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


dbinkele (dismissed):

kbirken:
Now we have the same workaround for mps20241 in PR #1762, thus I am closing this one. We will provide the workaround for mps20251 via the usual cascading merges.

PR #1743, opened 2026-04-14
Allow to modify interpreter report file name prefix

nkoester:
Adds getter and setter to [`myReportsFilenamePrefix`](http://127.0.0.1:63320/node?ref=r%3A3b363c34-a1ec-4ae6-92cc-95c0273ad0c2%28xml4JUnit%29%2F7174873163069082444) in `TestExecutorConfigForCommandLine`.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
   - default expression of another attribute in feature model
   - presence condition in artifact
   - left-hand-side condition in an ITabularVarPoint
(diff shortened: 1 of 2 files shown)


arimer (changes requested):
- I think we already have a getter in that class. Only the setter is missing.

Image

- How about we add your proposals to the [ICustomRunnerConfig](http://127.0.0.1:63320/node?ref=r%3A3b363c34-a1ec-4ae6-92cc-95c0273ad0c2%28xml4JUnit%29%2F6962548927774890821) interface? 

- Could you also please update the changelog?

arimer (approved):
Let's go. 🚀

PR #1744, opened 2026-04-14
fix: interpreter of `StringInterpolationExpr` added additional spaces

slisson:
It produced additional spaces around instances of `InterpolExprWord`. The main issue is is that the produced string doesn't match what is shown in the editor.

Even if this was intended, there would be no way to specify a string that doesn't have spaces around `InterpolExprWord` instances. That's why this has to be a bug. If the user wants these spaces, they should be typed explicitly.


dbinkele (approved):
LGTM

arimer:
Please try to keep the changelog up to date

PR #739, opened 2023-12-11
Cascading merge from 2021.3 to 2022.2 + upgrade to MPS 2022.2.1

alexanderpann:

PR #1760, opened 2026-04-21
Bugfix/creating a feature should be the default in the feature diagram editor #739

dbinkele:
solves IETS3/iets3.core#739


kbirken (changes requested):
When a new node in the feature model tree is created, it is always a `Feature` now. The ticket says

> switching a Feature to a FeatureInclude should be done differently (e.g., via pressing :)

Unfortunately, pressing `:` after the feature name does not convert it into a `FeatureModelInclude` anymore (the ticket still assumes this is working, but we obviously had a regression). Now the only way to create a `FeatureModelInclude` is by intention. Pls. fix the regression, so that pressing `:` converts a `Feature` to a `FeatureInclude`. Also add a test so that functionality does not get lost anymore.

kbirken (approved):
I updated the CHANGELOG and fixed an old parametric intention which is used for switching between Feature and FeatureModelInclude nodes (it basically just changed the type, but lost all other information like cardinality and constraints, which is bad). 

The PR looks good now, approved.

PR #1764, opened 2026-04-27
enhanced compatibility between number and int/real types

wsafonov:
Currently, `real` and `int` types have correct subtyping rules for the number type allowing to assign `number` with precision zero to `int` and any number to `real`, but not vice versa, i.e. assigning `int` to a `number` or `number{0}` is not allowed and also assigning `real` to `number{inf}` is not supported yet. This PR adds additional typesystem support for those cases.


kbirken (approved):
LGTM. The new typing logic seems valid to me, the repacement rules are properly implemented and the new tests cover all cases. Thanks!

Issue #1774, opened 2026-05-04
Workaround for MPS-34340 does not work inthe presence of logical children.


mfsoliveira:
Given a checking rule using the operator 'for all variants' which require multiple roots to execute the check, it fails due to bug https://youtrack.jetbrains.com/issue/MPS-34340 in MPS.

A workaround was implemented for it and re-integrated in iets3.os: https://github.com/IETS3/iets3.opensource/pull/1762

After testing it in a downstream project, I observed that the workaround does not work if the checking rule checks nodes that contains logic child.

In the workaournd code we can see that the artifactGroupLogical is not used the condition to prevent the check to run.
Only the roots queried from IVAAUtil.artifactGroup(ivaa) is used.

PR #1775, opened 2026-05-04
Consider logical child in mps34340 workaround in for all variants

mfsoliveira:
Use artifact group including logical children in the condition that test if it run multi check, so that it works around a bug existing in MPS.

solves  #1774


kbirken (approved):
The fix looks good, great that you found the problem. When I copied the workaround code from an older branch, I didn't take into account the addition logical children (which hasn't been supported on this older branch earlier than end of 2024).

Thanks also for updating the CHANGELOG.

Issue #1784, opened 2026-05-12
Physical Units: typePrefix is not reset


dbinkele:
When editing an expression with a unit (i.e. 20 m) then on the type system being triggered the unit prefix is set.
It is stored in a user object. This happens in (http://127.0.0.1:63320/node?ref=r%3Ad4f1532d-fc5c-419f-84ee-daef42867c8e%28org.iets3.core.expr.typetags.physunits.typesystem%29%2F778041577207590326).

We have now the following error:

1. Define a variable 

<img width="634" height="164" alt="Image" src="https://github.com/user-attachments/assets/c02d24ea-6f38-46ae-be0e-970a9782e5ce" />

2. Now the unit prefix is set to 'c'
3. Change ONLY 'cm' to 'm'. Do not delete '20'.
4. The unit prefix is still 'c' which will cause error downstreams when UnitConversions are computed

PR #1785, opened 2026-05-12
reset UnitPrefix (1784)

dbinkele:
solves #1784


kbirken (dismissed):
LGTM, resetting the prefix-related info from typesystem was a bug that existed for a long time.

kbirken (approved):
LGTM.

Issue #1773, opened 2026-05-04
Exception in DiscreteDateRangeValue.end()


slisson:
`DiscreteDateRangeValue.end()` throws an exception if `.begin` already has a value of `LocalDate.MAX`.

arimer:
Closed with #1790

PR #1790, opened 2026-05-20
fix: `LocalDate.MAX.monthRange.end` caused an internal interpreter exception

slisson:
Fixes https://github.com/IETS3/iets3.opensource/issues/1773


arimer (approved):
Thanks for the changes. I've updated the changelog as well.

PR #1854, opened 2026-06-30
fix: unique name check missing for function arguments and local vars (#1828)

### LINKED ISSUE #1828: bug (org.iets3.core.expr.toplevel:Function): Unique name check missing for arguments and local vars
**Problem**
In case you have a [Function](http://127.0.0.1:63320/node?ref=r%3Ada65683e-ff6f-430d-ab68-32a77df72c93%28org.iets3.core.expr.toplevel.structure%29%2F4790956042240148643) with a parameter with the same name as a local variable inside that function, no error is show. See screenshot
on the left hand side:
Image

**Expected Behaviour**
An error similar as in BaseLanauge is show on the local variable inside a kernelF function.

dbinkele:
## Summary

- Adds a unique-name constraint in `org.iets3.core.expr.lambda` behavior so that a `Function` parameter and a local variable with the same name produce an error (matching the behaviour already present in BaseLanguage)
- Adds a regression test in `m1@tests.mps` covering the duplicate-name case

## Related issue

Closes #1828

## Test plan

- [x] Open the test model `m1@tests.mps` in MPS and confirm the new test case passes
- [x] Create a `Function` with a parameter and a local variable sharing the same name — verify an error is shown
- [x] Confirm no error appears when names are distinct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## July 2026
+### Added
+- IFunctionLike takes arguments and named body content into account when performing ab uniqueness name check
+
 ## June 2026
 
 ### Added
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.lambda/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.lambda/models/behavior.mps
@@ -2048,7 +2048,7 @@
                       <node concept="2OqwBi" id="S0JwotCVZv" role="2Oq$k0">
                         <node concept="13iPFW" id="S0JwotCVZw" role="2Oq$k0" />
                         <node concept="3TrEf2" id="S0JwotCVZx" role="2OqNvi">
-                          <ref role="3Tt5mk" to="zzzn:49WTic8eSDm" />
+                          <ref role="3Tt5mk" to="zzzn:49WTic8eSDm" resolve="body" />
                         </node>
                       </node>
                       <node concept="2Rf3mk" id="S0JwotCWV8" role="2OqNvi" />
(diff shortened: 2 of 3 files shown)


arimer (commented):
I guess both concepts need to be covered: 

Image

arimer (changes requested):
Image

- still not wokring for **var**
- please also update the changelog

dbinkele:
@arimer I enhanced the test making sure 'var' exposes the same behavior

arimer (changes requested):
Image

This is **not** working. We also need to take not only vals, but also vars in to account when checkeing for the name uniqueness inside a body of the IFunction.

dbinkele:
@arimer very strange 
image

arimer:
Well, there is a unique check for the boby across vars/vals and there is a unique check for all args. But there is no check that is doing both: checking that args and all vars/vals are unique to avoid shadowing.

dbinkele:
> Well, there is a unique check for the boby across vars/vals and there is a unique check for all args. But there is no check that is doing both: checking that args and all vars/vals are unique to avoid shadowing.

@arimer done

arimer:
Altough your solution is not wrong, I am wondering if it makes more sense to check the uniqunenesss of the body in the body itself. The body is always defined by a blockexpression. That expressions defines also the list of [unique names](http://127.0.0.1:63320/node?ref=r%3A06389a24-a77a-450d-bc88-bccec0aae7d8%28org.iets3.core.expr.lambda.behavior%29%2F1307222191605647272) how about we move it over there? What do youthink?

dbinkele:
> Altough your solution is not wrong, I am wondering if it makes more sense to check the uniqunenesss of the body in the body itself. The body is always defined by a blockexpression. That expressions defines also the list of [unique names](http://127.0.0.1:63320/node?ref=r%3A06389a24-a77a-450d-bc88-bccec0aae7d8%28org.iets3.core.expr.lambda.behavior%29%2F1307222191605647272) how about we move it over there? What do youthink?

@arimer the only code change I did is the following. If you move this to 'Blockexpression' you cannot reference the 'args' and hence you cannot check. If I misunderstood then let's have conversation.
image

arimer (dismissed):

arimer (approved):

PR #1868, opened 2026-07-15
deps: fix group id of mps artifact

arimer:
I forgor to update the group id of the mps artifacts after branching form master, which is using mps pre releases. The group id of the actual releases is the one added in this PR


nkoester (approved):
very nice

PR #1901, opened 2026-08-05
Bugfix: Function unique name checking

### LINKED ISSUE #1883: (bug)IFunctionLike: unique name check checking too much
In this [PR ](https://github.com/IETS3/iets3.opensource/pull/1854) we updated a uniqueness check that is doing too much at the moment.

**Example: Those errors should not be there**
Image

You can find the example on that branch: `bugfix/arimer/WrongFunctionUniqueCheck`

**Expected**

Uniqueness check of named function arguments is checked against all vars/vals in all bodies. Each nested body shall not have val/vars with duplicated names. Same names of Val/Vars defined in an outer body, can be used in the inner body. The example from above shall be error free.

nkoester:
Fix IFunctionLike unique-name check to respect nested scopes (fixes #1883).

image

Summary of changes:

- getUniquelyNamedElements() used getDescendants() unbounded, so a
function's uniqueness check picked up val/var declarations from
nested lambdas/blocks too, flagging legitimate name reuse in
independent inner scopes as duplicates.

- Restrict the collected elements to the function's own scope: a
val/var counts only if its nearest enclosing IContainerOfUniqueNames
is the function itself or its own top-level body block (the body is
itself a BlockExpression/IContainerOfUniqueNames, so both cases need
covering). Arguments shadowed by a same-scope val/var are still
caught; siblings in independent nested blocks no longer conflict.

- Adds a regression test (`vars`) covering both the original false
positives and the still-required true positives (arg/local shadowing,
same-block duplicates, nested-block redeclaration shadowing).


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## August 2026
+### Fixed
+- Improve uniqueness name check of `IFunctionLike` `getUniquelyNamedElements()` behavior to avoid overzealous checking.
```
arimer, line 10: 😍

arimer (commented):
So small findings:

1. Not sure if the `==` is the best choice here. How about we use :eq:? 

Image

2. How about the following case: Although I am not sure if the general impl. of the KernelF scoping is suppose to work as in java.

Image

arimer:
I fixed 1. and updated some example with the "check to be error free" annotation. 

For 2. I am not even sure if it worked before or not. Maybe a topic for later?

nkoester:
I would suggest to leave (2) as is in this PR for now. If we decide to change this we can do it later on

arimer (approved):

PR #1925, opened 2026-08-24
Fix NullPointerExceptions and contract violations in RecordValue

slisson:
`RecordValue` (org.iets3.core.expr.toplevel.plugin) threw `NullPointerException` from both `equals` and `compareTo` whenever a record member had no value, and its `equals`/`hashCode`/`compareTo` trio was mutually inconsistent. This fixes all three, with test cases in `test.ts.expr.os.records`.

## What changed

**1. `equals` threw an NPE on null member values** (`7ec5d2e4`)

`!rv.memberData[e.key].equals(e.value)` dereferenced the other record's member value, which is `null` for a member that is present in the map but has no value. Replaced with the null-safe `:ne:` operator, so it generates as `Objects.equals(...)`.

**2. `equals` and `hashCode` disagreed** (`030cab47`)

`hashCode` was `Objects.hash(memberData, recordDeclaration)` while `equals` compared only `memberData`. Two record values with equal members but different record declarations were therefore `equals` with differing hash codes — broken for `HashMap`/`HashSet`. `equals` now compares the record declaration as well.

The declaration is compared by node identity, which is what `IRecordDeclaration.equals` (`thisNode :eq: declaration`) already does for KernelF's own `==`. Since records are nominal, both operands of a well-typed comparison carry the same declaration, so this changes nothing for well-typed programs; it fixes the contract for values that reach a hash-based collection across record types.

**3. `compareTo` threw an NPE, and was not a valid `Comparable`** (`934b968f`)

- *No record declaration.* The interpreter builds `new RecordValue(null)` for inline/projected records. `SLinkOperations.getChildren(null, …)` returns empty, so control fell into the `effectiveMembers()` branch, which returns `null` for a `null` node, and the for-loop NPE'd. Such a record now compares by member name, over the sorted union of both sides' member names.
- *Null member values.* These were skipped entirely, which made the ordering intransitive: `{y:1}` vs `{y:null}` and `{y:null}` vs `{y:2}` both compared equal while `{y:1}` vs `{y:2}` did not — a `TreeSet` of the three held two elements. A new private `compareMemberValues` sorts a member without a value before one that has a value; two present values still go through `OH.compare` unchanged.

The `comparisonOrder` branch is unchanged in intent — it still compares only the declared sort key, which is the feature, so `compareTo == 0` remains deliberately weaker than `equals`.

## Tests

Three test methods added to `RecordValueTest` in `test.ts.expr.os.records`. Against the unfixed code the two `compareTo` tests reproduce the defects — one with the `NullPointerException` at `RecordValue.compareTo`, one with an `AssertionError` because `withNull.compareTo(one)` was `0`:

- `equalsIsNullSafeForNullMemberValues` — regression test for (1)
- `compareToWorksWithoutRecordDeclaration` — inline records: ordering, reflexivity, and differing key sets
- `compareToOrdersNullMemberValuesConsistently` — `withNull < one < two`, i.e. transitivity

All 4 tests in `RecordValueTest` pass, and the existing `records` test case (10 tests) is unaffected.

## For the reviewer

- The KernelF `==` operator uses **neither** method: `EqualsHelper.equals` has its own `RecordValue` branch. `equals`/`hashCode` matter for Java-side collections, `compareTo` for ordering (`OH.compare`, sorting, `TreeSet`/`TreeMap`). So the risk surface here is Java-side, not language semantics.
- Interpreter suites in `test.in.expr.os` that touch record ordering were run: `enums`, `enum_sort`, `EnumsWithValuesOfSortableTypes`, `references`, `records`, `path`, `grouping` pass. `projection` fails, but it fails identically with `compareTo` reverted (same four items, `Type system returned null type for child` and an `invalid null value detected` in `SumOp`) — pre-existing and unrelated.
- Separate finding, not addressed here: `EqualsHelper.equals` returns `false` for two structurally identical inline records, because it invokes the `IRecordDeclaration.equals` behavior method on a `null` node and gets the boolean default back. So `==` between two projected records is always `false`. Worth its own issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2026
+
+### Fixed
+- KernelF: `RecordValue` no longer throws a `NullPointerException` from `equals` or `compareTo` when a record member has no value, nor from `compareTo` for inline records, which carry no record declaration. `equals`, `hashCode` and `compareTo` are consistent with each other again, so record values behave correctly in Java-side hash-based and sorted collections.
+
 ## August 2026
 
 ### Fixed
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/plugin.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.toplevel/models/org/iets3/core/expr/toplevel/plugin.mps
@@ -127,6 +127,7 @@
       <concept id="1068498886292" name="jetbrains.mps.baseLanguage.structure.ParameterDeclaration" flags="ir" index="37vLTG" />
       <concept id="1068498886294" name="jetbrains.mps.baseLanguage.structure.AssignmentExpression" flags="nn" index="37vLTI" />
       <concept id="1225271177708" name="jetbrains.mps.baseLanguage.structure.StringType" flags="in" index="17QB3L" />
+      <concept id="1225271221393" name="jetbrains.mps.baseLanguage.structure.NPENotEqualsExpression" flags="nn" index="17QLQc" />
       <concept id="1225271283259" name="jetbrains.mps.baseLanguage.structure.NPEEqualsExpression" flags="nn" index="17R0WA" />
       <concept id="4972933694980447171" name="jetbrains.mps.baseLanguage.structure.BaseVariableDeclaration" flags="ng" index="19Szcq">
         <child id="5680397130376446158" name="type" index="1tU5fm" />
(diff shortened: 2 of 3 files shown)


arimer:
FYI: The commit IDs (under 2. and 3.)  in the descriptions are no longer valid, probably due to a rebase.

arimer (commented):
Some minor questions:

1. when comparing record declarations as part of the equals impl. of RecordValue. Why are you using the `!=` [operator](http://127.0.0.1:63320/node?ref=r%3A93cd1fe8-b296-405c-a6e6-040c940ccfa1%28org.iets3.core.expr.toplevel.plugin%29%2F6539800829892750581) instead of the `:eq:`?  

Image

~~2. when comparing inline [records via its member names](http://127.0.0.1:63320/node?ref=r%3A93cd1fe8-b296-405c-a6e6-040c940ccfa1%28org.iets3.core.expr.toplevel.plugin%29%2F6539800829895251365), what happens to inline records that have a different amount of members? Couldn't we just have an early return in case the amount of members differ?~~

arimer:
Answering 2) myself:

```
- It's not needed for correctness — the existing per-member loop already terminates with a definite non-zero result whenever the member sets (and hence counts) differ.
- It's not a free optimization — it's a different ordering policy, and would need to be applied consistently (and documented) rather than bolted on as a shortcut, or you'd get inconsistent/contradictory orderings depending on whether the size check happens to trigger.
```

arimer:
Changed 1. as well to `:ne`

arimer (approved):
Looking good. Thx for the vibe 😎

slisson:
> 1. when comparing record declarations as part of the equals impl. of RecordValue. Why are you using the `!=` [operator](http://127.0.0.1:63320/node?ref=r%3A93cd1fe8-b296-405c-a6e6-040c940ccfa1%28org.iets3.core.expr.toplevel.plugin%29%2F6539800829892750581) instead of the `:eq:`?

It probably doesn't make any difference. Even though I like comparing nodes by equality instead of identity because in Modelix I have implementations of SNode where that matters but there are so many places in MPS where nodes are compared by identity that I can't rely on it. None of the two options is wrong.

PR #1926, opened 2026-08-24
IFunctionLike: ignore commented out code in unique name check

slisson:
## Summary

`IFunctionLike.getUniquelyNamedElements()` reported *duplicate name* errors for declarations that are commented out. Commenting out a `val` or a local variable was enough to make the surrounding function show errors on code that is not part of the program.

## What changed

`getUniquelyNamedElements()` collected `ValExpression`/`LocalVarDeclExpr` from `thisNode.body.descendants`. `descendants` also walks into `BaseCommentAttribute` children, so commented out declarations reached mbeddr's `check_ContainerOfUniqueNames` rule, which reports the error on the named element itself — i.e. on the commented out node.

The predicate now additionally requires `it.ancestor<concept = BaseCommentAttribute> == null`.

`BlockExpression.getUniquelyNamedElements()` needs no change: it reads the `expressions` role directly, and commented out children do not live there.

## Test

New `NodesTestCase` `commentedOutUniqueNames` in `test.ts.expr.os.m1`, covering both directions:

- a commented out `val x` duplicating a live `val x`
- a commented out `var a` duplicating the function argument `a`

```
fun commentedOutNamesAreIgnored(a: int) {
  val x = 1
  x + a
  /*val x = 2*/
  /*var a = 3*/
}
```

checked with `CheckNodeForErrorMessagesOperation` (allow warnings, include self).

Evaluated on that node, the old predicate yields `[a, x, x, a]` (two duplicate-name errors), the new one yields `[a, x]`.

`commentedOutUniqueNames` passes (1/1) and the existing `vars` suite still passes (12/12).

## Notes for reviewers

- No breaking changes, no migrations.
- Follow-up to the recent `getUniquelyNamedElements()` rework (`ab01697e28`, `493a3175df`); this closes the remaining case where commented out code was still checked.
- Base is `maintenance/mps20251`, which is where this branch was cut from.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ### Fixed
 - Improve uniqueness name check of `IFunctionLike` `getUniquelyNamedElements()` behavior to avoid overzealous checking.
 - Variability: Feature attribute values are not overwritten anymore if the value stays the same. This avoids changing the model if not necessary, esp. it avoids merge conflicts.
+- `IFunctionLike` `getUniquelyNamedElements()` no longer reports duplicate names for commented out code.
 
 ## July 2026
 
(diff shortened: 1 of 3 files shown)


arimer (approved):
Changes look fine. Thx for the fix. 💪🏼 

I moved the test to a slightly better related location.

PR #1928, opened 2026-08-26
Make EnumLiteral an ISmartReferent, drop QualifierRef

slisson:
## Why

The editor usability of EnumLiteralRef was poor and inconsistent. Looks like 78a6bc01bc6f46f5314dc24d8c2d29657accd3d3 was an attempt to improve it.

In addition, resolveInfo stored only the simple name causing broken references to rebind to a same-named literal in a different enum.

The `qualified` flag was also being enforced by hand in four places: a filter in the `EnumLiteralRef` scope, a `matchingTextFunction` in a hand-written substitute menu, a second menu part working around that filter, and an auxiliary `QualifierRef` concept giving a two-step way to reach what the filter hid.

## What changed

Make the display and resolution namespaces agree, then let MPS's own machinery do the rest.

- **`EnumLiteral implements ISmartReferent`**, overriding `getPresentation(reference)` with a new `presentableName()` — qualified when the enum is, bare otherwise. This feeds `NodePresentationUtil#presentation` and, through the default cascade, `getVisibleMatchingText`.
- **A constraints property getter computes `IResolveInfo#resolveInfo`** from the same method. That is the resolution namespace: it keys `ListScope#forResolvableElements`, it is what `getMatchingText` falls through to, and `ResolveInfoUpdater` rewrites every reference's persisted `resolve=` string from it on save. Computed, never stored — the arrangement `baseLanguage`'s `Classifier` uses for nested class names. `EnumLiteral` was already an `IResolveInfo` via `IValidNamedConcept`; only the getter was missing, which is exactly why the scope fell through to `name`.
- **The `EnumLiteralRef` editor uses a `ref-presentation` cell**, so the editor cell and the completion row read that one method instead of two functions that happened to agree.

With those in place the special-casing is unnecessary and goes:

- `QualifierRef`, the intermediate concept for the two step creation of references to qualified enums. Zero instances, never part of a valid model, so no migration.
- **Both hand-written `EnumLiteralRef` substitute menus.** MPS regenerates the equivalent from the implicit smart reference attribute — a bare `ReferenceScopeSubstituteMenuPart` that reads its matching text from `ISmartReferent`. The editor aspect now carries no `EnumLiteralRef` menu code at all.
- The qualified filter in the `EnumLiteralRef` scope. What keeps a qualified enum's literals out of the general completion list is `PatternUtil` requiring the first typed character to match, so they surface only once the enum's name is typed.

`renderReadable()` now agrees with the editor for qualified enums.

## Tests

Five `EditorTestCase`s in a new `test.kernelf.editor.enums` model, all passing:

| test | types | asserts |
|---|---|---|
| `UnqualifiedLiteralTypedByBareName` | `red` | → `Color.red` |
| `BareNameBindsToUnqualifiedEnumNotToQualifiedOne` | `large` | → `Sizes.large`, **not** qualified `Starbucks.large` |
| `QualifiedLiteralNotOfferedByBareName` | `large` | nothing substituted |
| `QualifiedLiteralTypedByQualifiedName` | `Starbucks:large` | → `Starbucks.large` |
| `RetypeQualifiedLiteralToSiblingLiteral` | `Starbucks:venti` over an existing ref | → `Starbucks.venti` |

The second is the regression test for the rebind bug; the last covers what the deleted parameterized menu part used to provide.

## Notes for review

- Rebinding for **unqualified** enums is untouched, deliberately — a bare literal name is the identity in that namespace, and `ListScope.resolve` already returns null when two candidates share it, so it only ever fires when the name is globally unique.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2026
+
+### Changed
+- KernelF: `EnumLiteral` now implements `ISmartReferent` and derives its `resolveInfo` from the same presentation (qualified for a qualified enum, bare otherwise), so completion, the editor cell and `renderReadable()` agree. The hand-written `EnumLiteralRef` substitute menus and the qualified filter in its scope were dropped in favour of the smart-reference machinery MPS generates. The auxiliary concept `QualifierRef` was removed; it had no instances and was never part of a valid model, so no migration is required.
+
+### Fixed
+- KernelF: a broken reference to a literal of a qualified enum no longer rebinds to a same-named literal of a different enum, because the persisted `resolve=` info now carries the qualified name instead of the simple one.
+
 ## August 2026
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


mfsoliveira (approved):
LGTM.
Tests are covering changes and rebind-safety was manually tested.
I updated the merge to solve a conflict in CHANGELOG.md.

PR #1929, opened 2026-08-26
Fix deadlock on navigation

sergej-koscejev:
Replace all uses of NavigationSupport with EditorNavigator or EditorPanelManager. NavigationSupport is prone to deadlocks in MPS 2025.1. See similar changes in [mbeddr](https://github.com/mbeddr/mbeddr.core/pull/3181) and [MPS-extensions](https://github.com/JetBrains/MPS-extensions/pull/1833).


sergej-koscejev:
Unfortunately I couldn't test it because I don't know exactly where the various methods are called from.

Issue #1921, opened 2026-08-19
Solver run recreates unchanged attribute value nodes, causing node-id churn and merge conflicts


kbirken:
## Why

Every solver run on a feature model configuration clears and rebuilds the values of all automatically assigned feature attributes — even when the newly computed value is identical to the one already there. Since each rebuild creates fresh nodes, this causes:

- large, meaningless diffs in configuration models after every solver run
- avoidable merge conflicts: two branches computing the same value produce two different nodes, so the merge cannot match them and has to be resolved by hand
- occasional corruption of `assignmentCause` when such a merge is resolved in the MPS merge dialog — a value the solver had only *proposed* ends up as *manual*, after which the solver never revises it again
- invalidated node pointers after each run (undo, editor selection, stored results)

Avoiding the pointless rewrite fixes all of these at the source, rather than teaching the merge to cope with them.

## How

Only write an attribute value when it actually differs from the value already present. The equality check for this already exists in `ConfigurationSolverFacade.hasAttributeValueChanged` (the `matches` operator) — it is used to decide *whether* an update is needed, but the writers then rewrite everything unconditionally. Apply the same check per assignment at write time, in `CommonBaseUtil.setAttributes` and in the blanket clearing done by `unsetAutomaticAttributeAssignment`. Unchanged values then keep their nodes, and their identity, across runs and branches.

Worth exposing the "write only if changed" step as a small reusable helper, so the companion change can call it instead of duplicating the logic.

## Companion issue

IETS3/iets3.core#1778 — the same write pattern in the default-value optimizer. Both need to be fixed together; this issue goes first, the `iets3.core` side follows once the dependency is bumped.

kbirken:
PR has been merged successfully, closing.

PR #1931, opened 2026-08-27
Variability: Do not overwrite feature attribute assignments if not necessary

kbirken:
When the solver runs on a configuration, compulsory attribute values are set if the value is mandatory. If the solver computes the previous value again, the value expression has been replaced. This has a couple of drawbacks, as the node identity of the value (e.g. a `NumberLiteral`) will be changed although the semantic value stays the same.

The PR adds a test which reproduces the problem and introduces the fix. It also does some API changes necessary for solving the companion PR in IETS3/iets3.core#1781.

Additionally, the helper classes `AssignmentValueMap` and `AssignmentValueList` have been introduced to make the new code more readable and also refactor the existing code. 

CHANGELOG has been updated.

Note: The API for iets3.core has been changed. So iets3.core has to be adapted to use this version of iets3.opensource. See also the companion PR in IETS3/iets3.core#1781.

TODO: I keep separate commits to make the review easier. Do not merge from maintenance/mps20241, instead squash the commits after the review.

This solves #1921.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,20 @@ Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelo
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
 ## August 2026
+
 ### Fixed
 - Improve uniqueness name check of `IFunctionLike` `getUniquelyNamedElements()` behavior to avoid overzealous checking.
+- Variability: Feature attribute values are not overwritten anymore if the value stays the same. This avoids changing the model if not necessary, esp. it avoids merge conflicts.
 
 ## July 2026
+
 ### Added
 - IFunctionLike takes arguments and named body content into account when performing ab uniqueness name check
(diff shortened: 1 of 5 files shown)


dbinkele:
@kbirken The two constructors can be made protected in my opinion http://127.0.0.1:63320/node?ref=r%3A83d54567-d361-47ca-a949-7374ea89921d%28org.iets3.variability.configuration.base.plugin%29%2F8045029366699539475

dbinkele:
@kbirken  the same is true for all methods except the 2nd constructor. Make them protected http://127.0.0.1:63320/node?ref=r%3A83d54567-d361-47ca-a949-7374ea89921d%28org.iets3.variability.configuration.base.plugin%29%2F8045029366703308175

dbinkele:
@kbirken There are two places where it is decided if the new and old value of a FAA 'match' (os: CommonBaseUtil, core: AttributeDefaultValueOptimizer). I think it makes sense to encapsulate this behavior in one method (maybe in CommonBaseUtil) to ensure future consistency.

dbinkele (approved):

PR #1934, opened 2026-08-31
tracing: increased length limit of entries in the trace explorer

slisson:
Entries in the the trace explorer where truncated with a very low limit, which made the trace hard to read.
There was also no way to see the full text.

Increased the limit and also changed the tooltip to show the non-truncated text.

**Before**

Trace explorer entry truncated at 25 characters, tooltip showing only FunctionCall

**After**

Trace explorer entry shown in full, tooltip reading defaultWithOtherwiseWithPredefXY(1, 1) = 1 [FunctionCall]


mfsoliveira (approved):
Limits increased and the view in trace is improved

PR #1936, opened 2026-09-01
fix: keep the interpreter trace when an evaluation throws.

slisson:
## Problem

When the interpreter throws an exception while a trace is being built, **Show Trace** shows nothing at all: no
tab opens and the exception surfaces as an IDE error. Everything the interpreter had already
computed is discarded with it.

The trace is not actually lost when that happens. A node's `ComputationTrace` record is added to its
parent *before* that node is evaluated, and every generated evaluator wraps whatever escapes into an
`InterpreterRuntimeException` carrying the node and its record — re-wrapping at each enclosing
level, so the innermost record sits at the end of the cause chain. `InterpreterEscapeException`
carries the same thing in `failedTrace`. Nothing was using any of it:

- `ITraceRoot.defaultRerunnableInterpreted` (the path behind Show Trace on functions, constants and
  function calls) and `AbstractTestItem.getRootTrace` did not catch at all.
- The test items' `catch` branches set an error message but never called `setTrace`.

## Fix

New `TraceFailures` in `org.iets3.core.expr.base.behavior`:

- `markFailure(Throwable)` walks the cause chain to the innermost record, stamps it with a one-line
  `failed with <Exception>: <message>` (an existing message is not overwritten), clears `showNever`
  and sets `showInAnyCase` + `markForReveal`, so the trace explorer's filters keep it and render it
  in red.
- `traceForFailure(node, partialTrace, throwable)` returns the root trace to show and never returns
  null — it falls back to a single record for the node when the interpreter did not get far enough
  to build one.

Wired into both trace roots and into the `catch` branches of `AssertTestItem`,
`AssertThatTestItem`, `AssertOptionTestItem` and `ConstraintFailedTestItem`.
`AssertTestItem.constructCustomFrame` rebuilds the root record, so it now carries a failure
recorded there over to the frame. `IETS3ExprEvaluator.getLastLog` no longer throws when `evaluate()`
failed before the helper existed.

Everything this uses (`getFailureTrace`, `failedTrace`, `unmarkAsShowNever`, `markAsShowInAnyCase`,
`markForReveal`, `rootTrace`) is already present in the released
`com.mbeddr.mpsutil.interpreter.rt` artifact, so no platform change is needed.

## Reproducing

Any trace root whose evaluation throws will do. For the screenshots below, a `Constant` in a
`Library` with one sub-expression that is fine and one that blows up at runtime:

```
val boom = (1 + 2) * (3 + 4) + 10 / (5 - 5)
```

The model is legal — the type checker reports no problem on it — but `10 / (5 - 5)` raises
`ConstraintFailedException` after `(1 + 2) * (3 + 4)` has already been computed. Invoke **Show
Trace** on it.

## Before

No tab opens. The exception escapes `ITraceRoot.defaultRerunnableInterpreted` into action dispatch
and lands in the IDE error log:

```
Action dispatch failed. Thread AWT-EventQueue-0, state ACTIVE, 1 active clients.
org.iets3.core.expr.base.plugin.ConstraintFailedException: division by zero
    at o.i.c.expr.base//org.iets3.core.expr.base.plugin.ArithmeticErrorHelper...
    at c.m.m.interpreter.rt//com.mbeddr.mpsutil.interpreter.rt.InterpreterBase...
    ...
    at o.i.c.expr.base//org.iets3.core.expr.base.behavior.IETS3ExprEvaluator...
    at o.i.c.expr.base//org.iets3.core.expr.base.behavior.ITraceRoot__Behavior...
```

## After

The tab opens with the trace as it stood when the exception was raised. `(1 + 2) * (3 + 4) ⇒ 21`
and its sub-results are there with their timings, and the node that threw is marked in red with the
exception and flagged `[R]`.

> The screenshot uses the *show all non-skippable trace nodes* filter so the whole tree is visible.
> Under the default *reduced* filter the same trace shows the marked failing node on its own — that
> filter keeps frames and show-in-any-case nodes, and a constant has no frames, so this is its
> normal behaviour rather than something this change introduces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2026
+
+### Fixed
+- KernelF: an exception thrown by the interpreter no longer costs the whole trace. `Show Trace` now opens with everything that was computed before the exception, and the node that threw is marked with the exception and highlighted in the trace tree. This covers the generic trace roots (functions, constants, function calls) as well as the test items, whose `catch` branches used to drop the trace.
+
 ## August 2026
 
 ### Fixed
--- a/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.core.expr.base/models/behavior.mps
@@ -20431,31 +20431,23 @@
                     <node concept="3clFbS" id="5IR_boHRmeo" role="3clF47">
                       <node concept="3cpWs8" id="5af9jCuJP7w" role="3cqZAp">
                         <node concept="3cpWsn" id="5af9jCuJP7x" role="3cpWs9">
-                          <property role="TrG5h" value="result" />
+                          <property role="TrG5h" value="evaluator" />
                           <node concept="2OqwBi" id="5af9jCuJP7y" role="33vP2m">
-                            <node concept="2OqwBi" id="5af9jCuJP7z" role="2Oq$k0">
-                              <node concept="2ShNRf" id="5af9jCuJP7$" role="2Oq$k0">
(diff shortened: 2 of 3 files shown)


arimer (approved):
Thanks for the conribution. Changes look decent. 🤗

PR #1846, opened 2026-06-23
Feature/create new i dot targets for enumerations #1827

dbinkele:
Closes #1827

## Motivation

`org.iets3.core.expr.toplevel` already provides two `IDotTarget`s for enumerations — `is()` (`EnumIsTarget`) and `isIn()` (`EnumIsInTarget`) — to check the presence of an enumeration element. Users frequently need the negated checks too, which previously had to be written by hand (e.g. `not(x.is(...))`). This PR adds first-class negated variants.

## Changes

- **New concepts (`org.iets3.core.expr.toplevel`):** Introduced `isNot()` and `isNotIn()`, the negated versions of `is()` and `isIn()`.
  - `isNot()` — checks that an enumeration element is **not equal** to another.
  - `isNotIn()` — checks that an enumeration element is **not contained** in a list.
  - They look and behave exactly like their non-negated counterparts (editor, constraints, type system).
- **Common super concept:** Behavior, constraints and type system shared by the negated and non-negated targets were consolidated into a common super concept to remove duplication and simplify the generator.
- **Generator:** Added `genjava` generation support for `isNot()` / `isNotIn()`.
- **Interpreter:** Added interpreter support for both new operators.
- **Tests:** Extended the generation (`test.ts.expr.os`) and interpretation (`test.in.expr.os.enums`) test models to cover the new operators.
- **CHANGELOG** updated.

## Testing

Generation and interpreter tests for the enumeration operators were extended and pass.

PR #1943, opened 2026-09-02
fix: keep enum 'is' and 'isIn' working on unmigrated models

slisson:
The PR https://github.com/IETS3/iets3.opensource/pull/1846 introduced a breaking change. Even though it also provided a migration, breaking changes should still be avoided because they make the migration mandatory, which makes working with feature branches unnecessarily difficult.

This PR restores the backwards compatibility.

## What changed

Made the editor and generator compatible with unmigrated concept instances.

Also fixed an incorrect implementation of `renderReadable()` that was discovered while working on this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ The project does _not_ follow Semantic Versioning and the changes are documented
 ## September 2026
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


NilsWeidmann (approved):
Works as expected!

PR #1952, opened 2026-09-08
Variability: Adapt all transitively extending configurations from an abstract configuration

### LINKED ISSUE #1948: Variability: Adapt all transitively extending configurations from an abstract configuration
When an **abstract** configuration is edited, every configuration that extends it (directly or transitively) becomes stale and has to be reconciled one by one via the existing single-config intention *"Adapt this configuration to the extended configuration"*. For deeper `extends` chains this is tedious and easy to get wrong, because the chain has to be worked top-down to be correct.

There should be a batch operation on the abstract configuration itself — the inheritance counterpart to the existing *"Update All Configurations"* intention on the feature model.

### Proposed solution

New intention **"Adapt Extending Configurations to Changes"** on `FeatureModelConfiguration`, visible only when the configuration is `abstract` (`org.iets3.variability.configuration.base.intentions`, intention `adaptExtendingFMCs`).

It runs a new task `UpdateAllExtendingConfigs` (`…configuration.base.behavior`) which

* collects the **transitive closure** of extending configurations, level by level, over `FeatureModelConfiguration.extendedFMC.config`, restricted to `ConfigurationScopeProvider.configurationsScope(featureModel)` of the abstract config's feature model — so the extension chain is processed top-down;
* reconciles each of them through the existing single-config logic `FixAdaptToExtendedFMC.run(config)`, i.e. intention, quickfix and batch run stay behaviourally identical;
* logs the number of updated configurations and the total duration.

### Reuse of the existing batch machinery

`AbstractUpdateConfigsTask` currently hard-codes the feature-model propagation (`ConfigUpdateHelper.propagateFeatureModelChangesToConfig`) and the progress text `"Updating structure according to feature model…"`. Generalize it with two overridable hooks so both batch operations can share it:

* `protected void propagate(node<FeatureModelConfiguration> config)` — default: the existing FM propagation;
* `protected string updateReason()` — default: the existing progress text; `UpdateAllExtendingConfigs` returns `"Updating transitively extending configurations of abstract configuration <name>"`.

This keeps the properties established in #1536/#1571 for free: modal progress dialog, strictly sequential execution, postponed event handling, timing logs, and the cancel semantics (already-updated configurations stay updated, re-run completes the rest).

dbinkele:
## Summary

Implements #1948: on an **abstract** feature model configuration, all configurations that extend it — directly or transitively — can now be reconciled in one step, instead of applying the single-config fix *"Adapt this Configuration to the extended Configuration"* one by one along the `extends` chain.

## Changes

- **Intention** — new `adaptExtendingFMCs` [(node url)](http://127.0.0.1:63320/node?ref=r%3Adfaa2422-aef5-456d-a8cd-942c81b870e6%28org.iets3.variability.configuration.base.intentions%29%2F8014811126264422850) *"Adapt Extending Configurations to Changes"* on `FeatureModelConfiguration`, visible only when the configuration is `abstract`. It runs the batch task in a modal progress dialog and refreshes the affected editors afterwards. The language therefore now depends on `com.mbeddr.mpsutil.intentions`.

- **Batch task** — `UpdateAllExtendingConfigs` [(node url)](http://127.0.0.1:63320/node?ref=r%3Adedd19c9-9ff3-4f30-aa73-ce61203b2296%28org.iets3.variability.configuration.base.behavior%29%2F8014811126261165998) collects the transitive closure of extending configurations level by level over `FeatureModelConfiguration.extendedFMC.config`, restricted to the `ConfigurationScopeProvider.configurationsScope(...)` of the abstract configuration's feature model, so the chain is processed top-down. Each configuration is reconciled through the existing `FixAdaptToExtendedFMC.run(config)`, so intention, quickfix and batch run stay behaviourally identical.

- **Shared machinery** — `AbstractUpdateConfigsTask` [(node url)](http://127.0.0.1:63320/node?ref=r%3Adedd19c9-9ff3-4f30-aa73-ce61203b2296%28org.iets3.variability.configuration.base.behavior%29%2F6279010743056184611) was generalized: which configurations to update, how to propagate, the progress text and the final log message are now hooks. The feature-model defaults (`ConfigUpdateHelper.propagateFeatureModelChangesToConfig`, *"Updating structure according to feature model…"*) moved into the new intermediate class `ConfigFromFeatureModelUpdater` [(node url)](http://127.0.0.1:63320/node?ref=r%3Adedd19c9-9ff3-4f30-aa73-ce61203b2296%28org.iets3.variability.configuration.base.behavior%29%2F6995660538504659200), which `UpdateAllConfigsTask`, `UpdateOneConfigTask` and `UpdateAllExtendingConfigs` extend. Modal progress, strictly sequential execution, postponed event handling, timing logs and the cancel semantics established in #1536 / #1571 are unchanged.

- **Read-only models** — the dirty-marking at the end of the task is now skipped for models with a read-only data source (`SModelBase.isReadOnly()`). Otherwise `setChanged(true)` logs *"Attempt to change a model with read-only data source"*, which fails the node tests in the packaged Gradle test run while passing in the IDE.

- **Test** — node test `UpdateAllExtendingConfigurations` in `test.org.iets3.variability.configuration.base.checking_rules` covering an `extends` chain of three configurations; it drives the task with a `DummyIndicator` so it can run without a progress dialog.

Resolves #1948. The base branch is `maintenance/mps20241`, so the ticket needs to be closed manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2026
+
+### Added
+- Variability: New intention "Adapt Extending Configurations to Changes" on abstract feature model configurations. It propagates changes of the abstract configuration to *all* transitively extending configurations in one step (with progress reporting), instead of applying the per-configuration fix "Adapt this Configuration to the extended Configuration" one-by-one.
+
+### Changed
+- Variability: The update-configurations tasks were refactored - the common functionality of the tasks updating one, all, or all extending configurations now lives in the shared base classes `AbstractUpdateConfigsTask` / `ConfigFromFeatureModelUpdater`.
+
 ## August 2026
 
 ### Fixed
--- a/code/languages/org.iets3.opensource/languages/org.iets3.variability.configuration.base/models/org.iets3.variability.configuration.base.behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.variability.configuration.base/models/org.iets3.variability.configuration.base.behavior.mps
@@ -25482,25 +25482,51 @@
             </node>
           </node>
         </node>
-        <node concept="3clFbF" id="6IPgvt9CNFG" role="3cqZAp">
-          <node concept="2OqwBi" id="6IPgvt9DciF" role="3clFbG">
-            <node concept="1eOMI4" id="6IPgvt9D9MA" role="2Oq$k0">
-              <node concept="10QFUN" id="6IPgvt9CQ81" role="1eOMHV">
-                <node concept="3uibUv" id="6IPgvt9D3ww" role="10QFUM">
-                  <ref role="3uigEE" to="mhbf:~EditableSModel" resolve="EditableSModel" />
-                </node>
-                <node concept="2OqwBi" id="5BtXES5opZX" role="10QFUP">
-                  <node concept="Xjq3P" id="5BtXES5onB7" role="2Oq$k0" />
-                  <node concept="2OwXpG" id="5BtXES5otak" role="2OqNvi">
-                    <ref role="2Oxat5" node="5BtXES5m45o" resolve="model" />
+        <node concept="3clFbJ" id="2FiNa4PSxEb" role="3cqZAp">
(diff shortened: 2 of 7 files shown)


mfsoliveira (approved):
It was long ago I have seem the variability code. The UX is getting nicer!
Progress bar, execution time, logging - cool

I saw the changes, they are working and they are covered with tests.

Now with more use of log I would always extract the log statement into a method, so that mixed concerns do not disturb so much. It is better to see it like in

UpdateAllExtendingConfigs.logFinalMessage

the same could be done for UpdateAllExtendingConfigs in the line 27 and for all logs in the middle of the code. It is not a MUST. It more a preference, with time we can extract them when we find one.

PR #1955, opened 2026-09-10
Variability: Apply inheritance to sub-configurations with unspecified content

### LINKED ISSUE #1953: Variability: Adapting to the extended configuration ignores sub-configurations with unspecified content
When a configuration extends another one and one of its sub-configurations is a **feature-model include** whose content is still the unspecified `FeatureModelConfigurationBase` (i.e. the nested feature-model configuration has never been inlined), the intention *"Adapt this configuration to the extended configuration"* silently skips that sub-tree: `FeatureModelConfigurationInheritanceUtil.applyInheritance(...)` only walks `InlineFeatureConfigurationContent`, so for an unspecified content there is nothing to walk and no value of the extended configuration is inherited.

### Steps to reproduce

1. Feature model `M` with a feature-model include `ink` pointing to a feature model `Ink` (features `A` … `E`).
2. Configuration `CM` for `M` which selects values below `ink`.
3. Configuration `CM2` for `M` which `extends CM` and leaves `ink` unspecified (content is a `FeatureModelConfigurationBase`).
4. Invoke *"Adapt this configuration to the extended configuration"* on `CM2`.

**Actual:** `ink` in `CM2` stays unspecified; nothing is inherited from `CM`.

**Expected:** the nested configuration is materialized and the values of `CM` are inherited.

### Proposed solution

At the start of `applyInheritance(...)` (`FeatureModelConfigurationInheritanceUtil`, model `org.iets3.variability.configuration.base.behavior`), replace an unspecified `FeatureModelConfigurationBase` content of the sub-configuration by an inline content generated from the referenced feature model — using the existing helper `FeatureModelConfigurationConstraintsUtil.configContentByFeature(<feature model root>)`, which is already used by the constraints and by the "inline configuration" intention.

The regular inheritance logic then runs on the materialized content, so no separate code path is needed and intention, quickfix and the batch operations stay behaviourally identical.

To be covered by a new test `ApplyExtendsWithUnspecified` in `test.org.iets3.variability.configuration.base.inheritance@tests`, next to the existing `InheritanceFeatureModelConfigurationBaseAdd/Remove` tests.

dbinkele:
## Summary

"Adapt this configuration to the extended configuration" silently skipped sub-configurations whose content was still the unspecified `FeatureModelConfigurationBase` — nothing was inherited from the extended configuration. `applyInheritance(...)` now materializes such a content as an inline configuration derived from the referenced feature model (via the existing `FeatureModelConfigurationConstraintsUtil.configContentByFeature(...)`), so the regular inheritance logic applies and intention, quickfix and batch operations stay behaviourally identical.

While fixing this it turned out that inheritance could also *discard* content of the extending configuration when the two sides use different content kinds. The adaption therefore now checks up front whether the contents can be reconciled at all.

## Changes

**`FeatureModelConfigurationInheritanceUtil`** (`org.iets3.variability.configuration.base.behavior`)
- `applyInheritance(...)`: an unspecified sub-configuration content is replaced by inline content generated from the referenced feature model before inheritance is applied.
- New `hasInheritanceConflicts(...)` / `contentOkToApplyInheritance(...)`: report a conflict instead of silently losing content. An unspecified content on either side is always fine; if the extended configuration references a configuration, the extending one must reference the same configuration or one that extends it; if the extended configuration is inlined, the extending one must be inlined too so both can be reconciled subfeature-wise.
- Adapting keeps an existing (narrowed) configuration reference instead of replacing it with inline content.

**Tests** (`test.org.iets3.variability.configuration.base`)
- New cases in `…base.inheritance@tests`: `ApplyExtendsWithUnspecified`, `ExtendsWithTwoUnspecifiedIncludes`, `ExtendsWithSameConfigRefInInclude`, `InheritanceExtendsRefNarrowedToSubConfig` and the three conflict cases `InheritanceExtendsConflict{DifferentRefs, InlineExtendsRef, RefExtendsInline}`.
- Existing inheritance tests tightened to expect the specific error/warning instead of any message.
- New `ConfigForModelWithDependencyCycle` in `…base.checking_rules@tests`.

CHANGELOG updated.

Resolves #1953.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project are documented in this file.
 Format of the log is _loosely_ based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 The project does _not_ follow Semantic Versioning and the changes are documented in reverse chronological order, grouped by calendar month.
 
+## September 2026
+
+### Fixed
+- Variability: "Adapt this configuration to the extended configuration" no longer skips sub-configurations whose content is still unspecified (`FeatureModelConfigurationBase`). Such a content is now materialized as an inline configuration derived from the referenced feature model, so the values of the extended configuration are inherited ([#1953](https://github.com/IETS3/iets3.opensource/issues/1953)).
+
 ## August 2026
 
 ### Fixed
--- a/code/languages/org.iets3.opensource/languages/org.iets3.variability.configuration.base/models/org.iets3.variability.configuration.base.behavior.mps
+++ b/code/languages/org.iets3.opensource/languages/org.iets3.variability.configuration.base/models/org.iets3.variability.configuration.base.behavior.mps
@@ -61,6 +61,7 @@
     <import index="alof" ref="742f6602-5a2f-4313-aa6e-ae1cd4ffdc61/java:jetbrains.mps.ide.project(MPS.Platform/)" />
     <import index="kvq8" ref="r:2e938759-cfd0-47cd-9046-896d85204f59(de.slisson.mps.hacks.editor)" />
     <import index="exr9" ref="1ed103c3-3aa6-49b7-9c21-6765ee11f224/java:jetbrains.mps.nodeEditor(MPS.Editor/)" />
+    <import index="guwi" ref="6354ebe7-c22a-4a0f-ac54-50b52ab9b065/java:java.io(JDK/)" />
   </imports>
   <registry>
     <language id="af65afd8-f0dd-4942-87d9-63a55f2a9db1" name="jetbrains.mps.lang.behavior">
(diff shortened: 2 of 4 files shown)


kbirken:
Started with the review. First finding: One unneeded dependency has been introduced to `org.iets3.variability.configuration.base.behavior`, another one (with `...intentions` should be removed as well. 

image

kbirken (changes requested):
I did a review and a smoketest. It is quite hard to test everything, as there are many cases. But I couldn't break it, which is at least a good sign.

Here are the remaining findings (not related to functionality):
- There are some unneeded dependencies in `org.iets3.variability.configuration.base.behavior`:  `java.io` (introduced by this PR), `configuration.base.intentions` (already introduced earlier). Should both be cleaned up.
- Test `ConfigForModelWithDependencyCycle` introduced two new model errors, should be fixed. Probably two `check` annotations should fix (hide) it here.
- Test `MustAdaptToChangedInheritedValue` also has a model check error which could be fixed (hidden) by adding a `check` annotation.

kbirken (dismissed):
LGTM now, thanks for fixing the findings. Pls. do not forget to squash the last commits before merging (everything after the last merge commit).

kbirken (approved):
Update of approval after migrate/remigrate.

Issue #1961, opened 2026-09-16
KernelF: Hexadecimal number literals cannot be deleted


TorstenGoerg:
`NumberLiteral` nodes can be deleted by pressing `DELETE` or `BACKSPACE` as long as the number literal value is persented in decimal format. If the number literal value is shown in hexadecimal format this kind of deletion is not possible. When pressing `DELETE` or `BACKSPACE` in this case the number literal value is just temporarily removed from the respective editor cell. But after selecting another cell in the editor, e.g., with a mouse click the number literal value appears again. The underlying `NumberLiteral` node is not deleted in this case.

PR #1962, opened 2026-09-16
bugfix: Allow deletion of hexadecimal number literals by adding a workaround

TorstenGoerg:
This PR adds a workaround that allows the deletion of whole number literals nodes by pressing `BACKSPACE` or `DELETE` in case that the number literal value is presented in hexadecimal format. This kind of deletion was only possible for decimal number literals before due to an internal implementation problem.

Closes bug issue https://github.com/IETS3/iets3.opensource/issues/1961 .


TorstenGoerg:
It would be nice to have respective editor tests for this bug fix. But this doesn't seem to be possible editor tests do not properly work in cases where grammar cells are used.

kbirken (approved):
LGTM. The workaround works :-). I tested it in a customer project where hexadecimal literals have been switched on. Didn't know how to test in plain iets3.opensource. 

Could you pls. file a ticket in mps-extensions specifying the problem (probably in `SplittableCell`), so that we can get rid of the workaround someday?

